PR #33, opened 2018-10-11
Add support for SVN repositories and source archives, add releases for Ini_Files and GNATCOLL core

pmderodat:


mosteo:
I will take a look to this and the other related one in alr ASAP.

pmderodat:
Great, thanks! :-)

mosteo:
Everything looks fine to me, thanks. Just a couple of comments:

- Source packages probably should have a hash field (sha-something?) to verify the downloaded file, this can be added later so I've created issue #34 for it.
- The Ini_Files/Gnatcoll index files are using the old way (which works okay) of declaring a constant; I have moved to a newer way that relies on packages to avoid the duplication of the version in the constant and in the string description. See for example https://github.com/alire-project/alire/blob/master/index/alire-index-alire.ads (in case you prefer it for future releases)
- The "08" version in Ini_Files will lose the leading zero when transformed to a semantic version (it parses to integers, so that will become 8.0.0). If that information is important it should be preserved in some other way (in the build field of the semantic version for example).

If none of those are a problem to you I'll merge the PR.

pmderodat:
I agree adding a hash field could be useful. Do you want me to rework my changes to include this feature? What about using SHA-512?

By the way, I was thinking it would also be useful for Git/SVN checkouts: I’m not sure relying on Git’s commit hashes is reasonable because of the currently known SHA-1 weaknesses and those that could come in the future.

I’ll try to refactor both of my index changes following your suggestions (new way to avoid duplication and semantic version). For the “08”, I just copied the project’s original version number, but since version numbers have semantics here, it’s probably going to be confusing to have 08.

Thank you for the review!

mosteo:
SHA-512 is in GNAT already so it seems OK to me.

For git I don't see any practical alternative to SHA-1... That's what they defined as unique id, so that's it I guess. (Incidentally, Mercurial also uses it).

It's been some years since I used any SVN. I'm thinking now that they rely on revision numbers, so I guess that that should be their unique id. Sorry for the oversight on first read. (Just checked it here: https://en.wikipedia.org/wiki/Comparison_of_version_control_software#Technical_information).

pmderodat:
I just have pushed updates for the GNATCOLL and Ini_Files package descriptions. There is one thing I don’t understand: in https://github.com/alire-project/alire/blob/master/index/alire-index-gnatcoll.ads#L46 we rename Base to `GNATCOLL.Project` while I would expect `GNATCOLL.Slim.Project` instead… is that a pasto? To sort it out, I wanted to check the list of available packages/versions, but `alr list` lists only package description: is there a way to get it list all available versions of a package?

I haven’t progressed on using SHA-512 because it’s not a simple feature: we could try to use the `sha512sum` command-line tool, but I’m not sure it’s realistic to expect it to be present on all systems. There is `GNAT.SHA512`, but then we need to feed it with the source archive’s content… That requires to read binary files, which either requires to use `GNAT.OS_Lib`, either `GNATCOLL.Mmap`. What do you think?

mosteo:
`alr search --list --full` should give you every indexed release.

I'm already using lots of GNAT.OS_Lib and GNATCOLL would be a new dependency in alr so I'd say to rely on OS_Lib if that's enough unless there's some clear advantage on the latter.

About the Renaming: this is kind of a misnomer and should probably be refactored to "Providing" (specially since that's used in https://github.com/alire-project/alire/blob/9ba483b9b945233a73a1b4b8bd9cb47c8f07ae9a/src/alire-releases.ads#L100). The thing is that a project that has Renaming will use that name for itself during dependency resolution. It's like the "provides" in debian packages, to impersonate or fill-in for another (or a virtual) package. 

The use case is (here, gnatcoll=gnatcoll-core): since gnatcoll-slim is a subset of gnatcoll, it would be a conflict to have both of them in dependencies (compilation would find duplicate files). So, one project that only needs gnatcoll-slim specifies that it depends on gnatcoll-slim OR gnatcoll (order matters). If another unrelated dependency in turn depends on gnatcoll, this will force the solver to choose gnatcoll. Otherwise gnatcoll-slim will be chosen first.

I think the only example of this OR dependency is in https://github.com/alire-project/alire/blob/master/index/alire-index-xstrings.ads right now (which I did in purpose to test this). There, gnatcoll is missing because it wasn't packaged at the time (due to not trivial dependencies).

Hope this helps!

pmderodat:
Thank you for your explanations regarding `Renaming` and `Provides`! I’ll have a closer look in the ligth of it.

Let’s go for SHA-512, then! Or instead, maybe we should give some room so that other hashsums could be used in the future: let users provide (hash-kind, hash value) couples. It’s not straightforward work, so are you fine with validating the current changes here and me doing this hash work as a next step? (I will file it under #34)

For security concerns, I’m not sure relying on VCS IDs is a good practice: if the remote server is impersonated, we need a secure hash to detect source changes. Git uses SHA-1 for its unique IDs but I’ve read SHA-1 had some weaknesses discovered several years ago so I think it’s worth considering using a stronger hash in addition.

Regarding `GNAT.OS_Lib` versus `GNATCOLL.Mmap`: using the former is probably fine. I was considering the latter purely for performance reasons. We can revisit later if need be, I guess.

mosteo:
OK, then I'm merging this one right away and we can discuss things further as they come.

The problem of repository impersonation does not seem trivial but also I don't feel it's pressing. I can see different approaches like hashing checked out contents and/or relying on certificates/manifests; in any case it would be something on top of the commit id which I think is unavoidable as the unique key. Also signed commits might have some relevance. Perhaps we can open another issue and discuss details there. I agree that ensuring integrity of external sources is important.

Linking these articles on this for future reference: 
https://www.mercurial-scm.org/wiki/mpm/SHA1
https://lwn.net/Articles/715716/

PR #41, opened 2019-01-30
RFC: Catalog as data files, not as code

pmderodat:
Hello @mosteo,

This PR is a proof of concept to replace the current package index (catalog), built up on Ada source files to be compiled with `alr`, into a set of simple data files which any library/program can load without recompilation.

I don’t intend to discuss here syntax specifics, but rather the mechanism to load the index and the formalism to describe packages. In favor of simplicity, I have used here the JSON format and GNATCOLL.JSON to parse it, but I don’t think this is to last: JSON isn’t the most friendly format for us human to write, so I think we should move to another format later (and that will probably require to write a parser).

Here is a tentative to formalize package descriptions: https://github.com/alire-project/alire/blob/73ae648a50c4ad3da6dff8a85491c08c7ddd8120/doc/catalog-format-spec.rst

This is just a prototype: it still misses several checks, for instance when loading a spec, checking that all of its dependencies actually exists, or checking that all referenced variables exists and that all their values are covered in case statements. Nevertheless, I think this prototype is ready enough to start debating. :-) You can use my [alr fork](https://github.com/pmderodat/alr/) to experiment with it locally.

Thank you in advance for having a look at this!


doc/catalog-format-spec.rst
```diff
@@ -0,0 +1,364 @@
...
+some point, some other syntax should be used, in particular one that is more
+human-friendly to read and write.
+
+The main interest here is the use of simple data structures (booleans, strings,
+lists and mappings) to write package descriptions.
+
+Big picture
+-----------
+
+Each package is described as a JSON file. For instance: ``aaa.json`` for the
```
gusthoff: Why use JSON format? Introducing yet another format to the toolchain just makes things more complicated. There are already two formats for building binaries (GPRbuild and Makefile). Instead, it makes much more sense to use the same format as GPRbuild, so the user gets an unified experience!

doc/catalog-format-spec.rst
```diff
@@ -0,0 +1,364 @@
...
+
+    * ``LGPL 3.0`` for the corresponding Lesser General Public License.
+
+    * ``MIT`` for the MIT License.
+
+    * ``Public Domain`` for code in the public domain.
+
+    * ``zlib`` fo the zlib License.
+
+    * ``custom:XXX``, to be used for licenses unlisted here, where ``XXX`` is the
```
gusthoff: What about: "where `XXX` is the name of the separate file (usually in the same directory) containing the licensing conditions"?

doc/catalog-format-spec.rst
```diff
@@ -0,0 +1,364 @@
...
+
+  .. code-block:: json
+
+   "licenses": ["GPL 3.0", "MIT"]
+
+  Custom license example:
+
+  .. code-block:: json
+
+   "licenses": ["custom:do What The Fuck you want to Public License\n\nVersion 1.0, March 2000\nCopyright (C) 2000 Banlu Kemiyatorn (d).\n136 Nives 7 Jangwattana 14 Laksi Bangkok\nEveryone is permitted to copy and distribute verbatim copies\nof this license document, but changing it is not allowed.\n\nOk, the purpose of this license is simple\nand you just\n\nDO WHAT THE FUCK YOU WANT TO.\n"]
```
gusthoff: Civility?

doc/catalog-format-spec.rst
```diff
@@ -0,0 +1,364 @@
...
+
+   "project-files": ["my_project.gpr", "utils/utils_for_my_project.gpr"]
+
+* ``gpr-externals``: optional object, giving a mapping from the name of external
+  variables in the project files to sets of possible values (as array of
+  strings), or ``null`` if this set is infinite. For instance:
+
+  .. code-block:: json
+
+   "gpr-externals": {
```
gusthoff: This could be directly included from GPR files, especially if the same syntax is used.

doc/catalog-format-spec.rst
```diff
@@ -0,0 +1,364 @@
...
+                          "windows": "ms-windows"}}
+   }
+
+* ``executables``: optional list of strings. Each is a path, relative to the root
+  of the source directory, to an executable provided by the package. For
+  instance:
+
+  .. code-block:: json
+
+   "executables": ["bin/my_main"]
```
gusthoff: This could be automatically detected based on information from GPR files...

gusthoff (commented):
Just added some comments on the format...

doc/catalog-format-spec.rst
pmderodat: Using project files sounds indeed appealing at first, because some information the package manager needs is already there. However, it’s far from perfect: it is not possible for instance in project files to express conditional dependencies, or to add version constraints to the `with` clauses. Another reason this is not the perfect match is that some projects are materialized as several project files, so there is no central place to contain information such as authors and website. Yet another one is that as of today, available technology to load project files does too many things, such as listing source files, which the package manager should avoid, for performance and because it should not need to load all files from all possible projects when performing a keyword search in the catalog. I would add one more reason: one project file describes only one release, whereas the package manager needs to have visibility over several releases simultaneously, knowing that the releases are different versions of the same package: project files seem to be a poor match to describe this conveniently.

So, in the end, I think using project files introduces more problems than the ones it solves. As a consolation, I’d like to point out that the goal of project files (just like Makefiles) is to describe how to *build* software. The package manager needs different describtions in order to provide a consistent database of programs/libraries and to ease the various actions to do before running builds: getting sources and setup the environment for the build to succeed.

My hope is to keep this description format simple enough and with as little surprise as possible in practice so that there’s not much cognitive overload in using it!

doc/catalog-format-spec.rst
pmderodat: Do you mean the license should be included in the catalog files tree? Hm… that sounds better than having a fixed list of licenses in Alire and would still allow one to filter searches by license…

doc/catalog-format-spec.rst
pmderodat: I was looking for a real world and very short license text to use as an example. I did not invent it: https://en.wikipedia.org/wiki/WTFPL

Do you have a suggestion to replace this example?

mosteo:
I'm sorry for being out of touch for so long! I will take a look at this the next week, I'm traveling until wednesday. However, if this is proof-of-concept, do we need to merge it? Or is there code that will be reused?

pmderodat:
Hello @mosteo!

No problem, let’s discuss this in details when you’re back. :-) But quickly: I don’t think it’s necessary to merge this right away (especially given that this rewrites the index), but I just wanted to provide a proof of concept to discuss the evolution of the index format (going from code to data).

So I suggest the following course of action: first discuss the shape of the index I propose (https://github.com/alire-project/alire/blob/73ae648a50c4ad3da6dff8a85491c08c7ddd8120/doc/catalog-format-spec.rst), then elect a specific syntax to represent the index. Once this is done, I’ll be able to update this pull request so that it can be merged.

Thanks! :-)

doc/catalog-format-spec.rst
mosteo: What about https://en.wikipedia.org/wiki/Beerware

doc/catalog-format-spec.rst
mosteo: Even if GPR files cannot be directly reused, might the GPR syntax be reused/extended? I'm here a bit torn between something that is industry-standard (although I would go to e.g. YAML before JSON, for the friendliness) and something that's familiar to Adaists. Having to learn yet another format is another hill to climb, but since GPR is already in the mix...

doc/catalog-format-spec.rst
mosteo: I think @gusthoff refers to the project supplying a license file (as most do), which is referenced in the "custom" field. But then, you don't have the file until you pull the project. The possibility of supplying the license in the catalog folders is a bit strange, like a mix of scopes, but it has the advantage of making the license readily available in full.

If going for a external resource, what about it being a URL. It could be file:// for something relative to the alire file, or http[s]:// for something online, maybe even project:// for something in the pulled sources.

Even if the possibility of inlined licenses is kept, I think custom licenses should specify a short name that can be listed in package summaries. So custom could be a mapping with mandatory name (one-liner), optional location (url) and description (inline text).

mosteo:
I've been doing a quick read and wew!, that's quite the amount of work you've put in. Some quick thoughts:

- This pretty much is how I imagined things would work using textual description/parsing. I see it matches closely the features previously available through Ada specs, and the json parser in the end uses the previous Ada calls. 

- I'm unsure whether some features have been lost in translation? I've not seen any and/or mentions in the format description. I understand this is deliberate since the syntax is not final and it's not the point of the RFC to have feature parity.

I think your work properly demonstrates the way to go to have textual representation and enable multiple and private repositories. I know I sound like a broken record, but maybe the ability to use Ada for package descriptions may be kept in for people that prefer contributing that to text files, even if on-the-fly recompilation is totally removed. I open this topic for discussion in #43.

You suggest discussing the index and then selecting the syntax. I will do a more thorough read of everything, but I think your pull request achieves by far the purpose of showing how things would work (I still have to try your build but I don't expect surprises there?), and any changes can be discussed with the new syntax. So I see syntax selection as the immediate next step. Since syntax specifics are not the purpose of this RFC, I have created #42 for it with a summary of the points raised here.

doc/catalog-format-spec.rst
pmderodat: I’ve thought about it, and my current position is that, GPR files as they are today don’t allow us to express what we need (please refer to my long reply to Gustavo above ;-)).

If we start thinking that we could extend its syntax or its semantics, then what’s the point? This means we’ll come up with something necessarily new, and thus something that users won’t be familiar with, defeating the purpose of re-using GPR files. Besides, our needs here to extend GPR files don’t match the needs of other GPR-based tools (gprbuild, GPS, …). So that means we’ll introduce specificities to GPR files that other tools won’t use. At this point, why not roll our own DSL? :)

The index is really just data, and expect for conditional parts (OS-specific deps, for instance), this data has a straightforward shape. This is why I think that re-using an already existing generic data format is the way to go. I see the following requirements for this format:

1. Avoid as much as possible obscure formats or re-inventing one, i.e. avoid as much as possible to force people to learn something new (but not as far as re-using something that is not appropriate!)

2. Straightforward to parse (1/2), so we don’t spend too much time writing a parser (or even better, re-use an existing parser)

3. User-friendly, so that writing a correct package specification is not combursome while still unambiguous for human brains.

To me, 3 rules out both JSON (cumbersome to write by hand, especially the need to have commas between array and object items but trailing ones are not allowed) and YAML (the spec is huge, there are just too many ways to express the same thing).

While looking at what others do, I’ve noticed that [TOML](https://github.com/toml-lang/toml) was quite satisfactory in all points above: it’s designed to be easy to parse programmatically, has a big user base thanks to Rust’s Cargo and it’s also designed to be easy to write by hand.

doc/catalog-format-spec.rst
pmderodat: Good idea! It’s even shorter than the WTFPL, so I’ll update the example next time :)

doc/catalog-format-spec.rst
Fabien-Chouteau: @mosteo as discussed in #42, TOML is really simple and in my opinion it will be the easier to learn "TOML + Alire" vs "something close to GPR but not quite the same + Alire".

mosteo: All right. Indeed I'm not super-fond of the GPR Ada-but-not-Ada syntax.

src/alire-json_index.adb
```diff
@@ -0,0 +1,1871 @@
...
+         TIO.Open (File, TIO.In_File, Filename);
+      exception
+         when E : TIO.Use_Error | TIO.Name_Error =>
+            Set_Error (Result, Filename, Exc.Exception_Message (E));
+            return JSON;
+      end;
+
+      declare
+         Content     : Ada.Strings.Unbounded.Unbounded_String;
+         JSON_Result : J.Read_Result;
```
mosteo: I'm getting an error here that Read_Result is not declared in JSON. Also in line 248 about Format_Parsing_Error.
I'm on GNATColl 2018, just wanting to confirm if the problem is mine or these are forthcoming additions.

pmderodat: The only way the Community 2018 release of GNATCOLL.JSON handles parsing errors is to print them on the screen. This is in appropriate in most cases, so I used gnatcoll-core’s development branch, in which we included this more powerful error handling. :-) I should probably have mentionned that in the PR or the commit… sorry for the trouble!

doc/catalog-format-spec.rst
pmderodat: Good point, having a short name could be very useful.

I think it’s important to have the text of custom license to be part of the Catalog: we don’t control external URLs, and I think it’s important to guarantee that the license text is always available.

What about having a `license` directory at the root of the index, containing text files whose names correspond to the short names referenced from the package descriptions?

src/alire-json_index.adb
mosteo: OK, thanks!

doc/catalog-format-spec.rst
mosteo: Right. I saw that e.g. Cargo simply allows either a standard license name or a custom license in a file given as a relative path to the manifest file.
The whole issue is a bit mind-bending because one scenario is when you're working with the index (as you say having the license file is good, even before downloading the package), and another when you're developing a package (having the license in your project root folder is common practice, and you're not caring about an index at that time). 
At this point I think is enough to keep this in mind for when the final spec is taking shape.

doc/catalog-format-spec.rst
mosteo: I see some points of friction with autodetection of info from GPR files:
- It may be interesting to have this info in the index, even before a project is downloaded
- Not all the information may make sense to be made available to the final user of a project

That said, for a packager it may be very convenient to have these fields autopopulated by analysis of project files...

More thoughts: in the current Ada format one can distinguish public (user-intended) and private (developer-intended) properties. 
Another somewhat related aspect: in Cargo e.g. you can define build profiles.

pmderodat:
>     * This pretty much is how I imagined things would work using textual description/parsing.

Great :-)

>     * I'm unsure whether some features have been lost in translation?

I tried to keep feature parity, with several intended exceptions:

1. No arbitrary nesting of conditionals/properties. For instance, it’s not possible anymore to set one website or another depending on the current OS. To me, this restriction is appropriate as some properties such as the website, the authors or the license, should not vary depending on the environment.

2. No private/public distinction for properties. To me, the whole package description is to use the package: private stuff can stay in the project itself (`.gpr`), as I don’t expect developers of package X to use alire to build X (Alire would be only useful to get a build environment, but build would be controlled by the user).

Point 2 probably deserves its own discussion, though… doing it this way made my work on JSON specs easier.

> I know I sound like a broken record, but maybe the ability to use Ada for package descriptions may be kept in for people that prefer contributing that to text files, even if on-the-fly recompilation is totally removed. I open this topic for discussion in #43.

Looks like there are more cons that pros in #43. ;-) I don’t think having multiple formats for package descriptions is the way to go: twice the maintenance in tools, twice the amout of things to know for people dealing with specs, … If we manage to get a format that is very easy to use (and I hope we will!), then I think it’s worth allowing Ada descriptions. Anyway, I can followup on #43 for this point if you want.

> So I see syntax selection as the immediate next step. Since syntax specifics are not the purpose of this RFC, I have created #42 for it with a summary of the points raised here.

Ok thanks. :-)

mosteo:
> 1. No arbitrary nesting of conditionals/properties. For instance, it’s not possible anymore to set one website or another depending on the current OS. To me, this restriction is appropriate as some properties such as the website, the authors or the license, should not vary depending on the environment.

Understood; conditionals were intended mostly for dependencies/build properties anyway.

> 2. No private/public distinction for properties. To me, the whole package description is to use the package: private stuff can stay in the project itself (`.gpr`), as I don’t expect developers of package X to use alire to build X (Alire would be only useful to get a build environment, but build would be controlled by the user).
> 
> Point 2 probably deserves its own discussion, though… doing it this way made my work on JSON specs easier.

The point about keeping development things in gpr is a good one. We can revisit this later with the toml dsl.

> Looks like there are more cons that pros in #43. ;-) I don’t think having multiple formats for package descriptions is the way to go: twice the maintenance in tools, twice the amout of things to know for people dealing with specs, … If we manage to get a format that is very easy to use (and I hope we will!), then I think it’s worth allowing Ada descriptions. Anyway, I can followup on #43 for this point if you want.

Yeah, lots of cons :) You mean "it isn't worth"? I have thought a bit more about this, will drop a few thoughts in #43. We can keep this topic isolated there.

doc/catalog-format-spec.rst
onox: For licenses I would suggest to use [SPDX license identifiers](https://spdx.org/ids).

pmderodat: Good idea, thanks! :-) I’ll try to integrate SPDX identifiers in the next proposal.

pmderodat: For the record: a lot of things are pending the move of the index to data files and it will take some time to implement the handling of SPDX identifiers, so I’ll open an issue to remember to do it later. This way, this PR will not block further developments longer. :-)

pmderodat:
I forgot to mention: I haven’t translated to TOML the spec for simple_components because it’s quite lengthy (although not particularly complex): of course I will do it if we go further this road. :)

PR #56, opened 2019-09-17
Renamed former child crates, readded Simple Components

mosteo:
Updates to the index to take advantage of alire-project/alire#200.


mosteo:
Won't be finally used.

PR #59, opened 2019-09-26
Convert archive-hash to origin-hashes

mosteo:
Required by alire-project/alire#210


mosteo:
Since we are going to make hashes for git mandatory, this PR is obsolete.

PR #62, opened 2019-11-23
Register the 0.1 release of ada-toml

pmderodat:


index/ad/ada_toml.toml
```diff
@@ -0,0 +1,49 @@
...
+description = "TOML parser for Ada "
+
+authors = ["AdaCore",
+           "Pierre-Marie de Rodat <pmderodat@kawie.fr>"]
+maintainers = ["pmderodat@kawie.fr"]
+maintainers-logins = ["pmderodat"]
+
+project-files = ["ada_toml.gpr"]
+
+licenses = ["""custom:
```
Fabien-Chouteau: I think the license is supposed to be one of the following identifier: https://github.com/alire-project/alire/blob/master/src/alire/alire-licensing.ads

pmderodat: I see: will fix, thanks! We should probably update the catalog spec, then (look for the `licenses` entry in https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.rst#general-info). That reminds me of issue alire-project/alire#191 by the way.

Fabien-Chouteau: Right, so actually it is allowed to put the full license text until alire-project/alire#191 is implemented?

pmderodat: IIRC it has been allowed since the transition to TOML indexes. I think it will still be useful if/when alire-project/alire#191 is implemented as restricting licenses to a whitelist seems too restrictive.

Anyway, regarding this crate, I now know thanks to you that I can actually use `BSD 3-Clause`, so inlining the license text is not necessary. Thanks!

Fabien-Chouteau: > restricting licenses to a whitelist seems too restrictive.

On the other hand we have to look at what this field will be used for. In my opinion it is mostly a declarative piece of info for people to chose a crate. So I think it is better to limit the choice to SPDX identifiers plus an `other` or `custom` choice. And then people with `other` or `custom` license can document it on their website or description field.

pmderodat: Fine by me! I guess this discussion should be tracked in alire-project/alire#191 FTR I’ve updated this PR to use the `BSD 3-Clause` license in `ada_toml.toml`

mosteo: Also FTR, I think that custom licenses already are implemented as discussed (license is either from the whitelist or custom -- otherwise error). So alire-project/alire#191 is a matter of swapping the current list from github by the one from SPDX.

PR #65, opened 2019-11-24
Add GESTE and GESTE_Examples

Fabien-Chouteau:


index/ge/geste_examples.toml
```diff
@@ -0,0 +1,19 @@
...
+                 "RPG/RPG.gpr",
+                 "racing/racing.gpr"]
+
+licenses = ["BSD 3-Clauses"]
+
+['1.0.0']
+origin = 'https://github.com/Fabien-Chouteau/GESTE-examples/releases/download/v1.0.0/GESTE_Examples-1.0.0.zip'
+origin-hashes = ['sha512:2d94ac6de0a7d5c7087d46d3f276ed69e35b3b22b9322a7e53031ec77a18e46b90162372182401ee4078923ffefb884eb2f1c11a6c00c7727a65e7ee4fe6fe83']
+['1.0.0'.depends-on]
+geste = "1.0.0"
```
mosteo: I guess the use of version-exact dependencies here is intended?

mosteo:
The examples look gorgeous, BTW.

index/ge/geste_examples.toml
Fabien-Chouteau: I didn't think this through, in theory the examples should work on any GESTE 1.x.x.

Fabien-Chouteau:
> The examples look gorgeous, BTW.

Thanks to the authors of the sprites listed in the READMEs :)

index/ge/geste_examples.toml
mosteo: Actually this is default that makes me a bit uncomfortable: a lack of operator is interpreted as "=". Rust, for example, defaults to "^" ([here](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)). If we required an explicit operator, however, equility results in ugly `geste = "=1.0.0"` (this is already accepted btw).

PR #66, opened 2019-11-24
Start a workflow to check contributions

Fabien-Chouteau:


mosteo:
Hi Fabien, the choice of one repo per action is on purpose? I would tend to go to a single "actions" repo. Just curious.

Fabien-Chouteau:
I think that is how GitHub intend the actions to be presented. Now you can also have one action in each branch `alire-projects/actions@action_a`, `alire-projects/actions@action_b`.

What is special about the `check-author` action is that it requires the GitHub login of the contributor as well as the ability to get the old version of the file. This can only be done with the GitHub API. But for the other checks that we want to do, I think it will be better to have a script in this repository that is called from the workflow.

mosteo:
On the same page about the API needs. I was referring to the second syntax described [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#referencing-an-action-from-a-public-repository): 

> If an action is defined in a public repository, you must reference the action using the syntax {owner}/{repo}@{ref} or {owner}/{repo}/{path}@{ref}

Personally, using different branches for different actions is counterintuitive to me; I see it more suited for versioning.

PR #67, opened 2019-11-26
Fix Basalt license string

jklmnn:
I tried to use Basalt added in #61 but I noticed that the license string was not accepted by alire. I initially failed to notice that only certain license strings are allowed. Also the AGPLv3 is not explicitly stated in the [specification](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.rst) but seems to be available in the source code (at least I found it in `src/alire/alire-licensing.ads` when grepping for it).

This PR only fixes the license string ~~but I'd suggest to setup a CI that checks correspondence of package configs with the specifications so that this kind of error can be fixed early~~. Seems you already did that.

### DIFF
--- a/index/ba/basalt.toml
+++ b/index/ba/basalt.toml
@@ -6,5 +6,5 @@ origin-hashes = "sha512:715cb1b41be4425b90adc5404fd4193d97c74980245ef75b09f5986b
 description = "Collection of formally verified building blocks"
 maintainers = ["kliemann@componolit.com"]
 maintainers-logins = ["jklmnn"]
-licenses = ["AGPLv3"]
+licenses = ["AGPL 3.0"]
 


Fabien-Chouteau:
Thank you @jklmnn ,

> This PR only fixes the license string ~but I'd suggest to setup a CI that checks correspondence of package configs with the specifications so that this kind of error can be fixed early~. Seems you already did that.

We do want to have as many checks as possible for the index PR, but we barely started.

jklmnn:
The CI says that `general.maintainers_login` is missing although the spec (and all other configs I looked into) say it should be `maintainers-logins`.

Fabien-Chouteau:
> The CI says that `general.maintainers_login` is missing although the spec (and all other configs I looked into) say it should be `maintainers-logins`.

I am on it. `check-author` was developed as a proof-of-concept before the `maintainers-logins` field was added to Alire.

Fabien-Chouteau:
@jklmnn It should be fixed now. Can you try to re-trigger the checks, I don't have the permissions to do it from my account.

mosteo:
Sorry, the initial problem was my oversight.

I retriggered the checks, but then:
```
'mosteo' not in maintainers_login for 'index/ba/basalt.toml'
```

Fabien-Chouteau:
> I retriggered the checks, but then:
> 
> ```
> 'mosteo' not in maintainers_login for 'index/ba/basalt.toml'
> ```

Right, because you triggered the event your the "actor" in the context.
I will see if I can find a better way to identify the author of the PR.

Fabien-Chouteau:
@mosteo can you please try again :)

mosteo:
It worked :-)

mosteo:
I'm going ahead and merge, since we have the main index in bad shape otherwise. Lesson learned!

jklmnn:
Thank you for taking the time to make this working!

PR #68, opened 2019-11-26
Basalt: Fix hash array

mosteo:
Go figure, I merged again without checking and there was an error remaining. Third time is the charm, I hope.


mosteo:
And we have confirmation that the check flagged me as a non-maintainer. Merging anyway ;)

PR #73, opened 2019-12-14
Add Ada Utility Library definitions

stcarrez:
This pull request defines several crates to allow using some parts of Ada Utility Library
(https://github.com/stcarrez/ada-util).

I try to follow the Debian policy for the GNAT project names and Debian package names.
As such, I'm using the GNAT project name as much as possible as the crate name.

I've managed to avoid using the configure part of ada-util project. I've stored in the
ada-util project several GNAT project files that are almost ready to be used by alire.
The alr tool does a good job at guessing/setting up the platform dependencies that is required for
the utilada project. Other utilada projects don't have much problems.

The pull request adds the following projects:
```
lzmada                Ada LZMA Library Binding                                          
utilada               Ada Utility Library                                               
utilada_curl          Ada Utility Library (CURL)                                        
utilada_lzma          Ada Utility Library (LZMA)                                        
utilada_xml           Ada Utility Library (XML)  
```

### DIFF
--- a/index/lz/lzmada.toml
+++ b/index/lz/lzmada.toml
@@ -4,15 +4,12 @@ licenses = ["MIT"]
 maintainers = ["Stephane.Carrez@gmail.com"]
 maintainers-logins = ["stcarrez"]
 
-    project-files = ["lzmada.gpr"]
+    project-files = [".alire/lzmada.gpr"]
 
     [general.gpr-externals]
     LZMA_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
     BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
 
-    [[general.actions]]
-    type = "post-fetch"
-    command = ["cp", ".alire/lzmada.gpr", "lzmada.gpr"]
-
 ['1.1.0']
-origin = "git+https://github.com/stcarrez/ada-lzma.git@3e55cdadbb46ba1428ac52a00b403039bd0fcaaa"
+origin = "https://github.com/stcarrez/ada-lzma/archive/1.1.0.tar.gz"
+origin-hashes = ["sha512:1344289480d43e12622c3ae822d9b4a34d927219341eafff2f80151f2ee6613aac6599d3ee46f602869d1f498129fd0607b688951d6cbf97b24b48af2cb44047"]
--- a/index/lz/lzmada.toml
+++ b/index/lz/lzmada.toml
@@ -1,0 +1,18 @@
+[general]
+description = "Ada LZMA Library Binding"
+licenses = ["MIT"]
+maintainers = ["Stephane.Carrez@gmail.com"]
+maintainers-logins = ["stcarrez"]
+
+    project-files = ["lzmada.gpr"]
+
+    [general.gpr-externals]
+    LZMA_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
+    BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
+
+    [[general.actions]]
+    type = "post-fetch"
+    command = ["cp", ".alire/lzmada.gpr", "lzmada.gpr"]
+
+['1.1.0']
+origin = "git+https://github.com/stcarrez/ada-lzma.git@3e55cdadbb46ba1428ac52a00b403039bd0fcaaa"
--- a/index/ut/utilada_curl.toml
+++ b/index/ut/utilada_curl.toml
@@ -1,0 +1,23 @@
+[general]
+description = "Ada Utility Library (CURL)"
+licenses = ["Apache 2.0"]
+maintainers = ["Stephane.Carrez@gmail.com"]
+maintainers-logins = ["stcarrez"]
+
+    project-files = [
+        "utilada_curl.gpr"
+    ]
+
+    [general.gpr-externals]
+    UTIL_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
+    BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
+
+    [general.depends-on]
+    utilada = "^2.0.0"
+
+    [[general.actions]]
+    type = "post-fetch"
+    command = ["rm", "-f", "config.gpr", "utilada_core.gpr", "utilada_sys.gpr"]
+
+['2.0.0']
+origin = "git+https://github.com/stcarrez/ada-util.git@9254d4427f1b5498a1896ee146e3cca61a6913ee"
--- a/index/ut/utilada_curl.toml
+++ b/index/ut/utilada_curl.toml
@@ -1,5 +1,5 @@
 [general]
-description = "Ada Utility Library (CURL)"
+description = "Utility Library REST support on top of CURL"
 licenses = ["Apache 2.0"]
 maintainers = ["Stephane.Carrez@gmail.com"]
 maintainers-logins = ["stcarrez"]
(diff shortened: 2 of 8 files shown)


index/lz/lzmada.toml
```diff
@@ -0,0 +1,18 @@
...
+
+    project-files = ["lzmada.gpr"]
+
+    [general.gpr-externals]
+    LZMA_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
+    BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
+
+    [[general.actions]]
+    type = "post-fetch"
+    command = ["cp", ".alire/lzmada.gpr", "lzmada.gpr"]
```
Fabien-Chouteau: Instead of copying the project file, you should be able to use `.alire/lzmada.gpr` in the field `project-files` above.

One the problem that we will have to consider with Actions is the cross-platform compatibility, `cp` is unlikely to work on Windows.

I think this applies to all the crates you submitted.

index/lz/lzmada.toml
```diff
@@ -0,0 +1,18 @@
...
+
+['1.1.0']
+origin = "git+https://github.com/stcarrez/ada-lzma.git@3e55cdadbb46ba1428ac52a00b403039bd0fcaaa"
```
Fabien-Chouteau: For security reasons we decided to forbid VCS origins from the community index. The rational is that we could not find a good way to have a hash to identify the content (discussion here: https://github.com/alire-project/alire/issues/198).

What I did for my project is a combination of `git archive`:
```
$ git archive --format=zip -o ada_voxel_space_demo-1.0.0.zip --prefix=ada_voxel_space_demo-1.0.0/ HEAD
```

Then publish the archive using GitHub release.

And then get the origin hash with `alr publish`:
```
$ alr publish --hash https://github.com/Fabien-Chouteau/ada-voxel-space-demo/releases/download/v1.0.0/ada_voxel_space_demo-1.0.0.zip
```

Then my crate release is:
```
["1.0.0"]
origin = "https://github.com/Fabien-Chouteau/ada-voxel-space-demo/releases/download/v1.0.0/ada_voxel_space_demo-1.0.0.zip"
origin-hashes = ["sha512:b7a0cf7cb5c740994bc523a40d7d51838bdbdbf66a4e15e04a90ee42251f13bf42d268b485c84e434808530f7166ba6d5617033e4942695ace448d787b18641c"]
```
Hopefully we can get the `alr publish` command to do that automatically someday.

index/ut/utilada_curl.toml
```diff
@@ -0,0 +1,23 @@
...
+    [general.gpr-externals]
+    UTIL_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
+    BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
+
+    [general.depends-on]
+    utilada = "^2.0.0"
+
+    [[general.actions]]
+    type = "post-fetch"
+    command = ["rm", "-f", "config.gpr", "utilada_core.gpr", "utilada_sys.gpr"]
```
Fabien-Chouteau: I wonder why you are removing those project files.

If possible it would be better to do that before packaging (see comment above) so that there is no action to do in Alire.

Which by the way makes me think that if we have a publish command someday, we could support `pre-publish` actions. (@mosteo)

Fabien-Chouteau (changes requested):
It's great to see all you libraries coming to Alire :)

I have a couple of comments and questions.
The one about VCS origin is blocking for now because we decided to forbid it from the community index.

Thanks,

index/ut/utilada_curl.toml
stcarrez: These files are necessary to build utilada but they enter in conflict when building utilada_curl.
Both utilada and utilada_curl point to the same git repository or git-archive. I don't want utilada to install the utilada_curl part and I've setup a separate Alire project for that. To build the utilada_curl it depends on utilada which brings the following GNAT projects

- config.gpr
- utilada_core.gpr
- utilada_base.gpr
- utilada_sys.gpr

For example, when I use several of the projects I've submitted, the Alire cache contains:
```
ciceron@zebulon:~/work/awa/alr-test/tst/testing$ ls -l alire/cache/projects/
total 28
drwxrwxr-x 12 ciceron ciceron 4096 déc.  14 21:40 ado_2.0.0_3de2eb8b
drwxrwxr-x 12 ciceron ciceron 4096 déc.  14 21:40 ado_postgresql_2.0.0_3de2eb8b
drwxrwxr-x 12 ciceron ciceron 4096 déc.  14 21:40 ado_sqlite_2.0.0_3de2eb8b
drwxrwxr-x 12 ciceron ciceron 4096 déc.  14 22:44 keystoreada_1.0.0_ba64f6e9
drwxrwxr-x 12 ciceron ciceron 4096 déc.  14 21:40 utilada_2.0.0_9254d442
drwxrwxr-x 12 ciceron ciceron 4096 déc.  14 21:40 utilada_xml_2.0.0_9254d442
drwxrwxr-x  7 ciceron ciceron 4096 déc.  14 22:30 wikiada_1.2.0_1705a18e
```

And utilada_* comes from the same repository/git-archive, likewise for ado_*.

I have similar issue on ado_ and its even more annoying because I don't want
to force users to have to install SQLite, MySQL, PostgreSQL. This way, there is
a clear separation. I can't split my project in separate git-archives because it would
be too complex for me.

Fabien-Chouteau (approved):
Thanks Stephane

index/ut/utilada_curl.toml
mosteo: Historical note: this is what was handled previously by child projects (a single checkout was shared  by several crates with possibly different dependencies, alire-project/alire#200). I think our current policy is that the separation should be managed with different upstream repositories.

mosteo:
@stcarrez , my understanding is that the utilada_aws crate depends on AWS but that dependency is not expressed (since AWS is not packaged in alire yet). Am I right about this?

stcarrez:
> @stcarrez , my understanding is that the utilada_aws crate depends on AWS but that dependency is not expressed (since AWS is not packaged in alire yet). Am I right about this?

@mosteo yes this is correct. OTOH it is packaged with the GNAT 2019 community compiler.
As soon as we are able to package AWS in Alire, we can add the dependency :-)

index/ut/utilada_curl.toml
stcarrez: @mosteo don't put constraints on upstream projects. There are many reasons why such upstream separation may not be possible.

mosteo:
I was asking because I don't like the idea of having hidden dependencies. I'm looking at how we could, with what we already have, express this explicitly, and I'll come back to you.

OTOH, the duplicity of packaged in Alire/bundled with GNAT libs is something I've wondered whether can bite us somehow.

Fabien-Chouteau:
I think we are good to go on this PR. @mosteo ?

mosteo:
@Fabien-Chouteau As-is this will fail with any compiler that doesn't bundle `aws`, right?. Up to now I have tried to avoid such conditionally failing crates.

For the short term, this crate could include an `available` for those compilers or, better yet, simply depend on the appropriate `aws` versions. If the latter, for the time being we could introduce an unavailable native `aws` so it is reported as a hint without further consequences (e.g. like [this](https://github.com/mosteo/alire-index/blob/feat/template-unavailable/templates/unavailable.toml)).

index/ut/utilada_curl.toml
mosteo: Sorry, @stcarrez, I somehow missed your last reply. I agree, and there were some features about that in the past that are now missing. In some cases they were half-backed (the one already mentioned), others it's a matter of reintroducing the syntax since the feature is already in the code (e.g. alire-project/alire#105).

Fabien-Chouteau:
I prefer to have those crates in the index even if there is something to install on the side, than not have them at all.

mosteo:
I understand, what I'm proposing is a minor change so the dependency on `aws` is already correct for the future and in the meantime the normal hinting will happen.

It's a matter of what you feel is more urgent: merging this as-is or wait for the addition of the `depends-on` in this patch by @stcarrez. Meanwhile I can prepare the `aws` native crate.

mosteo:
I pushed by mistake my `aws` crate to the current index devel branch (I though this was impossible -- will check the permissions) when going to prepare a PR. Anyway -- the crate is already there :S so we may keep it or revert the change.

mosteo:
Also I've created alire-project/alire#298 with my thoughts on this. After that I don't see this matter so clear cut, so I'm going ahead with merging this one (sorry for the delay!) and I'll appreciate your feedback in alire-project/alire#298.

mosteo:
The check failure is expected since I'm not a maintainer... the same will happen for the rest of the pending PRs for which I have to fix the index version. Unrelated, when I try to build this I get this error (more evidence to move away from aggregate projects):

```
==============Error messages for file: /tmp/a/utilada_2.0.0_5e0a1f54/alire/alr_build.gpr
     1. aggregate project Alr_Build is
                          |
        >>> cannot aggregate abstract project "UtilAda_Conf"
```

mosteo:
Merged via #82

PR #74, opened 2019-12-14
Add Ada Database Objects 2.0.0 crates

stcarrez:
This pull request adds the Ada Database Objects library available at https://github.com/stcarrez/ada-ado.

- the 'ado' crate is the core library and does not contain any database driver
- the 'ado_postgresql' crate depends on 'ado' and is the PostgreSQL driver
- the 'ado_sqlite' crate depends on 'ado' and is the SQLite driver

There is missing the 'ado_mysql' crate because the library dependency
is tricky: we may need one of libmysqlclient or libmariadb libaries.
This is supposed to be identified during the configure step of the library.

However, the crate don't run the configure script that are required traditionally by the source library.
Running configure is a real pain and until now I've successfully managed to avoid it...

### DIFF
--- a/index/ad/ado.toml
+++ b/index/ad/ado.toml
@@ -1,0 +1,29 @@
+[general]
+description = "Ada Database Objects (Core library)"
+licenses = ["Apache 2.0"]
+maintainers = ["Stephane.Carrez@gmail.com"]
+maintainers-logins = ["stcarrez"]
+
+    project-files = [
+        "ado.gpr"
+    ]
+
+    [general.gpr-externals]
+    ADO_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
+    BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
+
+    [general.depends-on]
+    utilada = "^2.0.0"
+    utilada_xml = "^2.0.0"
+
+    [[general.actions]]
+    type = "post-fetch"
+    command = ["rm", "-f", "config.gpr"]
+
+    [[general.actions]]
+    type = "post-fetch"
+    command = ["cp", ".alire/ado.gpr", "ado.gpr"]
+
+['2.0.0']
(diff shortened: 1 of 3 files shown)


Fabien-Chouteau (changes requested):
Same comments as #73 .

I am blocking the PR because of the VCS origins (alire-project/alire#198).

Fabien-Chouteau (approved):

mosteo:
> we may need one of libmysqlclient or libmariadb libaries.

In the past this was supported by `alr`: alire-project/alr#106. Just mentioning it as supporting evidence if we decide to reintroduce it.

stcarrez:
> > we may need one of libmysqlclient or libmariadb libaries.
> 
> In the past this was supported by `alr`: [alire-project/alire#106](https://github.com/alire-project/alire/issues/106). Just mentioning it as supporting evidence if we decide to reintroduce it.

I'm not sure this is the role of Alire to handle system library installation and dependencies.
Of course this is nice but it may be difficult to do it right for Ubuntu, Debian, Fedora, ArchLinux, FreeBSD, NetBSD, OpenBSD, Windows etc...

Hum... Looking at the system library dependencies, it seems that it's only supported for Debian|Ubuntu.
So, there is the issue on Windows and MacOS.

May be it's acceptable, in a first step, to skip native dependencies after all.

The issue that remains on my side is identifying the correct library options.
May be a light configure script could do it...

mosteo:
I was referring more to the issue of declaring two similarly valid dependencies where either of them is acceptable but only one is needed; this used to be possible but can't be done with the current index syntax, although the internal logic remains. We are not sure how much of a corner case this is.

On the system dependencies issue much ink has already run and the current impl has known shortcomings (see e.g. alire-index/alire#204). Our current intention is that it be a "nice to have" feature but transparent otherwise.

mosteo:
Merged via #82

PR #76, opened 2019-12-14
Add Ada Keystore 1.0.0 crate

stcarrez:
The library provides a secure keystore where documents, configuration files
can be stored.  The crate depends on 'utilada'.


Fabien-Chouteau (changes requested):
Same comments as #73 .

I am blocking the PR because of the VCS origins (alire-project/alire#198).

mosteo:
I think we may be waiting on each other :) @Fabien-Chouteau, while testing the origin hash for this PR I realized it uses a `git+https` origin, so no hash needed. You told me something about allowing again this kind of origins? Could we then move forward with this one as is?

stcarrez:
Closing because I've submitted a new pull request for Ada Keystore 1.1.0

PR #77, opened 2019-12-15
Add eagle_lander simulator

Fabien-Chouteau:


mosteo:
@Fabien-Chouteau, do you mind removing the crate from the `orphaned` folder as part of this PR?

mosteo:
Thanks! It seems the error is due to the new case of removing a file. I'm going to merge later unless you want to use this as test bed somehow.

PR #84, opened 2020-01-11
Add Matreshka League 18.1 crate

reznikmm:


mosteo:
Hello @reznikmm, thanks for your contribution. The crate looks and tests good as-is, but before merging it I wonder if you'd like to add some tags for classification. It should be a vector with keywords in the general section:

```
tags = ["spark", "security"]
```

Just let me know if you don't want to do it at this time so I'll merge right away.

reznikmm:
@mosteo I've added some tags. This is the first (root) crate. I'll push others (sub-)crates after merging, because they depend on this.

mosteo:
@reznikmm Merged, thanks.

PR #85, opened 2020-01-13
Add Matreshka XML, SQL, Servlet and FastCGI parts.

reznikmm:


Fabien-Chouteau:
So @reznikmm, those crates only contain a project file that withs one of the project files in the main crate?

reznikmm:
@Fabien-Chouteau yes, they are

Fabien-Chouteau:
Alright thank you, I wanted to point it out to make sure @mosteo have a look at this and give us his opinion.

mosteo:
@reznikmm, is this use of "shell" crates intended to allow better dependency separation in the future? Or do you see any other advantage over simply having all project files in a single crate?

reznikmm:
@mosteo Matreshka is distributed as single file archive. The archive contains many projects (libraries and executables). For instance Fedora builds many RPMs from a single matreshka SRPM. I want to have many crates for matrreshka, because some of them requires extra dependencies. For instance Oracle database driver needs corresponding Oracle client. Having just one crate for whole matreshka will require a lot of dependencies, that is difficult to install and makes no sense.

So my idea - sub-crates will just include a project file to build sub-library, while all source files are located in root crate - matreshka_league.

What do you think?

mosteo:
Ah, so the idea is to segregate the dependencies you need for matreshka itself. I guess that makes sense. It was actually one of the reasons behind the failed child crates idea (alire-project/alire#108).

I understand it comes at the cost of maintaining the extra separate repositories and project files, right?

The main issue I see is that this touches on an unwritten assumption of mine: that the `project-files` of a crate are the ones that the user of the crate has to "with". In this situation there are both `build_matreshka_fastcgi.gpr` (explicit in this subcrate) and `matreshka_fastcgi.gpr` (hidden in the main crate). 

So, is your idea that the user has to "with" the former or the latter? The former is ugly, the latter is hidden. And asking to reverse their places is asking for a change to upstream which I don't find reasonable (neither having to create new repositories to maintain).

This makes me think again that we need a way to reuse sources in several crates. This way there would be no need to maintain separate subcrate tarballs, and different dependencies per-crate could be achieved. It maybe a simple matter of having a new origin type for subcrates, where nothing is downloaded, but that informs us that project file paths are relative to another crate root.

reznikmm:
@mosteo "New origin" approach looks fine for me.

The user should definitely do `with "matreshka_fastcgi.gpr";` 

Currently the usage of separate repository for tarballs with auxiliary project file is the only way I found to make alire build sub-crates. We can go this way for now and rewrite crates latter if we have other approach.

mosteo:
Understood, @reznikmm. I'll ping you for testing whenever I get to it. Meanwhile I'll merge this one as-is shortly as we have no better alternative for now. And thanks for the feedback.

PR #86, opened 2020-01-19
Various embedded and video-game related crates

Fabien-Chouteau:


index/co/cortex_m.toml
```diff
@@ -0,0 +1,17 @@
+[general]
+description = "Drivers for Cortex-M micro-controllers (NVIC, SysTick, etc.)"
+maintainers = ["chouteau@adacore.com"]
+maintainers-logins = ["Fabien-Chouteau"]
+licenses = ["BSD 3-Clauses"]
+tags = ["embedded", "arm", "nostd"]
+
+project-files = ["cortex_m0.gpr",
```
mosteo, line 18: I'm thinking about how this affects our nightly build tests, which I guess will simply fail? ~~Should we add some `target` variable to the index syntax so we can skip attempting to test anything not native?~~

Fabien-Chouteau: We could install cross tool-chains for the nightly builds.

mosteo: What I'm trying to understand is, each of these project files is for a particular target, right? So we might need something like (NB `target` does not exist now):
```
[project-files.'case(target)']
cross-platform = "common.gpr" # if this made sense
cortex_m0 = "cortex_m0.gpr"
cortex_m4 = "cortex_m4.gpr"
# etc
```
to ensure that only approppriate project files are attempted (when we have a way to specify cross-targets, anyway).

Maybe you can give me a 15' intensive primer on GNAT cross-compilation at FOSDEM :)

mosteo: Since this is for the future anyway, I'm going to merge this as-is for now.

Fabien-Chouteau: The target is already specified in the project file, gprbuild will know which compiler to use.

And I think it is possible to have case statements in the project file that will select the sources based on the target. So I would say that this is covered in gprbuild.

One thing we could consider is declaring availability of crates depending on the target and run-time. I don't think this is top priority though.

mosteo: All right, thanks for the explanations.

Fabien-Chouteau:
Thanks

PR #90, opened 2020-02-03
Add Saatana v2.0.0 (formerly known as 'security' by HeisenbugLtd)

Jellix:
Hope I got it all fixed now. :)


pmderodat:
Hello @Jellix,

Thanks for your contribution! SHA256 indeed isn’t supported: see the documentation for `origin-hashes` in <https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.rst>.

While trying to build your project using my system’s GCC (9.2.0 release from the FSF), I got a compilation error:
```
gcc: error: unrecognized command line option ‘-fcallgraph-info=su,da’
```

You could specify that your project requires GNAT Community, but I guess it isn’t satisfactory, since one probably does not need the callgraph info in orde to use the Security project. :-) So I think the best option is to introduce some kind of build mode GPR variable (for instance `SECURITY_BUILD_MODE`) to control whether to use this option. You would then need to publish a corrective release (1.0.1?) to include it in Alire.

One last thing: the crate name contains a dot character. I thought that Alire forbids that: `alr help identifiers` also thinks so:
```
$ alr help identifiers

IDENTIFIERS

 Identifiers for crates and indexes must use lowercase alphanumeric characters 
 from the latin alphabet. Underscores can also be used except as the first 
 character.

 Length must be of 3 to 64 characters.
```
And so does the function that checks crate name validity (Alire.Is_Valid_Name). @mosteo does that ring a bell? Assuming that it should be forbidden, I would suggest to call this spec `heisenbug_security` instead to keep “generic” words first, similar to the `ado_*`, `matreshka_*` and `gnatcoll_*` crates. Thoughts?

mosteo:
Yup, dots are (or should be) rejected because they where reserved for child
crates, which we don't allow at the moment. I'm traveling until Friday,
I'll take a closer look then.

Assuming the crate has been never merged, I guess fixes can still be done
for the 1.0 release (unless upstream is already published elsewhere with
that version, to avoid confusion).

Fabien-Chouteau:
> Yup, dots are (or should be) rejected because they where reserved for child
crates,

`alr list` did not fail so it looks like they are not rejected.

Jellix:
> While trying to build your project using my system’s GCC (9.2.0 release from the FSF), I got a compilation error:
> 
> ```
> gcc: error: unrecognized command line option ‘-fcallgraph-info=su,da’
> ```

That is weird, according to:

https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gnat_ugn/Alphabetical-List-of-All-Switches.html#Alphabetical-List-of-All-Switches

the switch does exist. But as you said, it's not really needed, so I removed it. I haven't used any current FSF GNAT compiler, so I am not sure if there are any other incompatibilities. The minimum requirement is an Ada2012 compiler which understands SPARK annotations.

Regarding the other issues, they should all be fixed now.

pmderodat:
I just checked the commit history: this option has been documented in the GNAT UG at the FSF years before it was actually committed in the GCC tree (last November). So it will be available in the GCC 10 releases from the FSF. :-) Thanks a lot for your updates!

index/he/heisenbug_security.toml
```diff
@@ -0,0 +1,34 @@
...
+## at the root directory, then this property can be omitted.
+project-files = ["security.gpr"]
+
+## Keywords for classification
+tags = ["cryptography", "spark"]
+
+## A website with further information
+website = "https://github.com/HeisenbugLtd/security"
+
+['1.0.0']
```
mosteo: @Jellix IIUC, the 1.0.0 release is in practice broken with some compilers, right? In that case I think it might be better to remove it entirely since there's no risk yet that someone is using it already.

mosteo (changes requested):

Jellix:
Yep, it's broken. For every compiler. Apparently I forgot to push necessary changes before preparing the version. All fixed in 1.0.1, so I removed the whole entry for 1.0.0.

mosteo (approved):
Tested and working, thanks for your contribution.

PR #92, opened 2020-02-09
Add Ada Keystore 1.1.0

stcarrez:
This is version 1.1.0 for Ada Keystore.

I've added tags and descriptions as well as converted to use the hash signature.
I've dropped 1.0.0 because it's not worth to have it.


mosteo:
Merged, thanks.

PR #103, opened 2020-04-04
Deal with externals in crate builder check

mosteo:
New checks to avoid attempting to build the crate when 

- crate has an undetected system dependency
   - since build would likely fail
- crate itself is external and undetected


mosteo:
I'm sorry, I forgot I wanted to add this commit before requesting your review.

Fabien-Chouteau (approved):

PR #104, opened 2020-04-04
Update gtkada for Debian Stable

mosteo:


mosteo:
"Check PR" seems to be confused by the new external syntax.

mosteo:
> "Check PR" seems to be confused by the new external syntax.

Actually the problem is some exotic table syntax I used that ada-toml accepts but is rejected by other parsers ([SO question](https://stackoverflow.com/questions/61080922/toml-multi-level-table-syntax-inside-array-illegal-or-not)). 

I've fixed it but since the check is against the base ref that already contains the offending syntax it won't pass the check, so merging anyway. A few more crates will need this fix.

PR #105, opened 2020-04-07
Enable sdlada on MacOS

mosteo:


mosteo:
Not building on windows/macos since it depends on libsdl external. Still, this way it doesn't cause failure on dependent crates (they'll be notified of the needed externals).

PR #108, opened 2020-04-21
Fix obsolete AdaCore CDN links in asis, gnatcoll

### LINKED ISSUE #316: Review AdaCore CDN URLs
From gitter, reported by @reznikmm:

> @mosteo do you know that AdaCore has changed download site and URLs like http://mirrors.cdn.adacore.com/art/5b0819e0c7a447df26c27ab8 don't work any more? :(

mosteo:
Fixes alire-project/alire#316

### DIFF
--- a/index/as/asis.toml
+++ b/index/as/asis.toml
@@ -14,18 +14,26 @@ project-files = ["asis.gpr"]
     [general.depends-on]
     make = "any"
 
+[2019]
+origin = "https://community.download.adacore.com/v1/52c69e7295dc301ce670334f8150193ecbec580d?filename="
+archive-name = "asis-2019-20190517-18AB5-src.tar.gz"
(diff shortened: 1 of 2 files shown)


mosteo:
The ownership error is expected because pmderodat is listed as maintainer of gnatcoll.

The asis build on Windows seems to be hitting the problem with `tar`. I cannot reproduce it in my Windows VM. Any wisdom to share, @Fabien-Chouteau ?

Since these asis crates depend on the exact GNAT CE version that already bundles them, perhaps is moot having these definitions.

Fabien-Chouteau:
> The asis build on Windows seems to be hitting the problem with `tar`. I cannot reproduce it in my Windows VM. Any wisdom to share, @Fabien-Chouteau ?

Right, we have to do the same trick as the `Alire` CI and force the install of `tar` from `msys2`...
This is annoying.

mosteo:
> Right, we have to do the same trick as the `Alire` CI and force the install of `tar` from `msys2`

Ah, I was under the impression this was always happening on Windows. I'll take a look, thanks.

Fabien-Chouteau (approved):

PR #110, opened 2020-05-01
New releases of Ada Util, Ada EL, Ada Wiki, ADO, Ada Servlet, Ada Server Faces

stcarrez:
Today I've created a new release of all my projects.

This pull request updates the crate to take into account these new releases.

I also took the opportunity to provide some longdescription and tags in various crates.

### DIFF
--- a/index/ad/ado.toml
+++ b/index/ad/ado.toml
@@ -1,8 +1,46 @@
 [general]
 description = "Ada Database Objects (Core library)"
+tags = ["database", "uml", "sql"]
+
 licenses = ["Apache 2.0"]
 maintainers = ["Stephane.Carrez@gmail.com"]
 maintainers-logins = ["stcarrez"]
+
+long-description = """
+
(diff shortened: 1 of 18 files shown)


Fabien-Chouteau:
> I also took the opportunity to provide some longdescription and tags in various crates.

Thanks @stcarrez, this will look good on the [alire.ada.dev](alire.ada.dev) website :)

mosteo:
Hi @stcarrez . I'm sorry for the long overdue update. 

We have now official source releases for many AdaCore crates in the index, including `xmlada` which I think was at the root of your problems. 

In the interim, the manifest format has been updated and we have now one file per release. You can check your own crates in the current stable index branch, `stable-0.4`, as they were migrated automatically by ourselves. 

Sadly this means that your pending PRs will require migration too.

We also have a new `alr publish` that can help to prepare a manifest for the PR. The details on what to do depending on your upstream methodology are here: https://alire.ada.dev/docs/#publishing-your-projects-in-alire

Let me know if you find any problems.

stcarrez:
Dropping this pull request because I've made a new one after a new release and the old format is no longer supported.
The releases provided here will not be available as crate: users will have to use the newer version.

PR #111, opened 2020-05-01
Define crates for Ada Web Application

stcarrez:
I'm doing release 2.0 of AWA and take the opportunity to provide crates for Alire integration.


Fabien-Chouteau:
@stcarrez AWS is available in Alire now. Are you able to try it with your projects?

stcarrez:
I will not update the crate for this old AWA release.
Closing because I'll provide another pull request for next AWA rlease.

PR #114, opened 2020-05-10
Add socketcan

glencornell:


glencornell:
Forgive my being new to creating crates.  It looks like I might be missing two things in my socketcan.toml file:

1. A pointer to my gpr file.  it should have something like "project-files = ["src/socketcan.gpr"]".

2.  Since this is a Linux-only feature, I need to limit the platforms that this may be compiled upon.  Again, in my socketcan.toml file, I should have:

    [general.available.'case(os)']
    linux       = true
    macos    = false
    windows = false
    '...'          = false

Am I correct here?
thanks,
glen

mosteo:
Hi Glen, thanks for your contribution.

Correct on both accounts. For the OS availability, you can abbreviate it like this:
```
[general.available.'case(os)']
linux = true
'...' = false
```
since the '...' will already cover the other OSes.

Looking forward to your crate successfully building in Alire :-)

mosteo:
Looks good now, thanks.

Fabien-Chouteau:
Nice project @glencornell :+1: 

Don't hesitate to use the `tags` and `long-decription` fields to make your crate easier to find on the website and provide more info.
e.g. : https://alire.ada.dev/crates/geste
https://github.com/alire-project/alire-index/blob/devel-0.2/index/ge/geste.toml

PR #116, opened 2020-05-18
Fix build matrix excluded configuration

mosteo:


mosteo:
Change properly tested with an actual crate here: https://github.com/mosteo/alire-index/pull/9

PR #119, opened 2020-05-22
Add "Google Protocol Buffers implementation in Ada"

reznikmm:


mosteo:
Hi Maxim. I see that the declared project file in this crate is "compiler.gpr". Isn't that a bit unspecific? I'm worried about the global namespace that all projects in a build share.

Although I'm thinking that if this is a final app not used like a library then that's no problem.

reznikmm:
Good point, let me rename project file to match executable name

reznikmm:
Fixed. BTW do you have a plan to add executable to the PATH somehow? Or I shal do this in the crate? How can i do it?

mosteo:
Hi Maxim. Excuse me but I still see the same project file in the last commit.

> BTW do you have a plan to add executable to the PATH somehow? Or shall I do this in the crate? How can i do it?

There is ongoing work about this: https://github.com/alire-project/alire/issues/345. The first half about loading such variables was merged here: https://github.com/alire-project/alire/pull/389. Their actual use is ongoing work.

So, now that I think about it, you could already define the path modification in your crate and it should pass the check.

reznikmm:
> Excuse me but I still see the same project file in the last commit.

Indeed, my fault. Should be fixed now

>  you could already define the path modification in your crate and it should pass the check.

I've tried to add this:
```
    [general.environment]
    PATH.prepend = "${CRATE_ROOT}/.objs/compiler"
```
But it sets `PATH` only during the building. But I meant that `alr` should change `PATH` for the user of the crate, not for the crate itself. So after installing `protobuf` the user has `protoc_gen_ada` accessible through the `PATH` and can launch `protoc` without a manual modification of `PATH` (for instance as result of execution of `alr setenv`).

Fabien-Chouteau:
>  So after installing `protobuf` the user has `protoc_gen_ada` accessible through the `PATH` and can launch `protoc` without a manual modification of `PATH`

Alire doesn't handle installation of crates yet.

mosteo:
> But it sets `PATH` only during the building.

Actually that's unimplemented yet, unless I'm missing something.

> So after installing protobuf the user has protoc_gen_ada accessible through the PATH and can launch protoc without a manual modification of PATH

As Fabien says, there's no installation step for crates right now. However, the idea is that you can use `alr setenv` to use everything that's a dependency in a workspace, and that would include these paths. And, for crates depending on `protobuf`, the executable should be automatically available during their own build (during `alr build`/`alr get`)

reznikmm:
Ok. I've added this (PATH.prepend) just in case.

From my point of view `alr` should export such environment variables in `alr setenv` output together with ADA_PROJECT_PATH for project files.

Fabien-Chouteau:
> From my point of view `alr` should export such environment variables in `alr setenv` output together with ADA_PROJECT_PATH for project files.

It will, this is on my plate and it is not implemented yet.

PR #122, opened 2020-05-29
* Saatana: Added release v2.0.2, and updated project website.

Jellix:
Updated with current release and updated the project website.

### DIFF
--- a/index/sa/saatana.toml
+++ b/index/sa/saatana.toml
@@ -36,4 +36,4 @@ origin-hashes = ["sha512:1e0eaf5687cf51a76d9491fa45bb3ba91ed996684ce7678e05b01b2
 notes = "This version is deprecated. An error in the proof annotations has been identified which invalidates the proof of absence of runtime errors."
 
 ['2.0.2'.depends-on]
-gnat = "gnat^2020"
+gnat = "^2020"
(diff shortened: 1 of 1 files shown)


mosteo:
Hi @Jellix, I see you're removing the 2.0.0 release. I think that, barring a strong reason, the idea is not to remove older releases just in case someone is using them and doesn't want to upgrade.

Although, since we're still in beta, maybe it's acceptable. I see the [policy](https://alire.ada.dev/docs/#release-immutability) talks about withdrawing releases without going into much detail. Perhaps we should instead document them (using the `notes`?) with the reason to discourage use? Or even an `unsafe` flag (I'm thinking of identified vulnerabilities, for example) so the solver doesn't use them by default.

@Fabien-Chouteau, any thoughts on this?

Fabien-Chouteau:
A release should never be removed, some people may depend on it and it will break their builds.

mosteo:
OK, then we must update the policy to remove the withdrawing remarks.

Jellix:
Ok then. I reverted the removal of the 2.0.0 release, but added in the notes field that this version should not be used anymore.

mosteo:
Hi again, Vinzent. The checks show green but the build is failing for FSF 7 compilers, I guess because some attributes/aspects are too recent. To be completely in the clear, you need to add a dependency on the valid gnat versions (from my tests, that's `gnat^2020` or maybe `gnat^2019`). 

Meanwhile I am going to fix the slip in checking.

Jellix:
Not sure if I'm doing something wrong... I think, I fixed the syntax errors ALR was complaining about, but even with the updated version.

```
['2.0.2'.depends-on]
gnat = "^2020"
```

I do not see any 2020 build in the actions... What am I missing?

mosteo:
I see it as expected: testing combinations that use the CE compiler (windows, macOS, CentOS, Ubuntu+CE) pass the build while the rest skip it due to missing dependencies. So to me it looks all right now.

PR #123, opened 2020-05-31
Add zeromq-ada 4.1.5

reznikmm:


mosteo:
Hi Vincent. You mention python would be needed for proper configuring here. I understand that you circumvent the need to gain availability in more platforms, instead of adding Python as a system dependency?

Or, in other words, do you think that Alire should add system crates for python2/3?

mosteo:
A final note before merging. Since you rely on `/bin/sh` and `cp` that makes it platform dependent (or non-windows dependent?). Given the dependency on libzmq I guess that implicitly there's no problem; the crate will only resolve in supported distros.

However this makes me wonder about a few points: 

1. Will `/bin/sh` work in a Windows with msys2? My guess is: no, because of the absolute path.
1. I guess we should add a system crate for `coreutils`.
1. We need a section on cross-platform packaging in the docs.

@Fabien-Chouteau , can you confirm about 1)?

Fabien-Chouteau:
>     1. Will `/bin/sh` work in a Windows with msys2? My guess is: no, because of the absolute path.

It will not work indeed.

@reznikmm why `"/bin/sh", "-c", "cp libzmq.gpr.in libzmq.gpr; cp libsodium.gpr.in libsodium.gpr" ` and not `["cp", "libzmq.gpr.in", "libzmq.gpr"], ["cp", "libsodium.gpr.in", "libsodium.gpr"]`?

reznikmm:
> @reznikmm why `"/bin/sh", "-c", "cp libzmq.gpr.in libzmq.gpr; cp libsodium.gpr.in libsodium.gpr" ` and not `["cp", "libzmq.gpr.in", "libzmq.gpr"], ["cp", "libsodium.gpr.in", "libsodium.gpr"]`?

This is the simplest way I found to make it work with alire :smile: 

If my understanding is correct, the author provides a dummy configurator in Python. It doesn't fix anything in `libzmq.gpr.in`, `libsodium.gpr.in`. Both templates are just copied into corresponding `.gpr` file. 

`/bin/sh` could work in `msys2` if alr calls the Bash as shell, (not `cmd.exe` or PowerShell). I'm sure, having Bash as shell in all platforms will simplify crates a lot.

@mosteo BTW who is Vincent? :grin:

mosteo:
Oh my, sorry about the misnomer. There's been recent contributions from (at)Jellix and I just wasn't paying attention.

I think that Fabien meant why not use plain `cp` instead of a `/bin/sh` with arguments. Our cross-platform pain-reduction efforts rely on msys2 for Windows (already in use for tools like `tar`, `git`...), so a plain `cp` should work there too.

> /bin/sh could work in msys2 if alr calls the Bash as shell, (not cmd.exe or PowerShell). I'm sure, having Bash as shell in all platforms will simplify crates a lot.

 Right now, `alr` will spawn exactly what you give it in the action command. 

Mind you, this is somewhat academic for this crate since there's no `libzmq` for msys2 anyway, but if you wanted to future-proof it already, you might want to replace the current single action with:

```
[[general.actions]]
type = "post-fetch"
command = ["cp", "libzmq.gpr.in", "libzmq.gpr"]

[[general.actions]]
type = "post-fetch"
command = ["cp", "libsodium.gpr.in", "libsodium.gpr"]
```
and see how that fares. I think it should work just the same.

Fabien-Chouteau:
If you really want to use a shell, you can do:
```
"sh", "-c", "cp libzmq.gpr.in libzmq.gpr; cp libsodium.gpr.in libsodium.gpr"
```

That is with `sh`, instead of `/bin/sh`.

reznikmm:
I used shell, because I didn't know (or forgot) how to spawn two `cp` :disappointed: 

> there's no libzmq for msys2 anyway

Actually, there is. It is named as `zeromq` package, but I don't know how to write it in the crate, because `apt` is the only known packager in `alr`, right?

Fabien-Chouteau:
> Actually, there is. It is named as `zeromq` package, but I don't know how to write it in the crate, because `apt` is the only known packager in `alr`, right?

Actually msys2 is supported in Alire now.

You can do:
```
[[external]]
kind = "system"
    [external.origin.'case(distribution)']
    'debian|ubuntu' = ["libzmq3-dev"]
    'msys2' = ["mingw-w64-x86_64-zeromq"]
```

reznikmm:
> Actually msys2 is supported in Alire now.
> 

It's great! I've updated the `libzmq.toml`. Thank you!

mosteo (approved):

PR #124, opened 2020-06-02
Add 'SI_Units' library

Jellix:
Adds the SI_Units formatting library to ALR index.


Jellix:
Grmpff. Looks like `ubuntu-latest::debian-stable` is stuck...

PR #130, opened 2020-06-23
Add SPAT to Alire index

Jellix:
Let's see if I did it correctly...


mosteo:
Hi Vincent. I think your use case is manifesting a problem with our current indexing of gnatcoll. I need to think a bit more about it before discussing solutions. Meanwhile, could you tell me what platform/compilers are you interested in supporting?

Jellix:
Well, the code should be pretty portable, but requiring GNAT Community 2019/2020 (as these are the ones that come with SPARK) should be enough. Having SPAT without SPARK is a bit useless.

mosteo:
OK. So the situation is as follows: Community GNATs are bundled with  gnatcoll, so in that case it shouldn't be added as dependency. FSF GNATs, on the contrary, do not, so the dependency must be made explicit. To compound the problem, it seems our indexed `gnatcoll`s from sources (2018, 2019 versions) do not include the failing `libgpr` (and I don't see it on AdaCore download page). 

Taking all of that into account, and to cover all cases, your dependencies might look like this:
```
['1.0.0']
depends-on.si_units = "^0.1.0"
depends-on.'case(toolchain)'.system.gnatcoll = ">=19 & <2000"
```
The toolchain variable may be one of `system`, meaning the GNAT was detected as a FSF from system, or `user`, meaning it's user installed (a CE edition). Ubuntu 18.04 packages `libgnatcoll16.1.0-dev` with version `17.0.2017`, Ubuntu 20.04 packages `libgnatcoll18-dev` with version `19.0`. My testing on Ubuntu 18.04 shows its `gnatcoll` version is too old.

So the `gnatcoll` dependency is given like that to avoid the too old one in Ubuntu 18.04 and the broken ones packaged as sources. With these dependencies, it should build with GNAT CE 2020 (using the bundled gnatcoll) and, once we update the `gnatcoll`crate, in Ubuntu 20.04, using the system package.

Jellix:
Ok, I think I understand the issue, let's see what happens now. 🤔

Jellix:
Ok, that seems to work. All systems where we have dependencies not met are the ones that have been excluded by si_units, so that's expected. (I should check why si_units thinks ubuntu-latest:community-current is a debian, though).

mosteo:
> I should check why si_units thinks ubuntu-latest:community-current is a debian

Haven't checked the logs, but Ubuntu will be skipped because of the still unindexed `libgnatcoll18-dev`

Jellix:
> > I should check why si_units thinks ubuntu-latest:community-current is a debian
> 
> Haven't checked the logs, but Ubuntu will be skipped because of the still unindexed `libgnatcoll18-dev`

Yes, likely for this one, but si_units doesn't depend on GNATColl. But due the compilation issue with GNAT FSF at least on debian-stable I had to set si_units to unavailable for this system. What was mildly unexpected when I was checking the logs for the SPAT results, is that this caught the ubuntu-latest::community-current image, too - which with the below information is actually fine.

What I saw was that on ubuntu-latest::community the FSF compiler version was installed despite the name suggesting the Community Edition.

I think, the issue is in the build-crate.yml

```
    - name: Set up GNAT toolchain (FSF)
      if: matrix.os == 'ubuntu-latest'
      uses: ada-actions/toolchain@dev
      with:
        distrib: fsf # faster install?

    - name: Set up GNAT toolchain (Community)
      if: matrix.os != 'ubuntu-latest'
      uses: ada-actions/toolchain@dev
      with:
        distrib: community
```

which apparently always installs the FSF version on ubuntu-latest, so I think ubuntu-latest::debian-stable and ubuntu-latest::community-current are actually identical deployments.

mosteo:
> but si_units doesn't depend on GNATColl.

Ah, yes, sorry.

> which apparently always installs the FSF version on ubuntu-latest, so I think ubuntu-latest::debian-stable and ubuntu-latest::community-current are actually identical deployments.

Yes, this is in purpose to use the native compiler in platforms that actually have one. Ubuntu LTS and Debian stable may be identical in this particular iteration, but since Ubuntu freezes packages from Debian unstable, I have less trust on the [Ada] system packages for Ubuntu, so I thought valuable to test both explicitly, if only because Ubuntu is so popular.

Jellix:
> Yes, this is in purpose to use the native compiler in platforms that actually have one. Ubuntu LTS and Debian stable may be identical in this particular iteration, but since Ubuntu freezes packages from Debian unstable, I have less trust on the [Ada] system packages for Ubuntu, so I thought valuable to test both explicitly, if only because Ubuntu is so popular.

Ah, yes, that makes sense. :)

Well, FWIW, I think this can be merged now. The checks that are supposed to be successful are successful and those being skipped are skipped on purpose. Until I maybe figure out why the FSF compiler runs endless with the seemingly simple si_units compilation, there's nothing more to do.

mosteo (approved):

mosteo:
Merged, thanks!

PR #131, opened 2020-06-25
Update GNATCOLL Debian/Ubuntu system packages

mosteo:
User check failing because Pierre-Marie is listed as maintainer. Not sure if this information is current?


mosteo:
User check failing because Pierre-Marie is listed as maintainer. Not sure if this information is current?

Fabien-Chouteau (approved):

PR #133, opened 2020-07-10
si_units.toml: remove a test project file from project-files list

Fabien-Chouteau:
The project file is used for testing and is not supposed to be withed by users.

This way si_units is compatible with https://github.com/alire-project/alire/pull/458.

### DIFF
--- a/index/si/si_units.toml
+++ b/index/si/si_units.toml
@@ -4,7 +4,7 @@ licenses = ["WTFPL"]
 maintainers = ["vinzent@heisenbug.eu"]
 maintainers-logins = ["Jellix"]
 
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
@Jellix do you agree with this change?

Jellix:
Fine with me, yes. The primary use for this is to use SI_Units like a library, the other project file is basically used for actively developing on it (hence the tests).

So, from the user perspective, the `si_units_lib.gpr` is indeed the one that should be withed in almost all cases.

Jellix (approved):

mosteo (approved):

PR #136, opened 2020-08-25
Add ada-fuse library to index

NicolaiRuckel:


Fabien-Chouteau:
Thank you @NicolaiRuckel !

You also have to add a file for the external dependency on `libfuse`.
For instance: `libfuse.toml`
```toml
[general]
description = "Interface for userspace programs to export a filesystem"
licenses = ["LGPL 2.1"]
maintainers = ["fabien.chouteau@gmail.com"]
maintainers-logins = ["Fabien-Chouteau"]

[[external]]
kind = "system"
    [external.origin.'case(distribution)']
    'debian|ubuntu' = ["libfuse-dev"]
```

Fabien-Chouteau:
@mosteo the checks do not report any error even if there is a dependency on a crate that doesn't exists. That doesn't seem right.

NicolaiRuckel:
I also added `libfuse` to the index.

Fabien-Chouteau (approved):

Fabien-Chouteau:
Thank you @NicolaiRuckel

mosteo:
> @mosteo the checks do not report any error even if there is a dependency on a crate that doesn't exists. That doesn't seem right.

Right, at present we skip testing of crates with any missing dependency, since these will always require user intervention to build. This is different than depending on a system package that may exist only in some platforms, so the checks can be refined for the case you point out. I'm taking note of it.

PR #138, opened 2020-09-03
Update Ada-Fuse to version 1.0.1

NicolaiRuckel:
I tried to update the Alire configuration to the new format.


NicolaiRuckel:
If I understand the error message correctly, I should merge it into `devel-0.4`, which I did in #139.

mosteo:
Thanks, the default branch was still configured to be `devel-0.3` but indeed is now `devel-0.4`.

I'll check what's the problem over there, thanks for the contribution.

PR #148, opened 2020-09-07
Detect make from version output

mosteo:
This way it is available even in platforms without package manager (macOS).

This was the original reason it was defined like this, that I forgot. We can have both external definitions and the system installer will get called when needed.

### DIFF
--- a/index/ma/make/make-external.toml
+++ b/index/ma/make/make-external.toml
@@ -6,5 +6,9 @@ maintainers-logins = ["mosteo"]
 
 [[external]]
 kind = "system"
-[external.origin."case(distribution)"]
-"..." = ["make"]
+origin = ["make"]
+
+[[external]]
+kind = "version-output"
+version-command = [ "make", "--version" ]
+version-regexp = ".*Make ([\\d\\.]+).*"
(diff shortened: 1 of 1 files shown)


index/ma/make/make-external.toml
```diff
@@ -6,5 +6,9 @@ maintainers-logins = ["mosteo"]
 
 [[external]]
 kind = "system"
-[external.origin."case(distribution)"]
-"..." = ["make"]
+origin = ["make"]
+
```
Fabien-Chouteau, line 10: Does it mean that it won't be installed from the distrib anymore?

mosteo: No, that is just a shorter way of expressing the same. We don't need the case if we were assigning the same value to a single `'...'` alternative, as in practice that's assigning explicitly the same value to all case alternatives.

Fabien-Chouteau: Ah OK, I see it now.

Since external origins come from distrib package the previous was more explicit and readable to me.

Fabien-Chouteau (dismissed):

index/ma/make/make-external.toml
mosteo: I see, I'll put that back.

PR #151, opened 2020-09-17
samd51_hal 0.2.0

Fabien-Chouteau:


Fabien-Chouteau:
The build is not going to work because this is an embedded ARM project.

mosteo:
> The build is not going to work because this is an embedded ARM project.

Should we do something already about this kind of projects, or we leave this for the future? Impromptu ideas:

- Add dependencies on the cross-compiler
- Add a test action that succeeds and emits an informative message (although the CI script still doesn't use those #alire-project/alire#532)

Fabien-Chouteau:
In this instance, having the GNAT community arm-elf cross compiler installed would fix the build.
It is not available on macOS though.

mosteo:
> In this instance, having the GNAT community arm-elf cross compiler installed would fix the build.
> It is not available on macOS though.

So, IIUC, we could have a system crate for `gnat_arm_eabi` that detected the `arm-eabi-gnat` executable (I think that's the one?), and everything would work as expected in regard to Alire dependencies (there would be missing externals and this crate would be reported as unavailable, unless the compiler is installed).

As a separate issue, we could have an image with cross-compilers for testing, but rn I would simply add the dependency here.

mosteo:
I can take a stab at it, if you prefer.

Fabien-Chouteau:
> So, IIUC, we could have a system crate for `gnat_arm_eabi` that detected the `arm-eabi-gnat` executable (I think that's the one?), and everything would work as expected in regard to Alire dependencies (there would be missing externals and this crate would be reported as unavailable, unless the compiler is installed).

Toolchain handling in Alire is an interesting topic but we should keep it for after the first release IMO. 

> As a separate issue, we could have an image with cross-compilers for testing, but rn I would simply add the dependency here.

What do you mean here?

mosteo:
> What do you mean here?

Having a definition like in [this additional commit](https://github.com/alire-project/alire-index/compare/samd51_hal-0.2.0...mosteo:cross-example).

Fabien-Chouteau:
> Having a definition like in [this additional commit](https://github.com/alire-project/alire-index/compare/samd51_hal-0.2.0...mosteo:cross-example).

I think we should wait until we have a good toolchain handling.

mosteo:
OK then

PR #164, opened 2020-10-02
Various crates extracted from Ada Drivers Library

Fabien-Chouteau:


Fabien-Chouteau:
Again most are arm-elf crates and will not build in the checks.

mosteo (approved):

PR #165, opened 2020-10-03
Add AXMPP crate

reznikmm:


index/ax/axmpp/axmpp-1.0.0.toml
```diff
@@ -0,0 +1,61 @@
+##############################################################################
```
Fabien-Chouteau: Why put a license header here @reznikmm?

The license and authors are already defined in the manifest.

Fabien-Chouteau (changes requested):

index/ax/axmpp/axmpp-1.0.0.toml
reznikmm: I saw every file in that repository has a license header (even Makefile). I think the author did this intentionally. I'll ask him.

reznikmm: @Fabien-Chouteau I've dropped the header.

Fabien-Chouteau (approved):

PR #172, opened 2020-10-13
SI Units v0.1.2 (Alire integration).

Jellix:


mosteo (dismissed):

index/si/si_units/si_units-0.1.2.toml
```diff
@@ -0,0 +1,21 @@
...
+maintainers = ["vinzent@heisenbug.eu"]
+maintainers-logins = ["Jellix"]
+project-files = ["si_units_lib.gpr"]
+tags = ["utilities", "formatting"]
+
+[gpr-externals]
+BUILD = ["debug", "release"]
+
+[gpr-set-externals]
+BUILD = "release"
```
Fabien-Chouteau: Hi @Jellix,

If you set the `BUILD` mode here users won't be able to change it. You should have a default in the project file instead.
Also use `BUILD` as an external scenario variable is very likely to clash with other projects. You should rename it to `SI_UNIT_BUILD_MODE` for instance.

Fabien-Chouteau (changes requested):

Jellix:
Ok, will do.

Yet, from the catalog format spec:

> gpr-set-externals: optional dynamic table, giving a mapping from the name of external variables to the values to use ***by default*** when building the project.

Seems a bit misleading, if this default can not be changed.

Fabien-Chouteau:
> > gpr-set-externals: optional dynamic table, giving a mapping from the name of external variables to the values to use _**by default**_ when building the project.
> 
> Seems a bit misleading, if this default can not be changed.

Indeed, we have to change this.

Fabien-Chouteau (approved):

PR #178, opened 2020-10-15
minirest 0.1, curl external

mosteo:
Version currently in use by `alr`.

### DIFF
--- a/index/cu/curl/curl-external.toml
+++ b/index/cu/curl/curl-external.toml
@@ -6,7 +6,8 @@ maintainers-logins = ["mosteo"]
 
 [[external]]
 kind = "system"
-origin = ["curl"]
+[external.origin."case(distribution)"]
+"..." = ["curl"]
 
 [[external]]
 kind = "version-output"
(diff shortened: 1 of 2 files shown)


index/cu/curl/curl-external.toml
```diff
@@ -0,0 +1,14 @@
+name = "curl"
+description = "Command-line tool for transferring data with URL syntax"
+
+maintainers = ["alejandro@mosteo.com"]
+maintainers-logins = ["mosteo"]
+
+[[external]]
+kind = "system"
+origin = ["curl"]
```
Fabien-Chouteau: We can get it from the distrib

mosteo: Not sure what you mean here. That is the package in linuxes/msys2 in fact. If you mean the other external kind, this is the same situation as with `make` in which we fall back to detecting the executable when we don't have a package manager.

mosteo: Or do you mean making explicit the OS here, that's indeed probably clearer.

Fabien-Chouteau: My bad, I was confused be `orgin` is the same for all distrib.

PR #179, opened 2020-10-16
Ada Industrial Control Widget Library (AICWL)

Jellix:
I am not sure about the gtkada dependency. [Dmitry claims](http://www.dmitry-kazakov.de/ada/aicwl.htm#changes_log) it's

>   Changes (5 Aug 2018) to the version 3.18:
>    Switched to Gtk 3.14.15.

which he has been using since, but I am not sure which GtkAda version this translates to, so I went with ^19 which is probably slightly newer.


mosteo:
Hi @Jellix. I see gtkada is not being detected in any platform. If you don't want to stay within a major number, but to ensure a minimum one, you might try with ">=17" instead of "^17". This does not follow semantic versioning compatibility guidelines, but is a matter of knowing if gtkada itself follows them.

Only thing I know for sure ATM is that the version shipped with Ubuntu 20.04 is 19.0.

Jellix:
Updated, it is indeed a good idea to not require exactly 17, but rather at least 17.

> Only thing I know for sure ATM is that the version shipped with Ubuntu 20.04 is 19.0.

There was a ^19 there before, didn't seem to help in finding the dependencies.

Jellix:
Well, what do you know, surprise, surprise. At least on debian-stable (gtkada18) and ubuntu-lts (gtkada19) the crate gets compiled successfully. I guess that's ready to be included in the index.

I still need to do some clean-up, and provide better build scripts, but that's for later.

mosteo (dismissed):

mosteo:
> Well, what do you know, surprise, surprise.

I know this feeling :-)

Since this one involves gtkada, which is a complex beast, I'd wait for @Fabien-Chouteau to give us the final approval, in case I'm missing something.

Jellix:
Ok, if we're waiting then that's fine. I just hit a bit of a snag with git, the way I was planning to prepare releases, and how alr determines if the working git is up to date with the remote.

I set it to draft for now, so no worries.

Jellix:
Alright, looking good now and I think, I got my git back under control. :) So, ready to ~~rock~~ review, I'd say. :)

mosteo:
> how alr determines if the working git is up to date with the remote

You can publish using only the remote. Perhaps that suits your workflow better: `alr publish <remote git> [commit]`

Jellix:
Hmm. That kinda sucks. I tried minimizing dependencies, but `rsvg-2` is used by everything, `aicwl.gpr` withs `gtkada_contributions-rsvg.gpr` and then defines the Linker.Default_Switches with a `-lrsvg-2`, so no way around it.

Guess, I have to disable building any of the executables.

Jellix:
[That run](https://github.com/alire-project/alire-index/pull/179/checks?check_run_id=1270111342) looks weird. Looks as if alire is trying to build the same crate twice, despite having two different versions of the crate specified.

Opened [#577](https://github.com/alire-project/alire/issues/577) for that, because it looks like this could lead to other problems.

Best guess: I shouldn't specify two versions of the same crate in a single PR.

mosteo:
> [That run](https://github.com/alire-project/alire-index/pull/179/checks?check_run_id=1270111342) looks weird.

I had seen that error before but could not reproduce it. This could help me with isolating it.

About the test executables, don't fret too much about it as I guess we don't really want these to be built when a client uses the library. They could be moved to a separate crate (like we have `geste` and `geste_examples`) or they could wait until we have a way to define separate test projects/executables (not on the radar at the moment though, but a common enough occurrence I guess. I myself tend to have a test project too in my libraries.)

Jellix:
Well, I guess we can give it a go then (once @Fabien-Chouteau agrees wrt. gtkada).

index/ai/aicwl/aicwl-3.24.1.toml
```diff
@@ -0,0 +1,168 @@
...
+name = "aicwl"
+version = "3.24.1"
+authors = ["Dmitry A. Kazakov <mailbox@dmitry-kazakov.de>"]
+website = "http://www.dmitry-kazakov.de/ada/aicwl.htm"
+licenses = ["GMGPL 2.0", # according to website (http://www.dmitry-kazakov.de/ada/aicwl.htm)
+            "LGPL 3.0"]  # according to sourceforge (https://sf.net/projects/aicwl)
+maintainers = ["Vinzent \"Jellix\" Saranen <vinzent@heisenbug.eu>"]
+maintainers-logins = ["Jellix"]
+notes = "Due to missing libraries on the test machines, most executables are not currently built."
+project-files = ["sources/aicwl.gpr",
```
Fabien-Chouteau: Only project files used by users should be lister here.

index/ai/aicwl/aicwl-3.24.1.toml
```diff
@@ -0,0 +1,168 @@
...
+                 # my attempts resulted in some undefined symbol references
+                 # to libgtk3
+                 # "sources/test_gtkada/gtkada_contributions-source_view-tests.gpr",
+                 # "sources/test_gtkada/gtkada_contributions-tests.gpr",     # Needs rsvg-2
+                 "sources/test_strings_edit/strings_edit-test.gpr"
+                 # Requires components/xpm/xpm_parser
+                 # "sources/xpm2gtkada/xpm2gtkada.gpr"
+                 ]
+
+executables = [# examples/on_line_plotter/on_line_plotter.gpr
```
Fabien-Chouteau: The `executables` field is not supposed to be used to build tests or examples.

Fabien-Chouteau (changes requested):

index/ai/aicwl/aicwl-3.24.1.toml
Jellix: With "used by" do you mean projects files that are potentially `with`ed by other projects, or all project files that can be used with gprbuild in a meaningful manner (e.g. loading an example project in GNATStudio)?

index/ai/aicwl/aicwl-3.24.1.toml
Jellix: Removed all of them. xpm2gtkada would have been a candidate, but that doesn't build ATM (missing simple_components).

Fabien-Chouteau: If `xpm2gtkada` is a tool based on `AICWL`, it should be probably be in a separate crate.

index/ai/aicwl/aicwl-3.24.1.toml
Fabien-Chouteau: Projects that are "withed" by users of the library.

Fabien-Chouteau (approved):

mosteo (approved):

PR #181, opened 2020-10-16
GNATColl Core 21.0.0

Fabien-Chouteau:
This patch removes 2019 and 2018 releases. The rational is that those
releases follow the GNAT Community versioning (i.e. 20XX) and this is
not compatible with the versions from the GitHub repo (i.e. XX).  Since
we want to use the GitHub sources from now on, if we don't remove the
GNAT Community releases they will be the latest forever (in terms of
semver comparison).

This is against our policy that no releases should be removed from the
index. However we are doing this before the first release of Alire and
to avoid a mess of version handling for everyone (e.g. depends-on
gnatcoll=">=19 & <2000").

This patch also removes the external gnatcoll as there is now a GMGPL
source version.


mosteo (approved):

mosteo:
The failing check is expected given the removal of files, so going ahead.

PR #186, opened 2020-10-16
SI_Units v0.2.0

Jellix:


index/si/si_units/si_units-0.2.0.toml
```diff
@@ -0,0 +1,43 @@
+description = "Pretty print physical values in properly scaled metric (SI) units."
+long-description = """
+Provides generic conversion (`Image`) functions that convert values into human
+readable strings with appropriate SI prefixes. This is especially convenient
+when you have to deal with printing values from a potentially large interval,
+and your need to represent such values into something an average human will
```
mosteo: s/and your/and you/ ?

Jellix: Thanks for spotting it. So, somebody actually reads stuff like that. 👍

mosteo (approved):

PR #203, opened 2020-11-09
Second fix for spurious alire.toml detection

mosteo:
The fix in #201 was not complete. This one should work, as it's been fully tested in mosteo/alire-index#11.

As part of the fix, now we test with alr stable instead of rebuilding from master.

Took the opportunity to add Arch to the list of tested platforms.


mosteo:
Tested further with #205 so I'm going ahead and merge this one.

PR #209, opened 2020-11-23
rclada 0.1.0

mosteo:


mosteo:
Not sure what the error in https://github.com/alire-project/alire-index/pull/209/checks?check_run_id=1443565317 is about. Both the Ada and Python3 TOML parsers have no trouble with the file. Could it be an outdated javascript parser?

Fabien-Chouteau:
> Not sure what the error in https://github.com/alire-project/alire-index/pull/209/checks?check_run_id=1443565317 is about. Both the Ada and Python3 TOML parsers have no trouble with the file. Could it be an outdated javascript parser?

 - http://toml-online-parser.ovonick.com/ -> FAIL
 - https://toml-parser.com/ -> OK
 - https://pseitz.github.io/toml-to-json-online-converter/ -> OK
 - https://www.toml-lint.com/ -> FAIL

@pmderodat is that correct TOML in your opinion?

mosteo:
I'm pretty sure this is the same issue we discussed here: https://github.com/pmderodat/ada-toml/issues/5

So I'd say that it's the javascript lib here the one at fault. Still a problem for our checks though.

I'm going to merge if you don't object.

Fabien-Chouteau:
Ok for me :+1:

PR #212, opened 2020-12-01
update cortex_m and microbit_bsp

Fabien-Chouteau:


mosteo (approved):

index/co/cortex_m/cortex_m-0.3.0.toml
```diff
@@ -0,0 +1,33 @@
...
+website="https://github.com/AdaCore/Ada_Drivers_Library/"
+maintainers = ["chouteau@adacore.com"]
+maintainers-logins = ["Fabien-Chouteau"]
+project-files = ["cortex_m0.gpr", "cortex_m4.gpr", "cortex_m4f.gpr", "cortex_m7.gpr"]
+tags = ["embedded", "arm", "nostd"]
+
+auto-gpr-with=false # User has to select only one project file
+
+[[depends-on]]
+hal = "^0.1.0"
```
mosteo, line 28: Is this intended to be major- or minor- compatible?

index/mi/microbit_bsp/microbit_bsp-0.2.0.toml
```diff
@@ -0,0 +1,31 @@
...
+website="https://github.com/AdaCore/Ada_Drivers_Library/"
+maintainers = ["chouteau@adacore.com"]
+maintainers-logins = ["Fabien-Chouteau"]
+project-files = ["adl_middleware.gpr"]
+tags = ["embedded", "nostd", "microbit", "bbc", "nrf51", "bsp"]
+
+[[depends-on]]
+hal = "^0.1.0"
+nrf5x_hal = "^0.1.0"
+embedded_components = "^0.1.0"
```
mosteo, line 26: likewise

index/co/cortex_m/cortex_m-0.3.0.toml
Fabien-Chouteau: minor

index/mi/microbit_bsp/microbit_bsp-0.2.0.toml
Fabien-Chouteau: same

index/co/cortex_m/cortex_m-0.3.0.toml
mosteo: Asking because our caret implementation does not treat `0.x` versions in any special way. Hence to stay in a minor version you need the tilde.

I've read feedback defending this behavior and also the usual one where a caret is equivalent to a tilde for `0.x` versions (i.e., `'0.MAJOR.MINOR`). See for example: https://nodesource.com/blog/semver-tilde-and-caret/

I like that no special treatment means no surprises/hidden rules. But this may go against usual expectations.

Fabien-Chouteau: I feel like we need documentation/tutorial about this. Even more if there are different behavior out there.

mosteo: We state the plain meaning in the catalog spec, and it is also explicitly noted in the `semantic_versioning` repository. Certainly it would be better to make a more prominent warning in the catalog documentation (and perhaps detect it and emit a warning during publish?)

Fabien-Chouteau:
@mosteo can we merge this one. The new version fixes a nasty bug that makes other crates unusable.

mosteo:
Yup. I thought you were going to update the dependencies.

mosteo (approved):

PR #215, opened 2020-12-18
Add pkgconf for arch distribution to pkg_config

onox:
Arch Linux provides only core/pkgconf to provide pkg-config.

### DIFF
--- a/index/pk/pkg_config/pkg_config-external.toml
+++ b/index/pk/pkg_config/pkg_config-external.toml
@@ -8,4 +8,5 @@ maintainers-logins = ["Fabien-Chouteau"]
 kind = "system"
 [external.origin."case(distribution)"]
 "debian|ubuntu" = ["pkg-config"]
+arch = ["pkgconf"]
 msys2 = ["mingw-w64-x86_64-pkg-config"]


onox:
Debian/Ubuntu provide pkgconf as well: `apt show pkgconf` shows `Provides: pkg-config`. 

Is it possible to write in the toml file `"debian|ubuntu" = ["pkg-config|pkgconf"]`? (I guess alr doesn't look at the `Provides` property of apt and pacman?)

onox:
The debian-stable build crate fails:

```
ERROR: Loading externals from /home/runner/work/alire-index/alire-index/index/pk/pkg_config/pkg_config-external.toml:
ERROR:    external index 1:
ERROR:    case:
ERROR:    invalid enumeration value: arch
```

But on Arch Linux `alr show --external pkg_config` shows:

```
Kind           Description                 Details                            Available
System package 4 candidate system packages Debian: pkg-config                 True     
                                           Ubuntu: pkg-config                          
                                           Msys2: mingw-w64-x86_64-pkg-config          
                                           Arch: pkgconf
```

mosteo:
> Debian/Ubuntu provide pkgconf as well: `apt show pkgconf` shows `Provides: pkg-config`.
> 
> Is it possible to write in the toml file `"debian|ubuntu" = ["pkg-config|pkgconf"]`? (I guess alr doesn't look at the `Provides` property of apt and pacman?)

No, this is not supported (nor are these properties checked, as you correctly guessed). Also, we don't have currently a way to emulate either "provides" or to depend on two "xor" dependencies. (This was somewhat supported long ago: https://github.com/alire-project/alire/issues/106).

I'm looking into why Arch is not being recognized as a distro.

mosteo:
> I'm looking into why Arch is not being recognized as a distro.

Right, this is simply that Arch support was added after the latest stable release, and this branch tests against stable. We need to create a new devel branch. I will ping you so you can rebase against it when ready.

mosteo:
Hi, @onox, I think all the necessary changes are in place; so, could you please rebase this branch on top of `devel-0.5`?

mosteo (approved):

mosteo:
Thanks @onox, now it worked. Since your contributions to Arch require the development version of Alire, sadly they won't become generally available until the next stable release.

PR #217, opened 2020-12-23
Add sdl2 for arch distribution to libsdl2

onox:


mosteo (approved):

mosteo:
These patches raise the point that maintainers per platform do not mesh well with our current setup. Either every maintainer can edit all platforms, or some übermaintainer will have to manually review all edits to externals. One solution would be to move to separate external files per distro, which is a bit redundant. Another possibility would be to allow case syntax for maintainers.

Not a pressing issue I'd say, but something to keep in mind. Creating issue https://github.com/alire-project/alire/issues/653 to track this point.

PR #220, opened 2020-12-28
Adding the Lua scripting language

Patschkowski:
In the next step I want to make the Lua wrapper from AdaCore available to the Alire project

### DIFF
--- a/index/lu/lua/lua-5.4.2.toml
+++ b/index/lu/lua/lua-5.4.2.toml
@@ -6,7 +6,10 @@ maintainers-logins = ["Patschkowski"]
 licenses = ["MIT"]
 project-files = ["lua.gpr"]
 executables = ["lua"]
+
+[environment]
+PATH.prepend = "${CRATE_ROOT}/alire/build/bin"
 [origin]
-commit = "8ad492521f9c184ec8f8d900a86c8e5f97e635bf"
+commit = "b96b41bb7fce470363ad7a427d148cdfac3fd87e"
 url = "git+https://github.com/Patschkowski/lua.git"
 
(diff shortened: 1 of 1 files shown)


mosteo:
Thanks for your contribution, @Patschkowski. As I see, this cratebuilds the plain `lua` executable. So, for dependents on the crate, this executable should be in the PATH. Thus, your manifest should include:
```
[environment]
PATH.prepend = "${CRATE_ROOT}/alire/build/bin"
```

Patschkowski:
Integrated your proposal. Note, that for "alr publish" I have been using Alire on Linux as I am still seeing the same hang-up on Windows that I already reported.

mosteo (approved):

mosteo:
Thanks. Yes, I still have to get around to reproduce https://github.com/alire-project/alire/issues/638, hope to do it soon.

Fabien-Chouteau:
Hello @Patschkowski,
This is a pure C project that you can get from you distribution.

It doesn't fit into the Alire ecosystem which is a package manger for the Ada/SPARK language.

cc @mosteo

mosteo:
This one has to be reverted then. @Patschkowski, if you need the `lua` executable as a dependency, you could propose an external that installs the appropriate system package. For example, we have already one for the lua devel library: https://github.com/alire-project/alire-index/blob/stable-0.4/index/li/liblua/liblua-external.toml

It's possible the bindings you wanted to provide indeed depend on `liblua`?

Patschkowski:
I was not aware that there is also the option of external TOML files to get a package in your distribution installed - nice one. Does this also work on Windows? How to adjust that liblua-external.toml file to achieve something similar on Windows?

Fabien-Chouteau:
> How to adjust that liblua-external.toml file to achieve something similar on Windows?

On Windows we use the msys2 ecosystem. The package `mingw-w64-x86_64-lua` seems to be the one you want.

See `libsdl` for example: https://github.com/alire-project/alire-index/blob/stable-0.4/index/li/libsdl2/libsdl2-external.toml

PR #222, opened 2020-12-30
pick littlefs 0.1.0 into devel

mosteo:
So it doesn't get lost when we make devel the new stable.


mosteo:
Created against wrong base branch and now actions are confused, will recreate.

PR #223, opened 2020-12-30
littlefs 0.1.0 (into devel) (#219)

mosteo:
So it doesn't get lost when `devel-0.5` becomes the new stable branch.


mosteo:
Failed test, closing

PR #225, opened 2021-01-04
Added new crate mb_accel

KLOC-Karsten:
Added a new crate mb_accel (accelerometer detector for the micro:bit board).


Fabien-Chouteau:
Hello @KLOC-Karsten, thanks for you contribution.

I think this package would be better for the Ada Drivers Library than its own Alire crate: https://github.com/AdaCore/Ada_Drivers_Library

PR #226, opened 2021-01-05
Convert licenses to SPDX expressions

Fabien-Chouteau:
To be merged once https://github.com/alire-project/alire/pull/629 is merged.


index/li/liblzma/liblzma-external.toml
```diff
@@ -1,6 +1,6 @@
 description = "The GNU Readline Library"
 name = "liblzma"
-licenses = ["GPL 2.0", "GPL 3.0", "LGPL 2.1"]
+licenses = "GPL-2.0-or-later AND GPL-3.0-or-later AND LGPL-2.1-or-later"
```
mosteo, line 3: My understanding from reading SPDX docs [here](https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/) is that when you're giving the choice the appropriate operator is `OR`.

index/ze/zeromq_ada/zeromq_ada-4.1.5.toml
```diff
@@ -3,7 +3,7 @@ name = "zeromq_ada"
 version = "4.1.5"
 website = "https://github.com/persan/zeromq-Ada"
 authors = ["Per Sandberg <per.s.sandberg@bahnhof.se>"]
-licenses = ["MIT", "GMGPL 2.0"]
+licenses = "MIT AND GPL-2.0-only"
```
mosteo, line 6: Same thing here. Although it's possible both licenses apply and then `AND` is correct. Hi @reznikmm, are these licenses up for choice or do both apply simultaneously (due to reused code perhaps?)

mosteo:
Sorry for the delay, I think the review got stuck in an unsent state...

index/li/liblzma/liblzma-external.toml
Fabien-Chouteau: liblzma licensing is a tad complicated: https://git.tukaani.org/?p=xz.git;a=blob;f=COPYING

Maybe the library itself in public domain so we can only use that one here.

mosteo: All right, from reading that, indeed the proper operator would have been AND. But, as you say, the compiled bits seem to be all public domain.

index/ze/zeromq_ada/zeromq_ada-4.1.5.toml
Fabien-Chouteau: I asked the author here: https://github.com/persan/zeromq-Ada/issues/17

mosteo: Thanks, Fabien.

Fabien-Chouteau:
Looks like I cannot delete the `devel-05-spdx` branch, can you?

mosteo:
> Looks like I cannot delete the `devel-05-spdx` branch, can you?

Right, `devel-` branches are protected. I temporarily removed the protection to delete it.

PR #245, opened 2021-02-08
Adding GWindows to Alire

Patschkowski:

PR #246, opened 2021-02-08
Fix temporarily broken build checks

mosteo:
As the devel and stable index branches are converging, yet the alr installer still uses the beta, the build workflow needs to be temporarily tweaked.


mosteo:
I'm merging this one so #245 can be merged too.

PR #247, opened 2021-02-10
inotify 1.0.1

onox:


onox:
@mosteo `Check PR` doesn't understand `alr search --crates`

mosteo:
> @mosteo `Check PR` doesn't understand `alr search --crates`

I'm on it

mosteo:
The problem with the "Check PR" workflow should be fixed now... But there is some other problem with the macOS build.

onox:
> But there is some other problem with the macOS build

The macOS and windows builds should not run, inotify is for Linux only.

mosteo:
Right you are. For some reason the linux alr version was compiled instead. I'll check what's going on.

mosteo (approved):

mosteo:
Should be fixed for the future, sorry for the inconveniences.

onox:
Thank you :+1:

PR #249, opened 2021-02-11
Temporary workaround for the main CI workflow

mosteo:
Same problem as fixed in #246


mosteo:
Merging as this is the same as #246. Should be reverted together once `setup-alire` is updated for 1.0

PR #250, opened 2021-02-12
Add UXStrings library to Alire.

Blady-Com:


mosteo:
Hello @Blady-Com, thanks for your contribution.

From the error I'm seeing in the automatic checks, I'd say you have some unstated dependency on a minimum version of GNAT:
```
uxstrings1.ads:33:07: error: "String_Literal" is not a valid aspect identifier
```
so you should add something like
```
[[depends-on]]
gnat = ">=x.y" #  replace x.y with the minimum version that can compile your project.

**Comment by Blady-Com** [comment]
Hello @mosteo, I use GNAT Community 2020 on macOS 10.15, then is it correct to add:
gnat = ">=2020" ?
For my information which GNAT version is used in automatic checks?
Which version of macOS?
Where automatic check scripts could be found?

**Comment by mosteo** [comment]
> gnat = ">=2020" ?

Yes, but that might be overly restrictive in some cases. Probably not in this one, as my cursory test with Arch shows that GNAT 10.2 does not yet understand the problematic aspect.

> Which version of macOS?

The `macos-latest` in GitHub Actions: https://github.com/actions/virtual-environments . I think it's the same you use.

> Where automatic check scripts could be found?

The builds are triggered by the [build-crate.yml](https://github.com/alire-project/alire-index/blob/stable-1.0/.github/workflows/build-crate.yml) file. There you can see the different configurations.

> For my information which GNAT version is used in automatic checks?

For Windows/macOS, that would be CE2020. For Linux, we have a suite of [docker images](https://github.com/alire-project/dockerfiles) and we test with the natively packaged GNAT, or else CE2020.

Not all of those images are used; only the latest version of each. You can find exactly which ones in the workflow YAML file I just linked above.

File: index/ux/uxstrings/uxstrings-21.2.7.toml [hunk 1]
```diff
@@ -0,0 +1,21 @@
+name = "uxstrings"
+description = "Unicode Extended Strings utilities"
+version = "21.02.07"
```
mosteo: Hi Pascal, this version looks to be the current date. As we require [semantic versioning](https://semver.org), you have several choices:
1. If you intend to follow semantic versioning, pick an initial version. You can still keep the date in the build part of the version: `1.0.0+21.02.07`, for example. Anything after `+` doesn't influence the version ordering.
2. If you don't intend to follow SV, you can put the whole date as the major number: `210207.0.0`, to make sure that every release is a clean break from the previous one.
3. Also if you don't intend to follow SV, you could use a mix of both ideas and simply increase the major number with every release you publish: `1.0+21.02.07`, `2.0+21.03.15`, etc.

mosteo: I'm so sorry, Pascal, my last message stayed in a draft state for a week :woman_facepalming:

Blady-Com:
> Le 22 févr. 2021 à 21:17, Alejandro R Mosteo <notifications@github.com> a écrit :
> 
> @mosteo commented on this pull request.
> 
> In index/ux/uxstrings/uxstrings-21.2.7.toml:
> 
> > @@ -0,0 +1,21 @@
> +name = "uxstrings"
> +description = "Unicode Extended Strings utilities"
> +version = "21.02.07"
> 
> Hi Pascal, this version looks to be the current date. As we require semantic versioning, you have several choices:
> 
> 	• If you intend to follow semantic versioning, pick an initial version. You can still keep the date in the build part of the version: 1.0.0+21.02.07, for example. Anything after + doesn't influence the version ordering.
> 	• If you don't intend to follow SV, you can put the whole date as the major number: 210207.0.0, to make sure that every release is a clean break from the previous one.
> 	• Also if you don't intend to follow SV, you could use a mix of both ideas and simply increase the major number with every release you publish: 1.0+21.02.07, 2.0+21.03.15, etc.

Hello Alejandro,

In fact, I used the tag stamp of my Github project (https://github.com/Blady-Com/UXStrings/releases/tag/UXS_20210207).
I had a look in some projects like VSS where it seem to be similar: vss-21.0.0.toml:3: version = "21.0.0".

Well, you're correct, this is the date as up to now UXStrings remains let's say in a very alpha state with up coming versions which could break compatibility following feedbacks of users. However, I wouldn't like that the major version might increase.

What about: "0.0.0-alpha+20210207"?

mosteo:
> What about: "0.0.0-alpha+20210207"?

Sure, it the project is in alpha the appropriate thing is to be at `<1.0`. But, bear in mind the dash means "before", and that the part after `+` is not used for comparison, so if the only difference is in that part, the versions are equivalent. This is to say that if you intend to release several alphas, I would go with something like `0.0.x+alpha-20210207`, increasing the `x`; or perhaps `0.0.20210207+alpha`.

On the projects that you mention that use the year for the major number: my understanding is that they don't really intend for the yearly releases to be fully backwards compatible, so the year can double as major version. But it's something that I don't really like, to be frank, as on the one hand it puts you in a corner if you were forced to release two breaking milestones in the same year; and on the other hand, if you happen to have two consecutive releases that are compatible, you lose the opportunity to have a safe upgrade.

mosteo (approved):

mosteo:
Thanks for your patience, @Blady-Com, it's merged now.

PR #257, opened 2021-02-25
vss-22.0.0-20210224.toml: licenses should not be an array anymore

Fabien-Chouteau:
There is a warning for that and we should reject it from the community index.


mosteo:
I'll promote this warning to an error for our server-side checks.

mosteo (approved):

PR #259, opened 2021-03-02
Add Raspberry Pi RP2040 and Pico packages

JeremyGrosser:


JeremyGrosser:
It looks like the CI system doesn't have an arm-eabi toolchain available, so the builds fail.

Fabien-Chouteau:
> It looks like the CI system doesn't have an arm-eabi toolchain available, so the builds fail.

Indeed, I will check it on my setup and let you know.

Fabien-Chouteau (approved):
Compilation is ok with GNAT Community 2020 arm-elf.

mosteo (approved):

mosteo:
> Compilation is ok with GNAT Community 2020 arm-elf.

Thanks, I'm going to merge now. Thanks @JeremyGrosser for the submission.

> It looks like the CI system doesn't have an arm-eabi toolchain available, so the builds fail.

Indeed, let's hope to find a solution for this in the not distant future.

PR #260, opened 2021-03-06
Replace wrong 'word_size' with 'word-size'

mosteo:
The old parser for some reason accepted both, when the catalog spec always uses `word-size`.


mosteo:
I see the Windows build is failing, but that's unrelated to the fix in this PR.

Fabien-Chouteau (approved):

PR #263, opened 2021-03-24
sdlada 2.5.4-1

mosteo:


Fabien-Chouteau (dismissed):

index/sd/sdlada/sdlada-2.5.4-1.toml
```diff
@@ -0,0 +1,64 @@
...
+command = ["make", "-C", "build/gnat", "SDL_PLATFORM=windows"]
+
+[available."case(os)"]
+'linux|macos|windows' = true
+"..." = false
+
+[available."case(distribution)"]
+# There is some problem I don't readily know how to solve:
+# /usr/sbin/ld: /usr/lib/libharfbuzz.so.0: undefined reference to `fstat64@GLIBC_2.33'
+arch = false
```
onox, line 43: @mosteo Have you tried `pacman -Syu`? This version of the crate works just fine here on Arch (I did get an exception in alr when trying to upgrade sdlada 2.3 with `alr update` if I just set `arch = true`, so I had to remove the whole block).

mosteo: Perhaps our test docker for Arch is stale. I will try refreshing it. Do you have any more info on the exception?

mosteo: Nvm, I just hit the same problem and got the stack trace.

mosteo: Tracking it at https://github.com/alire-project/alire/issues/719

mosteo: The good news is that with the updated Arch image it indeed builds, so I'll enable it.

PR #267, opened 2021-04-01
Added raspberry_bsp package

TamaMcGlinn:
Hi Alex,

The package itself works well enough for a 1.0 release, as you can blink LEDs and access buttons on the raspberry pi.
However, it's unclear how to add this to alire, since you have modified the release distribution method heavily since writing the Contributing.md guide. That still says to make a .tar.gz and get it's hash, but I think you support direct commit hashes now. It also says:

> Note that, as of this writing (Dec 2019), only development branches exist, until the first stable release of alr is made.

So it's also not clear how to decide between these two branches. The stability of the alr tool itself doesn't seem to have any bearing on the stability of packages.

Currently, I get this error:

> error: Malformed shelf folder name: /home/tama/.config/alire/indexes/community/repo/index

Finally, I would strongly recommend you allow the opening of issues on the alire-index project itself. I am sure someone would have already noted regarding your Contributing.md guide much earlier, if it were possible to notify you of that without a pull request.

### DIFF
--- a/index/ra/raspberry_bsp/raspberry_bsp-1.0.0.toml
+++ b/index/ra/raspberry_bsp/raspberry_bsp-1.0.0.toml
@@ -1,0 +1,16 @@
+name = "raspberry_bsp"
+description = "Board Support package for Raspberry PI v1, v2 and B+"
+version = "1.0.0"
+
+authors = ["Tama McGlinn"]
+maintainers = ["Tama McGlinn <t.mcglinn@gmail.com>"]
+maintainers-logins = ["TamaMcGlinn"]
+
+licenses = "MIT"
+tags = ["raspberry", "pi"]
+website = "https://github.com/TamaMcGlinn/ada_raspio"
+
+[origin]
+commit = "a15c773223e31c3ba97dc806897ab1feec2ff5a3"
+url = "git+https://github.com/TamaMcGlinn/ada_raspio.git"
+
--- a/index/ra/raspberry_bsp/raspberry_bsp-1.0.0.toml
+++ b/index/ra/raspberry_bsp/raspberry_bsp-1.0.0.toml
@@ -14,3 +14,5 @@ website = "https://github.com/TamaMcGlinn/ada_raspio"
 commit = "a15c773223e31c3ba97dc806897ab1feec2ff5a3"
 url = "git+https://github.com/TamaMcGlinn/ada_raspio.git"
 
+[available."case(os)"]
+windows = false


TamaMcGlinn:
Okay, I figured it out; the contributing.md guide should just say to run alr publish and follow the instructions in there.

TamaMcGlinn:
The build fails on windows because the bcm2835 library is not installed to D:/usr/local/lib. In principle, it should be perfectly usable on windows, provided you have that library installed somewhere, and you have a crosscompiler to raspberry pi up and running. How would I indicate to the CI that this library should first be installed?

onox:
> How would I indicate to the CI that this library should first be installed?

AFAIK you need to create an Alire crate for each external library. See https://github.com/alire-project/alire-index/tree/stable-1.0/index/li for some examples. Alire on Windows uses the msys2 distribution, but I don't think it's available there. Is the library open source under a license that allows distribution and modifications? In that case you may want to put it on Github with an alire.toml file and make sure it builds through gprbuild. Then you can reference that crate in your raspberry_bsp crate.

onox:
If that BCM library is always installed in some specific well-known location on Windows, then there is another trick you might try: 

Move the import of `bcm2835_library` to some OS dependent .gpr file (one for Linux and one for Windows), each importing the appropriate version of the bcm2835 lib (`bcm2835_linux` and `bcm2835_windows`)

In your alire.toml file you can then write:

```toml
[project-files.'case(os)']
linux = ["raspberry_bsp_linux.gpr"]
windows = ["raspberry_bsp_windows.gpr"]
```

Both these files can import `raspberry_bsp` containing all the Alire stuff. See https://github.com/onox/awt/blob/master/awt-linux.gpr for an example.

But this would result in 5 .gpr files and users need to manually install that BCM library.

mosteo:
> Finally, I would strongly recommend you allow the opening of issues on the alire-index project itself. I am sure someone would have already noted regarding your Contributing.md guide much earlier, if it were possible to notify you of that without a pull request.

Hi @TamaMcGlinn , thanks for you contribution and feedback. We are currently centralizing all work, docs, issues, etc, at https://github.com/alire-project/alire . I'll take care of fixing the documentation discrepancies. Also, our latest up-to-date docs should always be at https://alire.ada.dev (for the latest stable version) or the alire github (for the development branch). And, as you mention, the stability of the index is managed independently of the tool itself (currently, only the `stable-1.0` index branch is in use) since the release of `alr 1.0`.

In regard to your issues on Windows, if you want to get your crate merged ASAP, you can disable it on that platform until the necessary dependencies are available, with an available entry:

```
[available."case(os)"]
windows = false
```

A more complete solution would be, as @onox suggested, to submit currently missing dependencies, even if they're not available in all platforms. This will ensure that, for platforms that don't have them, the build is skipped automatically. That means that you can submit the external crate definition and, for windows, leave it unset (or provide a [hint external](https://alire.ada.dev/docs/#external-releases) indicating that it must be installed by hand).

Here is an example of an external definition where Windows is left out of the picture: https://github.com/alire-project/alire-index/blob/stable-1.0/index/li/libx11/libx11-external.toml

Let me know if you need guidance with anything else.

Fabien-Chouteau:
> Move the import of `bcm2835_library` to some OS dependent .gpr file (one for Linux and one for Windows), each importing the appropriate version of the bcm2835 lib (`bcm2835_linux` and `bcm2835_windows`)

Actually it's better to set the library path depending on the OS package manager:
https://github.com/alire-project/alire-index/blob/f545e1b830db1dc84489ad6967237fde589ee223/index/sd/sdlada/sdlada-2.5.4-1.toml#L51

The less gpr project files the better.

onox:
> Actually it's better to set the library path depending on the OS package manager

Ah, didn't know about this. Less gpr files would indeed be better.

mosteo:
I'm a bit out of the loop here... @TamaMcGlinn , should this crate build be succeeding on Windows?

TamaMcGlinn:
I appreciate the thoughts on how to include a dependency that is operating system dependent. I think, though, that it doesn't make a lot of sense to split this into a separate bcm2835 package, as the only intent of the raspberry_bsp package is to provide a nice type-safe Ada binding to bcm2835. However, it would be nice if possible, to make the following steps somehow automatic, and I think we could achieve that by putting the sources of bcm2835 into this repo. From the readme:

```
wget http://www.airspayce.com/mikem/bcm2835/bcm2835-1.68.tar.gz
tar zxvf bcm2835-1.68.tar.gz
cd bcm2835-1.68/
./configure
make
sudo make check
sudo make install
```

However, maybe it is still better to upload bcm2835 as a separate alire package. That would have the advantage that if someone decides my binding isn't right, they can still reuse that alire package to build their own abstraction on.

mosteo:
> to make the following steps somehow automatic,

That could be achieved with an `[[actions]]` entry, look for it in https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md

Basically you can trigger a script run before the build.

But now that the build is passing, is there something else to wait for? I confess I'm not grasping the finer details in this PR.

TamaMcGlinn:
For now I'd rather merge this as-is, and do that in a next version. I just checked and it works as long as you follow the steps in the readme.

mosteo (approved):

PR #270, opened 2021-04-27
Add external crate template

mosteo:


mosteo:
This is just a template outside the actual index, so the failures are of no concern.

Fabien-Chouteau (approved):

PR #276, opened 2021-05-04
Openglada: Correct libx11 version

WickedShell:
I tried to install openglada, but it fails due to the libx11 dependency. As far as I can tell the most recent release of libx11 is 1.7.0 in November 2020, and there is no version 2. This lowers the dependency to version 1. I'm guessing that this worked for someone originally as the Ubuntu package oddly has `2:` prefixed at the start. On Arch (or most other distro's) that give the actual version this is breaking.

I don't know how to locally test an Alire package, so this is a change by inspection, and not locally tested as I'm not sure how to test it.

### DIFF
--- a/index/li/libx11/libx11-external.toml
+++ b/index/li/libx11/libx11-external.toml
@@ -7,4 +7,5 @@ maintainers-logins = ["mosteo"]
 [[external]]
 kind = "system"
 [external.origin."case(distribution)"]
+"arch" = ["libx11"]
 "debian|ubuntu" = ["libx11-dev"]
--- a/index/op/openglada/openglada-0.6.0.toml
+++ b/index/op/openglada/openglada-0.6.0.toml
@@ -19,8 +19,11 @@ libx11 = "^1"
 Auto_Exceptions = ["enabled", "disabled"]
 GLFW_Version = ["2", "3"]
 Mode = ["debug", "release"]
-[gpr-set-externals."case(os)".linux]
-Windowing_System = "x11"
+
+[gpr-set-externals."case(os)"]
+linux = { Windowing_System = "x11" }
+macos = { Windowing_System = "quartz" }
+windows = { Windowing_System = "windows" }
 
 [origin]
 url = "https://github.com/flyx/OpenGLAda/archive/v0.6.tar.gz"
--- a/index/op/openglada/openglada-0.6.0.toml
+++ b/index/op/openglada/openglada-0.6.0.toml
@@ -10,8 +10,8 @@ maintainers-logins = ["flyx"]
 project-files = ["opengl.gpr", "opengl-glfw.gpr", "opengl-soil.gpr", "opengl-test.gpr"]
 
 [[depends-on]]
-[depends-on."case(os)".linux]
 gnat = "<8"
+[depends-on."case(os)".linux]
 libglfw3 = "^3"
 libx11 = "^1"
 
--- a/index/op/openglada/openglada-0.6.0.toml
+++ b/index/op/openglada/openglada-0.6.0.toml
@@ -11,6 +11,7 @@ project-files = ["opengl.gpr", "opengl-glfw.gpr", "opengl-soil.gpr", "opengl-tes
 
 [[depends-on]]
 [depends-on."case(os)".linux]
+gnat = "<8"
 libglfw3 = "^3"
 libx11 = "^1"
 
--- a/index/op/openglada/openglada-0.6.0.toml
+++ b/index/op/openglada/openglada-0.6.0.toml
@@ -12,7 +12,7 @@ project-files = ["opengl.gpr", "opengl-glfw.gpr", "opengl-soil.gpr", "opengl-tes
 [[depends-on]]
 [depends-on."case(os)".linux]
 libglfw3 = "^3"
-libx11 = "^2"
+libx11 = "^1"
 
 [gpr-externals]
 Auto_Exceptions = ["enabled", "disabled"]


mosteo:
Thanks, @WickedShell . This fix is likely correct as the problem with the `"2:"` epoch was fixed in https://github.com/alire-project/alire/pull/652 for Arch (Debian/Ubuntu should detect epochs since the beginning).

mosteo:
Now the funny thing is that Arch compiles while Ubuntu fails, both with gnat 9.3... I will have to research this particular point further.

mosteo:
My bad. Yesterday I was low on sugar or something. Arch is using gcc 10.2 (which will fail just the same as 9.3 in Ubuntu/Debian, or CE2020), and there is an additional change this patch needs: to add `libx11` for Arch in the `libx11-external.toml` index file. It is currently indexed only for Ubuntu/Debian. 

With that, openglada will resolve dependencies in Arch and fail at the same point: 

```
gl-objects-textures.adb:547:07: warning: in instantiation at gl-enums-indexes.ads:17
gl-objects-textures.adb:547:07: warning: cannot call "Get_Max" before body seen
gl-objects-textures.adb:547:07: warning: Program_Error will be raised at run time
gl-objects-textures.adb:556:07: warning: in instantiation at gl-enums-indexes.ads:17
gl-objects-textures.adb:556:07: warning: cannot call "Get_Max" before body seen
gl-objects-textures.adb:556:07: warning: Program_Error will be raised at run time
```

So the patch will improve things but a more recent version of openglada has to be indexed (I see a `0.8` release in the repo).

@flyx, excuse me for bringing you here. Can you confirm that this problem is fixed in more recent gnats? We are discussing `0.6` here.

flyx:
@mosteo This problem has indeed been fixed in flyx/OpenGLAda@31511a9f6e3101705637ce8a434b4cfe61ae4993, which has been released in 0.7.0.

mosteo:
I've pushed the fixes mentioned, plus marked openglada 0.6 depending on old gnat versions. Things should work properly once a new release of openglada is indexed ;-)

mosteo (approved):

WickedShell:
@mosteo Sorry for the late response. Is there a reason why Alire is tagging 0.6.0 here instead of 0.8.0 for Openglada? Given that it fixes the compilation error that seems like a better step forward to me then just trying to limit the GNAT version...

Can you also point me at docs for how to use a local Alire package description instead of from the repo so that I can test this type of change locally? I'm sure it exists, but I appear to have read right past it...

Is there a better place to ask this kind of question? I noticed the issues list is off here, and wasn't sure where else to ask.

mosteo:
> Sorry for the late response. Is there a reason why Alire is tagging 0.6.0 here instead of 0.8.0 for Openglada?

Just that that's the version that the maintainer has actually published to date in Alire. Hopefully they'll publish the more recent version soon. We prefer to wait for the official maintainers to publish the versions they know are best to distribute, unless there's a critical bottleneck that has to be solved fast.

> Given that it fixes the compilation error that seems like a better step forward to me then just trying to limit the GNAT version...

Indeed, having the last version will get rid of that trouble. At the same time, it's appropriate to mark that `0.6` won't work out of the box with newest compilers, so users are not presented with unexpectedly broken solutions.

> Can you also point me at docs for how to use a local Alire package description instead of from the repo so that I can test this type of change locally? I'm sure it exists, but I appear to have read right past it...

Actually this might not be explicitly documented yet, although it is doable with existing options. The simplest way would be to clone the community index and configure alire to use it via `alr index` subcommands. If you don't want to mess with your default configuration, you can set `$ALR_CONFIG`, or use the `-c` global option, to temporarily use another config folder. Still, you'll have to configure the cloned index with `alr index`. The last step is to modify the local index to do your experiments.

> Is there a better place to ask this kind of question? I noticed the issues list is off here, and wasn't sure where else to ask.

Yes, the issues are off here because I prefer to centralize all issues for the project at the https://github.com/alire-project/alire repo, at least for now. For this kind of user questions, you can use stackoverflow which I monitor for Ada/Alire questions, or https://gitter.im/ada-lang/Alire, as you prefer.

WickedShell:
> Just that that's the version that the maintainer has actually published to date in Alire

@mosteo Who is the maintainer though? I actually e-mailed flyx originally since he was listed on the package, but he told me he isn't actually the maintainer of the Alire package, and I'm not sure who to chase up from here.

Thanks for the advice on testing changes locally, I'll give it a shot soon hopefully.

mosteo:
Curious. Since `git blame` blames me for the maintainer lines, I guess at that time *I* packaged it but we hadn't properly defined the meaning of these fields and I used the upstream maintainer. Or I simply was having a bad day.

No matter, that means that I must update the release to list myself as maintainer, and I can also make a new release for `0.8`. Will do shortly. Of course, flyx can take over maintainership any time he wants.

WickedShell:
@mosteo Haha, thanks, I do appreciate it.

flyx:
@mosteo I am in process of crafting a new OpenGLAda release, and I plan to have an official Alire listing when it's done.

mosteo:
@flyx great, let me know if you need any guidance.

flyx:
@mosteo One thing is that the import name has has always been *„opengl“* since the GPR file is named `opengl.gpr`. The additional utilities (GLFW, image loading etc) are separate libraries (e.g. *opengl-glfw*) and I will split them into own Alire crates as suggested in the best practices.

Since the best practices say that the crate name should match the `.gpr` project file, I wonder if I should create the Alire crate as *opengl* instead of *openglada*. I always treated OpenGLAda as name of the umbrella project providing the base binding and additional utilities. On the other hand, naming the project *opengl* might be understood as being some kind of *official* binding, which it is not. But I do feel dropping the *Ada* suffix makes sense in the crate name since it is superfluous in an Ada package manager. In the crate description, I would give something like „Thick OpenGL bindings provided by the OpenGLAda project“.

What do you think?

mosteo:
Hi @flyx. On the splitting of crates, that makes sense if each subproject introduces different dependencies. This way a client is not forced to depend on something not necessary. Otherwise it's less important. 

On the naming, I'm with you in that `opengl` looks too official. Also your binding is thick, so it brings extra value. For that reason in this case a name that isn't just `opengl` would be better IMO. If your project is generally known already as `openglada`, I would stick with it even if in our context is redundant. There are already many crates that keep the `ada` for this reason.

Finally, that the project file matches the crate name is desirable, but not a hard line. For a new crate that would be the logical choice; for a established project like yours I think it's better to keep the same names that people may be already using and expecting. You can specify these names in the manifest and alire will use them transparently anyway.

flyx:
@mosteo Fair enough. I'll go with project names `openglada`, `openglada_images` etc. They do have different dependencies, e.g. the GLFW binding obviously depends on the GLFW library while the core OpenGL binding does not (and could theoretically also be used along with the existing SdlAda binding, which makes GLFW unnecessary).

I see that `libfreetype` has no Alire definition yet and I will add that since it is a, yet unmentioned, dependency of OpenGLAda's text rendering facilities.

mosteo:
Great, thanks!

flyx:
@mosteo I am almost done, however I can't figure out how to have multiple Alire projects in the same repository. See [my current repo structure](https://github.com/flyx/OpenGLAda/tree/master):

The `opengl` directory contains the `openglada` Alire crate. When I execute `alr publish` there, I get:

```
error:    Remote sources are missing the 'alire.toml' manifest file.
```

It seems to search the `alire.toml` in the root directory. Is that a bug?

mosteo:
Just in case someone reads this PR at a later time, for now the only way to index several crates in a repo is to generate zips for each of them. More info in PR alire-project/alire#737

PR #278, opened 2021-05-17
added external libfreetype

flyx:
This is needed as a dependency of OpenGLAda's text rendering facilities.

I am not entirely sure whether the msys2 name should be `mingw-w64-freetype` (base package) or `mingw-w64-x86_64-freetype` (architecture-specific package) since I have no experience with pacman.


mosteo:
From cursory inspection of other windows crates, it seems the x86_64 version is the one we need.

flyx:
Changed it accordingly.

mosteo (approved):

mosteo:
Merged, thanks

PR #280, opened 2021-05-18
OpenGLAda 0.9.0

flyx:
Split into four crates due to optional dependencies.

### DIFF
--- a/index/op/openglada/openglada-0.9.0.toml
+++ b/index/op/openglada/openglada-0.9.0.toml
@@ -7,6 +7,7 @@ licenses = "MIT"
 maintainers = ["Felix Krause <contact@flyx.org>"]
 maintainers-logins = ["flyx"]
 project-files = ["opengl.gpr"]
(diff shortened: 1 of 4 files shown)


mosteo:
Thanks a lot, @flyx, this is great.

I'd only suggest to add some tags to the crates. You don't need to repackage, just update the manifests in this PR. This way your crates can be better located at https://alire.ada.dev/

flyx:
The macOS build possibly fails because of the required non-Ada dependencies. Strange that there is no log available.

mosteo (approved):

mosteo:
GH actions are not 100%-proof. They seem to be acting up right now, I'll relaunch them in a while.

mosteo:
Indeed https://www.githubstatus.com/ reports degraded actions.

mosteo:
In the end it was GH's own trouble and now all tests have passed. Merging, thanks!

PR #281, opened 2021-05-28
usb_embedded 0.1.0

Fabien-Chouteau:


mosteo:
@Fabien-Chouteau , is this one of those that fail because of the cross-compiler?

Fabien-Chouteau:
Looks like there another problem. I will close until take the time to fix it.

PR #283, opened 2021-05-31
Fix utilada Windows build

jquorning:
windows32/64 are not legal for UTIL_OS. win32/win64 are.
This patch is not testet,

### DIFF
--- a/index/ut/utilada/utilada-2.0.0.toml
+++ b/index/ut/utilada/utilada-2.0.0.toml
@@ -26,10 +26,10 @@ UTIL_OS = "linux64"
 [gpr-set-externals."case(os)".macos]
 UTIL_OS = "macos64"
 [gpr-set-externals."case(os)".windows."case(word-size)".bits-32]
-UTIL_OS = "windows32"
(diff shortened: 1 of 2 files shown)


Fabien-Chouteau:
Looks good to me. @stcarrez do you confirm?

mosteo (approved):

mosteo:
I think this can be safely be merged, and tweaked/reverted down the line if we get further feedback from Stephane

stcarrez:
Yes, it was fixed.

OTOH, I have new releases of Ada Util but hadn't the energy the update and provide another pull request.

I also provide ready-to-use Debian packages and it turns out I found it easier.

Now, if you accept this pull, I may reconsider and put more energy into my Alire submissions.
Don't take this as a criticism, it's just that on my side it requires time and energy to provide an Alire update for all my projects.

mosteo:
@stcarrez , thanks for the feedback. Yes, I understand that your case is particularly time-consuming. We may have some improvements down the road for the case of monorepos (https://github.com/alire-project/alire/issues/737).

PR #284, opened 2021-06-09
dir_iterators 0.0.1

pyjarrett:
An iterator-based recursive directory walker and other iterator-based walker which traces up parent directories.

### DIFF
--- a/index/di/dir_iterators/dir_iterators-0.0.1.toml
+++ b/index/di/dir_iterators/dir_iterators-0.0.1.toml
@@ -10,6 +10,6 @@ maintainers-logins = ["pyjarrett"]
 tags = ["dir", "files", "walk"]
 
 [origin]
-commit = "89a8d0991ee6bb7f372783c57fa3025460346c9b"
+commit = "2e1fcc63ab6a31437096d352d5fe88f3702c1a77"
 url = "git+https://github.com/pyjarrett/dir_iterators.git"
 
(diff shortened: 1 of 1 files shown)


mosteo:
Hi @pyjarrett , thanks for your contribution. Given the [error](https://github.com/alire-project/alire-index/pull/284/checks?check_run_id=2781372315#step:7:366), I guess you'll need to narrow down the compilers that are able to build your code. For that you need to add dependencies on more precise versions of the compiler.

If you have the time, you can play with the docker images at https://hub.docker.com/repository/docker/alire/gnat. Alternatively just give the version that you're using as the minimum.

Unfortunately there's not an easy way yet (https://github.com/alire-project/alire/issues/161) to test unindexed crates in all these docker images. You'd need to launch the docker manually, passing the alr binary and your crate folder via `-v` arguments. As I said, time consuming...

pyjarrett:
It's cool, I'll figure it out, I'm use to this sort of thing.  It's been impressive that everything has been so smooth so far to get to this point :)

pyjarrett:
@mosteo Ok, I actually found and fixed two different errors on Ubuntu and Debian (same error both places).  I didn't see a macos image, but hopefully it'll work since it's just using stock Ada 2012 now.

mosteo:
Right, for macOS and Windows we rely on the GH Actions environments. Let's see what happens.

mosteo:
Merged, thanks.

PR #288, opened 2021-06-16
Add progress_indicators 0.0.1... and 0.0.2 

pyjarrett:


index/pr/progress_indicators/progress_indicators-0.0.1.toml
```diff
@@ -0,0 +1,15 @@
+name = "progress_indicators"
+description = "Ways of displaying progress in command line tools."
+version = "0.0.1"
+website = "https://github.com/pyjarrett/progress_indicators"
+authors = ["Paul Jarrett"]
+licenses = "Apache-2.0"
+
+maintainers = ["Paul Jarrett <jarrett.paul.young@gmail.com>"]
+maintainers-logins = ["pyjarrett"]
+tags = ["progress"]
```
mosteo: What do you think of having "command-line" as an extra tag?

mosteo (commented):
Looks good to me, see my comment about an extra tag.

index/pr/progress_indicators/progress_indicators-0.0.1.toml
pyjarrett: I actually had that in there originally as "command line" but it complained so I removed it.  I've added it back with the hyphen.

pyjarrett:
I've had to fix some usability things due to some preliminary testing from @onox (thanks onox!), so I'm going to cut this and resubmit.

PR #290, opened 2021-06-22
Add progress_indicators 0.0.1

pyjarrett:


index/pr/progress_indicators/progress_indicators-0.0.1.toml
```diff
@@ -0,0 +1,21 @@
...
+website = "https://github.com/pyjarrett/progress_indicators"
+authors = ["the progress_indicators authors"]
+licenses = "Apache-2.0"
+
+maintainers = ["Paul Jarrett <jarrett.paul.young@gmail.com>"]
+maintainers-logins = ["pyjarrett"]
+tags = ["command-line", "console", "terminal", "progress"]
+
+project-files = ["progress_indicators.gpr", "progress_indicators_tests.gpr"]
+executables = ["progress_indicators_test"]
```
mosteo: Hi @pyjarrett ; test projects/executables should not be included as they mean a distributed overhead on clients of your crate. They can remain in your sources so you can test locally, but they should not appear in the manifest. (I know some of our existing crates violate this guideline, if you got the idea there.)

mosteo (changes requested):

index/pr/progress_indicators/progress_indicators-0.0.1.toml
pyjarrett: Does this mean I should remove from `executables` or from `project-files` as well?

pyjarrett: I missed "projects/" when I read it the first time... fixed.

Fabien-Chouteau: > Does this mean I should remove from `executables` or from `project-files` as well?

Yes because your crate is a library, people that will use this library don't need or want to build the test executable.
I suggest that you put your test gpr in a subdirectory like `tests/` and don't make it visible in the Alire manifest.

pyjarrett: I feel like this is a missed opportunity for an equivalent of CPAN's common `make check`, but this is still achievable manually under these constraints.

Fabien-Chouteau: We may have that in the future, but this is not the way to do it.

mosteo:
Thanks, @pyjarrett . Merging now.

mosteo (approved):

index/pr/progress_indicators/progress_indicators-0.0.1.toml
ohenley: I fought that decision yesterday. In my case, this is not true:

> Yes because your crate is a library, people that will use this library don't need or want to build the test executable.

As a package consumer, to concretely assess the general quality of `progress_indicators` before committing to building on it, I have the following choices:

1) build `progress_indicators` and unroll my own exec wrapping libProgress_indicators.
2) build `progress_indicators_test` and get a feel of what it does, if at least the test is not broken at runtime etc.
3) code review only.

Now, 

1) For obvious reasons, someone just wanting to assess an unknown package, this is way too much work. I am just shopping at this point, not committed to work around an unknown individual's work. (nothing personal @pyjarrett)
2) as-is, building `progress_indicators_test` is broken. 
    a) Trying alire fails to build it. 
    b) Because the .gpr 'withs' dependencies are not relative to the base package, nor provided along `progress_indicators` in an 'orthodox' way, gprbuild fails. 
    
    You either have to :
    a) hack alire configs to make it build.
    b) swap the 'withs' .gpr dependencies for absolute paths (to alire's cache of given dependency) or install env. paths to the dependencies so gprbuild can proceed.

    IMHO, a) b) are major friction. At the instant I got to fight that yesterday, I was feeling miserable and missed git submodules. Don't get me wrong, I really appreciate Alire's work. I also need to underline that, IMHO, this curating, as-is, is a hard sell.
3) please buy my house from pictures only...
    
Maybe, what is missing is clear communication to lib producers that they should provide an additional demos/tests crate if they are not 'allowed' to offer lib client executables alongside the main crate. But doing so will probably trigger maintenance complaints and would be hard to enforce anyway.

onox: After pyjarrett/progress_indicators#5 has been merged, you can `cd tests` and then do `alr run` to run the tests. It uses a pin to the progress_indicators crate in the root directory of the repository, so you need a recent version of Alire.

pyjarrett: I wasn't going to mess with anything until [alire custom tasks](https://github.com/alire-project/alire/issues/496) or whatever had been implemented.  I didn't see a reason to do a bunch of custom work.

ohenley: @pyjarrett the matter does not concern your package only, just a good opportunity to raise awareness that this canonic case, an initial 'repo intent/ implicit contract' --> providing a buildable lib .gpr and a test .gpr <-- gets stopped in its track no matter how you try to build the test project, within the confine of normal expectations: 

"Hey if nothing is said about downloading/installing dependencies and it flies over a package manager, I should surely get away with building what is available here, in this repo. Using alr, nop, using gprbuild, nop, ok git clone recursive ... jeez nop. Ok lets study alr so I can unroll this trivial build myself... or should I just hack gpr deps ... or should I store ansiada somewhere and point gprbuild to it? I just want to see that guys code spinning dear god, now instead I have a little furry toolchain puzzle showing its head, yay!". 

I am repeating myself here only to share that the whole experience becomes, again in my humble opinion, asymmetric and ... I would respectfully use the term 'opiniated'. We remember that the master plan is for alire to take over, so better question the matter now than later.

Thx for considering my two cents.

pyjarrett: I get what you're saying.  It definitely highlights a need for this behavior to be supported.

Look at the date of this though, I barely knew any Ada at this point--I left the gpr because I didn't want onox's work to go to waste and assumed that there'd be something to do in the future.  Yeah, maybe I should have cut it stuffed it into a branch.  I didn't mean to waste anyone's time.  I'm scrambling to fix things because I didn't realize someone would post an article which mentions all my projects on hackernews, much less it would be a popular post.  I can't imagine the Alire team is any less frantic than I am right now... :(

ohenley: > I left the gpr because I didn't want onox's ... Yeah, maybe I should have cut it stuffed it into a branch.

@pyjarrett No, personally I am super happy you left the test gpr. Actually, when people do not wrap their 'code' with some 'showtime' code I just leave, except if it's a must-have or It comes from a really reliable source. Even so ... if I am not super stoked to try/give time/attention to such a project I just "whatever". There is only 24h in a day to be curious so...

BTW, your article is awesome, I loved it. My personal best is "A Free-Function Focused C++, or a Strongly Typed C". You found what I tried to oneliner people wanting to 'go back to c' and those wanted to 'modern C++'. Thx for it.

pyjarrett: Alright.

Just keep in mind that you're probably going to need [trendy_terminal](https://github.com/pyjarrett/trendy_terminal) for UTF-8/VT100 sequences to work correctly on Windows.  I think [ansiada](https://alire.ada.dev/crates/ansiada) will also work.  Trendy terminal got released as a stopgap to enable other work though.  It has a bunch of work I haven't published yet (working on input editing), it might not be fully stable.

Fabien-Chouteau: @ohenley I do like the example based learning myself, others prefer to look at documentation. We don't have a definite workflow for either in Alire at the moment.

Even if I like to have examples, I still think that building executables alongside libraries is not a good solution. As the number of dependencies increases it will have a significant impact on build time. You also have the issue of "poluting" your `PATH` and `GPR_PROJECT_PATH` with executable and projects that may clash with other things.

In my opinion we have a possible solution with `alr-1.1.0-rc1` and its new pin system. Crates should have a `example` sub-dir that provides an example crate. This example crate doesn't have to be contributed to the community index, and actually should probably not be contributed.

When I want to try a crate, I just have to:
```console 
$ alr get new_fancy_crate
$ cd new_fancy_create*/example
$ alr run
```

ohenley: @Fabien-Chouteau sure having a graph of dependencies that each produce their tests, examples, tutorials is wrong if your end goal is an 'end app'. Nevertheless, the diversity of intent in a repo also sketches the need for what I would illustrate as "profiles" of build chains. I want to build lean only libs under me, I want to build all deps that leads to this test project, I want to build 10 tutos, etc.

A solution is a great news! I totally agree snippets, tests, examples should not be in the index but yeah, we agree they should also viably live where they are intended to be. 

Thanks for the feedback.

ohenley: @Fabien-Chouteau 'pin' concept, is it that any folder can have an alire config but configs marked 'pinned' will not be indexed? Thx.

Fabien-Chouteau: We can try to look at what others do. But I don't know any package manager to provide a special feature for examples.

For pins see:
https://github.com/alire-project/clic/blob/ac734659560ffa98346c4ddc97d8966471d70374/example/alire.toml#L12
You can ask Alire to get a dependency from a local folder instead of the index.
It's useful for tests, examples, and multi-crate repos.

onox: > You can ask Alire to get a dependency from a local folder instead of the index.
> It's useful for tests, examples, and multi-crate repos.

That's what I did at pyjarrett/progress_indicators#5

mosteo: > It's useful for tests, examples, and multi-crate repos.

And simultaneous development of several interdependent projects. You can also pin repositories, so you can use unpublished changes until they get indexed.

PR #291, opened 2021-06-22
First release

heharkon:


mosteo (changes requested):
Hi @heharkon , thanks for your contribution.

We are trying to move away from the use of "ada" in package names, since in the context of an Ada package manager it doesn't add much. I understand it might not be easy to come with another name that it's not too generic (and most of us are guilty of this, if only because of past habits).

In this case, I would suggest at least to use `cmd_ada` instead, as this way at least it is alphabetically at their proper place instead of in a bunch of `ada_whatever`.

index/ad/ada_cmd/ada_cmd-0.0.1.toml
```diff
@@ -0,0 +1,14 @@
+name = "ada_cmd"
+description = "Command based text UI, inspired by Python cmd module"
+version = "0.0.1"
+
+authors = ["Henrik Härkönen"]
+maintainers = ["Henrik Härkönen <heharkon@iki.fi>"]
+maintainers-logins = ["heharkon"]
+tags = ["ui"]
```
mosteo, line 8: I would add here `command-line`

heharkon:
> In this case, I would suggest at least to use `cmd_ada` instead, as this way at least it is alphabetically at their proper place instead of in a bunch of `ada_whatever`.

I was also actually thinking this while pushing this, it makes sense! I'll rename it and add that one tag also.

heharkon:
Starting from scratch, renaming all was so intensive process.

PR #297, opened 2021-07-11
Add crate for asfml: Ada binding to Simple and Fast Multimedia Library

mgrojo:


mgrojo:
Windows requires manual download of [CSFML](https://www.sfml-dev.org/download/csfml/), so the build check will surely fail. Can that be solved using Alire? In macos, I cannot test, but it's probably the same.

mosteo:
> Windows requires manual download of [CSFML](https://www.sfml-dev.org/download/csfml/), so the build check will surely fail. Can that be solved using Alire? In macos, I cannot test, but it's probably the same.

The first stop should be to check if this is available for msys2. I just did and it seems so: https://packages.msys2.org/package/mingw-w64-x86_64-csfml?repo=mingw64

So you can try to add that in your libcsfml external (see e.g. https://github.com/alire-project/alire-index/blob/stable-1.0/index/li/libglfw3/libglfw3-external.toml ) and we'll see if this satisfies the tests.

mosteo:
(Just to clarify: msys2 is the solution we use on Windows for lack of an official system package manager.)

mosteo (changes requested):
Your external manifest filenames must end with `-external.toml`.

mgrojo:
I've added msys2 and arch, renamed the external as suggested, and removed the dependency to `libglu1-mesa`, which is only available in Debian/Ubuntu, but it is nevertheless automatically installed by the system package manager.

Is macOS using Homebrew as package manager? I've also found [csfml](https://formulae.brew.sh/formula/csfml) there, but don't know if that is used in the case of macOS. I'll add it too, if confirmed.

mosteo:
> Is macOS using Homebrew as package manager?

No, unfortunately at this time we do not support any PM for macOS.

mosteo:
The current failure seems unrelated to your changes. I'll see what happens.

mosteo:
In the end, it was the external new file that was misplaced. I corrected it for you since I had already tracked down the problem. An lo and behold, it builds on Windows :)

mosteo (approved):

mgrojo:
> In the end, it was the external new file that was misplaced. I corrected it for you since I had already tracked down the problem. 

Thanks. I take mental note for any possible future contribution.

> An lo and behold, it builds on Windows :)

Oh, Alire makes that so easy!

Fabien-Chouteau:
> Is macOS using Homebrew as package manager?

I'd love to have an Homebrew support in Alire like we did for msys2. But I don't have much time to work on it at the moment, and I am not a real macOS user.

mosteo:
> I am not a real macOS user.

I am not either, but I've got hold of an old mac at work so maybe when we run out of urgent issues I can take a stab at it.

PR #298, opened 2021-07-14
Add blake2s 0.1.0

lkujaw:
Hi,

This is only the library component, as I want to keep the existing Makefile setup unified if possible. I'm not too familiar with gprbuild, so if there's an easy way to run the test program as part of the build, please let me know.

Thanks, Lev

### DIFF
--- a/index/bl/blake2s/blake2s-0.1.0.toml
+++ b/index/bl/blake2s/blake2s-0.1.0.toml
@@ -1,0 +1,15 @@
+name = "blake2s"
+description = "SPARK83 implementation of the BLAKE2s hash function"
+version = "0.1.0"
+
+authors = ["Lev Kujawski"]
+maintainers = ["Lev Kujawski <int21h@mailbox.org>"]
+maintainers-logins = ["lkujaw"]
+licenses = "MIT-0"
+website = "https://github.com/lkujaw/blake2s"
+tags = ["spark", "hash", "blake2", "blake2s"]
+
+[origin]
+commit = "822db20dbfd078f05d2d2048bf6af24c65fc37fc"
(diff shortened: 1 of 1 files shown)


mosteo:
Thanks for your contribution, @lkujaw . It seems that the default checkout on Windows is preserving line terminators, which you are checking in your style settings:

```
octets.ads:1:79: (style) incorrect line terminator
```

If this check is not critical to you, you could disable it with `-gnaty-d`. Otherwise, I guess the repository should include [configuration on how to treat line terminators](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings).

mosteo:
> This is only the library component, as I want to keep the existing Makefile setup unified if possible. I'm not too familiar with gprbuild, so if there's an easy way to run the test program as part of the build, please let me know.

Actually we prefer that test programs are not run by default. We may add features to perform self-tests in a separate way in the future.

lkujaw:
Thanks for the feedback. I decided to remove the CRLF check altogether; hopefully that will make working with the source code easier on those platforms.

mosteo (approved):

mosteo:
Merged, thanks.

Fabien-Chouteau:
@lkujaw I see that you are using the gpr file generated by Alire. The default in the gpr file is to enable all checks and compile in debug mode. You can change that to avoid compilation issue like the CRLF thing for the user of your crate, and also compile in "optimize" mode for better performance.

Maybe we should change the defaults for the gpr generated by `alr`. I did it this way for a better programming experience (debug enabled, all checks to catch your errors), but then it's bad when the crate is published in the index...

PR #302, opened 2021-07-23
Rename GNAT crates

Fabien-Chouteau:
I changed the names of the crates to what I think is more meaningful.

I  merged the Linux, Windows and macOS under a gnat_native that will be
easier to understand and use in my opinion.

I also added a licenses field.

One thing that we may want to improve is the values for the case
statements, for instance `[origin."case(os)".macos."case(word-size)".bits-64]`
will soon be a blocking point to support macOS on the M1.

Instead of word size we probably need something like: "case(CPU)" or
"case(arch)". Or even use full platform triplets: "case(host_triplet)".


Fabien-Chouteau:
@mosteo I didn't add the `provides` in `gnat_external` because I don't know how to write it. The others have a version in it `gnat=10.3.0-1` but what do we do for external where the version is computed from the compiler in PATH?

mosteo:
> Instead of word size we probably need something like: "case(CPU)" or "case(arch)". Or even use full platform triplets: "case(host_triplet)".

I think I like better `arch`, as the triplet conflates everything. But I defer to your better judgement on all things embedded.

> @mosteo I didn't add the provides in gnat_external because I don't know how to write it. The others have a version in it gnat=10.3.0-1 but what do we do for external where the version is computed from the compiler in PATH?

I prepared this just yesterday. Inside the `[[external]]` section you can add a `provides="gnat"` and it will use the detected version.

mosteo:
I've added the missing provides so this one can be merged already.

mosteo (approved):

PR #304, opened 2021-07-28
muntsos_blinky release 1.0.0

pmunts:
Alire crate for building a blinky application for MuntsOS Embedded Linux.

### DIFF
--- a/index/mu/muntsos_blinky/muntsos_blinky-1.0.0.toml
+++ b/index/mu/muntsos_blinky/muntsos_blinky-1.0.0.toml
@@ -1,0 +1,27 @@
+name = "muntsos_blinky"
+description = "MuntsOS Embedded Linux Blinky Test"
+version = "1.0.0"
+licenses = "BSD-1-Clause"
+website = "https://github.com/pmunts/muntsos"
+
+authors = ["Philip Munts"]
+maintainers = ["Philip Munts <phil@munts.net>"]
+maintainers-logins = ["pmunts"]
+
+project-files = ["muntsos_blinky.gpr"]
+
+executables = ["blinky"]
+
+tags = ["embedded", "linux", "muntsos", "blinky", "libsimpleio", "led",
+"beaglebone", "pocketbeagle", "raspberrypi"]
+
+[available."case(os)"]
+'linux' = true
+"..." = false
+
+[origin]
+hashes = [
+"sha512:7cafe868e8dfda555ab9d4832c98e5404ae9b2b6d1cc08d09199a328239931045c9aae40c6fc725a969635f020b54458c412a459aed5d7e7566b655a024d659e",
+]
+url = "http://repo.munts.com/alire/muntsos_blinky-1.0.0.tbz2"
+


mosteo:
Hi Philip, thanks for the PR.

We generally do not expect test crates to be added to the community index (you can still have a nested test crate, see [here](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing), that is not exposed to clients). On the other hand, if these serve as examples to learn about the main crate, that would be OK as `libsimpleio_static_examples` or `libsimpleio_static_demos`.

I'm also not sure about the `static` part in the name. If you intend to have different build modes, you can manage that via configuration variables and have a single `libsimpleio` crate. This also will prevent conflicts by clients depending on two different versions of the library.

I've launched the checks though so you can already test the build of this PR.

pmunts:
In this case, "_static" means "use libsimpleio.a in the crate instead of the system shared library libsimpleio.so".

The Linux Simple I/O Library example application I chose to put in a crate is named test_hello.adb.  I can rename the crate to libsimpleio_static_example_hello.

Fabien-Chouteau:
Hello @pmunts ,

> In this case, "_static" means "use libsimpleio.a in the crate instead of the system shared library libsimpleio.so".

You shouldn't have to do that because there is already a setting for the library kind in your project file:
https://github.com/pmunts/libsimpleio/blob/b6358a2d7c3087e30867127b68c0d732580ce68c/ada/alire/libsimpleio_static/libsimpleio.gpr#L17
 
> The Linux Simple I/O Library example application I chose to put in a crate is named test_hello.adb. I can rename the crate to libsimpleio_static_example_hello.

It would be better to have `libsimpleio_examples` and then you will be able to add more examples in this crate in the future.

pmunts:
I'm trying to create a workable namespace scheme for my crates.  Soon there will be another crate called libsimpleio_examples (or maybe just libsimpleio_example_hello that does use the system libsimpleio.so.

I need some way to differentiate between crates that build libsimpleio.a (actually liblibsimpleio.a) on the fly (by depending on libsimpleio_static) and those that use the system libsimpleio.so.

pmunts:
Also, I put:

[available."case(os)"]
'linux' = true
"..." = false

in my .toml files to indicate the crates are specific to Linux, but that is not copied to the published manifest file.  How am I supposed to mark crates as Linux only?

pmunts:
How about libsimpleio_standalone instead of libsimpleio_static?

Fabien-Chouteau:
> I'm trying to create a workable namespace scheme for my crates. Soon there will be another crate called libsimpleio_examples (or maybe just libsimpleio_example_hello that does use the system libsimpleio.so.
> 
> I need some way to differentiate between crates that build libsimpleio.a (actually liblibsimpleio.a) on the fly (by depending on libsimpleio_static) and those that use the system libsimpleio.so.

You should not try to have two crates that provides the same project.
You can either have:
 - A crate that provides libsimpleio from sources (C and Ada)
 - A crate that provides Ada bindings only for the system's libsimpleio (external)

Looking at http://repo.munts.com/alire/libsimpleio_static-1.19805.1.tbz2 I think the first solution is the best here.

mosteo:
> Also, I put:
> 
> [available."case(os)"]
> 'linux' = true
> "..." = false
> 
> in my .toml files to indicate the crates are specific to Linux, but that is not copied to the published manifest file. How am I supposed to mark crates as Linux only?

This is unexpected, as the `publish` command should only be adding the `[origin]` bit to your local manifest. Anyway, you can add that bit to the file generated by the publishing assistant before submitting it for the PR. (Or correct it if already in an open PR.)

Incidentally, my Linux (Ubuntu 20.04) does not find anything when searching for `simpleio`. Is this a commonly packaged library? (Thinking on whether an external definition may make sense or not.)

pmunts:
The Linux Simple I/O Library aka libsimpleio is a library I have written that tries to wrap the ugliness of Linux embedded systems I/O (e.g. GPIO) in some smallish C wrapper functions that are easy to call from Ada, Free Pascal, .Net and other languages.  Source code is available at https://github.com/pmunts/libsimpleio and Debian 10 (Buster) packages are available at http://repo.munts.com/debian10.  libsimpleio isn't in the regular Debian package repository, so I don't know how to make an external dependency to it.

The libsimpleio source code tree includes an Ada library project for a binding to libsimpleio, but the Ada parts are not compiled into the shared library libsimpleio.so.  The precompiled Ada library project is included in the binary packages for libsimpleio.

The muntsos_blinky program crate is for Munts OS Embedded Linux, another of my projects.  libsimpleio is always part of MuntsOS, and libsimpleio.so is included in everty MuntsOS kernel image and a precompiled Ada library project is part of every MuntsOS cross-toolchain.  So the muntsos_blinky crate does not depend on libsimpleio_static or any other crate, and libsimpleio.gpr is referenced in muntsos_blinky.gpr but NOT in muntsos_blinky.toml.  When doing a cross-build, gprbuild is able to find the proper libsimpleio.gpr without any help, because libsimpleio.gpr is installed where gprbuild expects to find it.

Your build hosts won't ever be able to build muntsos_blinky unless you install the MuntsOS cross-toolchains.

I would like to have a library crate named libsimpleio_static (or better, I suppose, libsimpleio_standalone) and an example program crate called libsimpleio_standalone_examples that depends on libsimpleio_standalone.  NEITHER of those two crates depend on any of the libsimpleio Debian packages or libsimpleio.so.

Your LINUX (NOT MacOS or Windows) build system should be able to build both libsimpleio_standalone and libsimple_standalone_examples.

AND I would ALSO like to have another program crate libsimpleio_examples that, like muntsos_blinky, does require libsimpleio to be installed, does use the precompiled libsimpleio for Ada library project, and does use libsimpleio.so.  In fact, muntsos_blinky and libsimpleio_examples are very similar except that the latter crate does not include the MuntsOS stuff (default.cgpr and the application note PDF file).  And libsimpleio_examples will have the same test program as libsimpleio_standalone_examples.

Your build hosts won't ever be able to build libsimpleio_examples unless you install native libsimpleio.

The reason for all this is that there are THREE similar but different pathways for using libsimpleio:

1. If you want to develop native Linux programs, and are willing to install the libsimpleio Debian packages, you start with alr get libsimpleio_examples.  (You can actually cross-compile for MuntsOS, too, if you manually link or copy default.cgpr.  In fact you ought to be able to use any cross-GNAT toolchain if you can find or generate a default.cgpr for it.)  There are some important advantages to installing libsimpleio, such as hotplug scripts that set device permissions and a precompiled libsimpleio.gpr.

2. If you want to develop native Linux programs, and DON'T want to install the libsimpleio Debian packages, you start with alr get libsimpleio_standalone_examples.

3. If you want to develop cross-compiled programs for MuntsOS embedded Linux, and have installed the MuntsOS cross-toolchains, you start with alr get muntsos_blinky.

So, please help me accomplish all this...

mosteo:
Thanks for the detailed info, Philip. I have still to digest it all but we'll come up with a solution, if what Fabien suggested is not enough.

General musings I had before reading your post: in the Rust world, for example, system libraries are bound via `name-sys` crates. We simply do not add the `-sys` part, and indeed I do not see a compelling reason to differentiate them in the general case. Also, up to now, we haven't faced a situation where it was important to differentiate between a system-provided lib and a from-source implementation of the same, so it never has been important to have two crate names for the same project.

Now, with the new `provides` feature that is going to be available soon, we could have for example a `libfoo_external` and `libfoo_src`, both providing `libfoo`. So a dependent has the choice of using whichever is available via `libfoo`, or a specific one via the `libfoo_xxx` specific name.

I'm not saying this is what you need, just commenting ideas I had. I'll come back after I have thought a bit about your last post.

Fabien-Chouteau:
> The reason for all this is that there are THREE similar but different pathways for using libsimpleio:

For your first scenario, if users are willing to install the libsimpleio Debian packages, then you should also provide the Ada binding as Debian package and this will be a full Debian workflow for your users. I am not sure why Alire would be involved in this scenario.

Your second scenario seems like a full Alire workflow everything required is in the crate. This makes more sense to me.

The third scenario sounds similar to the first one, if you have your own custom toolchain that comes with libsimpleio packaged, why not also put the Ada package in there?

I think you have to pick which one of the scenario is going to be available in the Alire community index. Having three ways to do the same thing is not good, it's going to be confusing for users and more work for you to maintain.

mosteo:
You're touching on several things that Alire can't do at the moment. One, we don't have yet a way to prioritize crates that provide the same version, e.g., from sources or from a system package. So even if you add an external for your debs, the user won't have control over which implementation is chosen by the solver.

Another, we don't have a way to specify conditional dependencies besides our predefined environment variables. So it is not yet possible to omit a dependency based on an arbitrary condition, or on the target.

You may want to watch https://github.com/alire-project/alire/pull/775 to see how cross-target compilers are finally integrated. This is still a bit in flux.

As things stand, I cannot think of an immediate satisfactory solution. If you try to go with different crates you'll have an explosion of combinations, packaging the same code under different names just because of differing dependencies. I'd rather not have that, and find another way.

pmunts:
I freely acknowledge I'm probably trying to leverate Alire in ways you didn't imagine.

When I work with .Net, I'm a heavy user of NuGet.  I have 4  packages in NuGet that are built from the libsimpleio code base: libsimpleio, which is a library package for Linux, exactly analgous to the libsimpleio_standalone (or whatever) I want to publish in Alire; libremoteio, which is a library package, which implements the Remote I/O Protocol and works on Linux, MacOS and Windows, and two corresponding project template packages: libsimpleio-templates and libremoteio-templates.

I have found project templates to be a very elegant way to encapsulate the arcane details of an embedded software project and mostly hiding them from the developer.

NuGet is a tremendous force-multiplier for software development, as it is very easy to find and use software components.

So part of what is going on here is that I view Alire as "NuGet for Ada", whether you intended that or not.

So, as I see it, muntsos_blinky (cross) and libsimpleio_standalone_exampes (native) are project template crates, exactly analgous to libsimpleio-templates.  .Net doesn't have "native" or "cross" projects, so there is only  one project template package in NuGet.

muntsos_blinky = project template crate for MuntsOS Embedded Linux targets
libsimpleio_standalone = library crate for Linux targets
libsimpleio_standalone_examples = project template crate for Linux targets
(future) remoteio = library crate for Linux, MacOS, and Windows targets
(future) remoteio_examples = project template crate for Linux, MacOS, Windows targets

What I would like for you to do now is merge muntsos_blinky, even though you can't build it. Then I am going to rename libsimpleio_standalone to just libsimpleio and republish it with a separate PR.  Eventually I will publish remoteio with a separate PR as well.

I will defer the example crates for now until I have done some more thinking.  They might just degenerate to HOWTO's in future versions of the library crates.

Thank you very much for your efforts in all of this.

pmunts:
Will update the documentation inside the crate and republish at a later date.

Fabien-Chouteau:
> I have found project templates to be a very elegant way to encapsulate the arcane details of an embedded software project and mostly hiding them from the developer.

I agree, we should have template crates on the TODO list for Alire.

> So part of what is going on here is that I view Alire as "NuGet for Ada", whether you intended that or not.

They are certainly comparable, but not exactly the same.

PR #309, opened 2021-08-04
remoteio release 1.19928.2

pmunts:


index/re/remoteio/remoteio-1.19882.1.toml
```diff
@@ -0,0 +1,29 @@
...
+tags = ["embedded", "linux", "remoteio", "adc", "dac", "gpio", "i2c", "motor",
+"pwm", "sensor", "serial", "servo", "spi", "stepper"]
+
+[available."case(os)"]
+'linux|windows' = true
+"..." = false
+
+[[actions."case(os)".windows]]
+type = "pre-build"
+command = ["sh", "-c", "mkdir -p ./bin && cp src/win64/*.dll ./bin && mkdir -p ./lib && cp src/win64/*.dll ./lib"]
```
mosteo: Hi @pmunts, sorry for the long wait. If this action is needed to run only once, it should be better a `post-fetch` one. Also we currently have a bug in which pre-build actions aren't run for dependencies.

mosteo (changes requested):

mosteo:
Also, it seems some external dependencies will be needed given the check failure.

pmunts:
I have tried to specify a dependency upon the Debian or Ubuntu `libhidapi-dev` package, but I can't get it to work:

`error: Cannot continue with invalid session:
error:    Failed to load /export/home/pmunts/src/alire/remoteio/alr-hcfm.tmp/alire.toml:
error:    Cannot load manifest remoteio, please review contents:
error:    Loading release from manifest: /export/home/pmunts/src/alire/remoteio/alr-hcfm.tmp/alire.toml:
error:    invalid property: external
`
I have tried the following, copied from the libzmq crate:

`[[external]]
kind = "system"
"debian|ubuntu" = ["libhidapi-dev"]
`

and also this shortcut version:

`[[external."case(os)".linux]]
kind = "system"
hint = "This crate depends on the libhidapi-dev Debian package."
origin = ["libhidapi-dev"]
`

Both fail with the same pretty worthless error message above.  I've tried alr -d -vv but that didn't yield anything useful.

mosteo:
I'm understanding that you tried to add the ``[[external]]`` definition to your own manifest. Instead of doing so, what you need is a separate file with the external definition in your local index.

But in truth, to work locally, you still don't need externals. You just install with your Linux package manager whatever you need (and take note of which libraries those are, for the time of publishing).

It is at the time of crate submission that Alire must know all dependencies and how to install them. So, for example in this case, we have that
```
remoteio --> libhidapi
```
The `remoteio` manifest is just a plain one, like you have submitted, that must have a `depends-on` on `libhidapi`. For `libhidapi`, as this comes from system packages, you need a second manifest which, instead of being for a plain Ada project, is for a system package, which is described using the external special syntax: https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#external-kinds-system-packages

You can test it locally by adding the external to your local index (and the dependency to your manifest), and checking that `alr update` finds the package for your dependency, or wait for the checks in the PR.

An example of a plain Ada crate that relies on a system package is https://github.com/alire-project/alire-index/tree/stable-1.0/index/zl/zlib_ada . You'll see that it depends on `zlib`, which is the external one.

As a side note, I guess we could implement pins for system packages, to simplify working with this kind of dependency before they exist in the index. But, in any case, the external manifest would be needed for the published crate in the community index.

mosteo (approved):

PR #312, opened 2021-08-05
mcp2221 release 1.19928.2

pmunts:


Fabien-Chouteau:
Hello @pmunts,

I don't understand the name of this crate because as far as I can see the sources are the same as libsimpleio and there many more drivers than just mcp2221.

pmunts:
I didn't mean to close it.

Fabien-Chouteau:
Hello @pmunts, you didn't answer my comment above.

pmunts:
Both the **mcp2221** and **remoteio** crates are subsets of libsimpleio.  The most important difference, however, is that **mcp2221** and **remoteio** can run on Linux and Windows (and potentially MacOS, though I no longer have a Mac to test on) while **libsimpleio** is strictly Linux only.

Fabien-Chouteau:
What is the difference between this `mcp2221` crate and the `remoteio` one?

From what I can see `mcp2221` is `remoteio` plus the mcp2221 drivers.
So why make a separate crate for that and not put the mcp2221 drivers in `remoteio`?

pmunts:
The difference between mcp2221 and remoteio is the underlying communication protocol.  The MCP2221 raw HID protocol is defined by Microchip and implemented on the MCP2221 and MCP2221A chips, which are in fact preprogrammed PIC16F1455 microcontrollers.  The MCP2221 protocol supports GPIO, I2C, 10-bit A/D, and 5-bit D/A and nothing else.

The Remote I/O Protocol was **inspired** by both the MCP2221 and Labview Linx protocols, and I developed it out of frustration with the limitations of those.  It is a far more general,  regularized, and efficient protocol, and can be used over additional kinds of transport channels, such as serial ports, UDP, ZeroMQ, etc.  It also supports more kinds of devices, such as SPI and PWM.

The only real advantage of the MCP2221 over the Remote I/O Protocol is that the MCP2221 device provides a USB serial port as well as the raw HID device for GPIO.  In all other respects the Remote I/O Protocol is far superior.  The MCP2221 Ada code only exists because I wanted to make fair performance comparisons between the two protocols.  There is a Linux kernel module available for the MCP2221, but it is incredibly slow in practice, apparently because of the interaction between kernel locking or scheduling and the excess baggage of the MCP2221 protocol.

Fabien-Chouteau (approved):
Thank you @pmunts I understand now. I think there would be a way to factorize.

pmunts:
Thanks for merging this crate.  But don't forget remoteio, too.

I have thought long and hard about how to factorize the libsimpleio project.  But the intersection of how alire works, how gprbuild library projects work, and how Ada works, seems to make it impossible.

For example, I thought about splitting libsimpleio into libsimpleio_bindings, libsimpleio_interfaces, libsimpleio_objects, libsimpleio_devices, etc.  Then mcp2221 would become libsimpleio_mcp2221 and remoteio would become libsimpleio_remoteio.

But the set of devices supported by mcp2221 is smaller than either of remoteio or libsimpleio (For example, the MCP2221 provide an SPI master while both libsimpleio and remoteio do, so the mcp2221 crate can't support SPI devices.).

This is one of my longstanding problems with gprbuild over gnatmake: gprbuild trys to compile everything, especially for a library project, while gnatmake only compiles what is necessary for a particular program.  I don't know how to get around this, unless you were to add a source-only crate feature to alire.

PR #317, opened 2021-08-18
felix 0.2.0

lkujaw:


lkujaw:
Hi,
I seem to have run into a problem with the Windows build (although it shows as passed): I need the nl_types.h header from https://packages.msys2.org/package/msys2-runtime-devel but I cannot add a dependency with hyphens apparently.

```
[[depends-on]]
[depends-on.'case(os)'.windows]
msys2_runtime_devel = ">=3.0"
```

Could you please give me some advice on how to resolve this?

Thanks, Lev

mosteo:
Hi, Lev. As `msys2_runtime_level` is not yet known to Alire, you need to provide an external definition for it so it can be detected and installed. See for example https://github.com/alire-project/alire-index/blob/stable-1.0/index/li/libglfw3/libglfw3-external.toml

That file is basically a mapping from the Alire name (that as you saw cannot contain hyphens) to possible system packages. So: 

```
[[external]]
kind = "system"
   msys2 = ["msys2-runtime-devel"]
```

index/ca/catgets/catgets-external.toml
```diff
@@ -0,0 +1,10 @@
+name = "catgets"
```
Fabien-Chouteau, line 1: Hi @lkujaw,

I am not sure about this one. Looks like catgets is part of POSIX and therefor not

Right now the Windows part of Alire aims at building applications that are native Windows and not cygwin application.
If you are using Alire in msys2 the `msys2-runtime` package will always be here.

My feeling is that this is getting outside the scope of Alire and you should not have an explicit dependency on either catgets or msys2-runtime.

PR #318, opened 2021-08-19
catgets

lkujaw:
Hi Alejandro,
I decided to call this package catgets, since I've only ever seen X/Open label it the "Native Language System" within their specifications, which is a bit vague IMO. Thanks for the help, and if you'd prefer a different name, please let me know.

### DIFF
--- a/index/fe/felix/felix-0.2.0.toml
+++ b/index/fe/felix/felix-0.2.0.toml
@@ -1,0 +1,27 @@
+name = "felix"
+description = "X/Open Native Language System (NLS) for Ada"
+version = "0.2.0"
+
+authors = ["Lev Kujawski"]
+maintainers = ["Lev Kujawski <int21h@mailbox.org>"]
+maintainers-logins = ["lkujaw"]
+licenses = "MIT-0"
+website = "https://github.com/lkujaw/felix"
+tags = ["ada1995", "i18n", "nls",
+        "localization", "localisation", "l10n"]
+
+[[depends-on]]
+[depends-on.'case(os)'.windows]
+msys2_runtime = ">=3.0"
+
+[gpr-set-externals.'case(os)']
+'windows' = { SYSTEM_API = "WIN32" }
+'...'     = { SYSTEM_API = "POSIX" }
+
+[environment.'case(distribution)']
+'msys2' = { C_INCLUDE_PATH.append = "${DISTRIB_ROOT}/usr/include/" }
+
+[origin]
+commit = "7409bde64df3467f6d8b6aceb96eba2aac920014"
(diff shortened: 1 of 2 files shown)


mosteo:
Hi, Lev, I'm a bit confused here. Is this because the `catgets` call is provided by the package? 

Anyway, what we normally do is to take the system package name, so I'd say here `msys2_runtime` would be the obvious choice.

Ah, I think I see where the confusion is coming, because the description doesn't seem to match the crate either. You need a separate crate for the external that describes the runtime (this one, fixing the description), and your crate that depends on this one (the one you submitted in #317, I guess). You can submit both in the same PR, to be sure everything works as intended.

Let me know if I'm off-track here.

Fabien-Chouteau (changes requested):
Hi @lkujaw,

(moving my message here from #317)

Right now the Windows part of Alire aims at building applications that are native Windows and not cygwin application.
If you are using Alire in msys2 the msys2-runtime package will always be here anyway.

My feeling is that this is getting outside the scope of Alire and you should not have an explicit dependency on either catgets or msys2-runtime.

lkujaw:
Hi,
Thanks for the feedback. I went ahead and changed the name of the external dependency to msys2_runtime and added the Felix package too per Alejandro's suggestion, so now the latter builds on all the Alire platforms (however, there's something with the 'Check PR' action that's eluding me.)

@Fabien-Chouteau, I agree that a dependency on the msys2 runtime is less than ideal, but I am uncertain about how to best address your concerns. There is a libcatgets package that works directly above mingw (see https://sourceforge.net/projects/mingw/files/MinGW/Extension/catgets/), but it was replaced in favor of the msys2 runtime library according to https://packages.msys2.org/package/msys2-runtime-devel.

Thinking of a way forward, is it only the Cygwin DLL dependency that is problematic or the use of the POSIX catgets interface on Windows? If it is the former, I could revert the Alire Windows platform support for Felix until I can re-implement catgets in Ada for Win32. One of my goals with Felix was to eliminate the burden on translators to learn yet another translation system (hence the implementation of a type-safe printf emulation package), and catgets is, to my knowledge, the only ISO-standardized string localization API.

Thanks, Lev

Fabien-Chouteau (approved):

Fabien-Chouteau:
Let's go with the msys2_runtime dependency, that will mean that this crate is only compatible with msys2 on Windows.

@mosteo looks like the "check authors" TOML parser doesn't like our format. I tested the content in an online parser and it works fine: https://pseitz.github.io/toml-to-json-online-converter/

So I think we are ok.

mosteo:
Thank you both, merging it now.

PR #322, opened 2021-08-30
Update VSS version

reznikmm:
This version includes grapheme clusters iterators and many more improvements.

### DIFF
--- a/index/vs/vss/vss-22.0.0-20210830.toml
+++ b/index/vs/vss/vss-22.0.0-20210830.toml
@@ -1,0 +1,20 @@
+name = "vss"
+description = "Advanced string and text manipulation with Unicode support"
+version = "22.0.0-20210830"
+tags = ["unicode", "json", "text"]
+
+authors = ["AdaCore"]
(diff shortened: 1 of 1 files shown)


reznikmm:
I've put compiler restriction in the crate, but CI still fails on old compiler. Any advice?

```
[[depends-on]]
gnat = ">=2020|(>=10 & <=2000)"
```

mosteo:
Well, it seems that gnat 10 is also unable to compile that. You might try upping the lower bound to 11? Which compiler are you working with?

reznikmm:
> Well, it seems that gnat 10 is also unable to compile that. You might try upping the lower bound to 11? Which compiler are you working with?

Indeed, GCC 10 lacks String_Literal indeed. I've updated crate constraint to GCC 11.

Thank you!

mosteo:
The current failure is my fault. Will fix shortly.

mosteo:
@reznikmm , if you rebase on top of `devel-1.1` again, we should see the result of your last change.

mosteo (approved):

PR #325, opened 2021-08-31
 libgpr-21.0.0: incompatibility with GNAT Community 2021

Fabien-Chouteau:


mosteo:
What I've done up to now is to merge into `devel-1.1` submissions to `stable-1.0` from time to time, so I think this one is not really necessary. However, it highlighted the need to fix our tests, so please don't close it yet.

mosteo:
Merged from `stable-1.0` so no longer needed, thanks.

PR #330, opened 2021-09-01
GNAT 11.2

Fabien-Chouteau:


Fabien-Chouteau:
@mosteo 
Check PR fails because of TOML parser
and Build crate because there is not gprfile to build in those crates.

mosteo:
> and Build crate because there is not gprfile to build in those crates.

I see. `alr get --build` should not attempt to build binary crates, noted.

mosteo (approved):

index/gn/gnat_native/gnat_native-11.2.0-1.toml
```diff
@@ -0,0 +1,28 @@
+name = "gnat_native"
+version = "11.2-1"
```
mosteo, line 2: Since the title/commit don't mention it... Is this meant to be a pre-release? I assumed so for the `10.3-1` ones.

Fabien-Chouteau: Because of the `-1`?
No it's because I want to be able to make several packing versions of the same GNAT release. In case something is broken in the way it is build or packaged.

mosteo: > Because of the -1?

Yup. I see the dilemma. The build part is unordered, and the patch number would imply that something has changed in the compiler. As long as people refer to the major number only, I guess there is no problem, although if these are truly ">=11.2", we are lying in a semver sense.

Fabien-Chouteau: @mosteo Since GCC is not using the patch revision number. We could use that instead of `-1`.

mosteo: Sounds good!

PR #334, opened 2021-09-05
Add septum

pyjarrett:


pyjarrett:
I ran into a host of issues about dependencies of the parent in the internal test crate in `tests/`, I probably didn't do it correctly since I just re-added all the parent dependencies.

mosteo:
The error about not finding the config file makes me think this could be due to needing the devel branch. You could try to rebase on top of `devel-1.1` and change the base branch of the PR.

pyjarrett:
I'm just leaving this here because I've been doing my own feature work, eventually, I'll get around to getting this to be able to merge.

PR #339, opened 2021-09-07
Port stable-1.0 into devel-1.1

mosteo:
Checks will fail at least because of the microbit crate


mosteo:
Merging once verified that the index loads properly, and the failure is the expected one.

PR #340, opened 2021-09-09
Fix version extraction in build checks

mosteo:
We were losing the pre-release part


mosteo:
Merging already as this is a trivial fix

PR #341, opened 2021-09-10
Disable toolchain installation in setup-alire@latest-devel

mosteo:
As we want to use the ones set up externally. At some point, for non-docker, non-distro tests, we should rely on the toolchain installed by the action instead.

### DIFF
--- a/.github/workflows/build-crate.yml
+++ b/.github/workflows/build-crate.yml
@@ -78,6 +78,8 @@ jobs:
     - name: Set up devel `alr`
       if: contains(github.base_ref, 'devel-')
       uses: alire-project/setup-alire@latest-devel
+      with:
+        toolchain: --disable-assistant # We want to use the external ones
 
     - name: Test crate (Linux)
       if: matrix.os == 'ubuntu-latest' # docker testing only for linuxes


Fabien-Chouteau:
> As we want to use the ones set up externally. At some point, for non-docker, non-distro tests, we should rely on the toolchain installed by the action instead.

Are you sure about that? If people use the Alire toolchains to build their crates we will have unexpected failure by using other versions of the compiler.

mosteo:
> Are you sure about that? If people use the Alire toolchains to build their crates we will have unexpected failure by using other versions of the compiler.

I'd like to keep checking with the distro compilers, at least in the distros that are actively maintaining them. Otherwise I think we'll be doing a disservice to the Ada reputation for portability.

Fabien-Chouteau:
> I'd like to keep checking with the distro compilers, at least in the distros that are actively maintaining them. Otherwise I think we'll be doing a disservice to the Ada reputation for portability.

Do we have tests with the Alire toolchain as well?

mosteo:
> Do we have tests with the Alire toolchain as well?

Not yet, but I will add those soon.

mosteo:
This one has to be merged now that https://github.com/alire-project/setup-alire/pull/18 is also merged. I'll enable the tests with the native Alire toolchains in a separate PR/workflow.

PR #343, opened 2021-09-10
Add ada_toml 0.3.0

pmderodat:


mosteo:
The error is puzzling. I don't think at this time that the problem is in you packaging, will investigate.

pmderodat:
Ok thanks! Even more puzzling to me ;-)

mosteo:
The problem is that I cannot replicate it locally even on the very same OS version...

mosteo:
OK, found the issue, for which I will submit a patch to our CI scripts shortly. I've tested your submission in my fork so it's safe to merge.

mosteo (approved):

pmderodat:
Awesome, thank you very much for the quick fix! :-)

PR #345, opened 2021-09-12
Adabots 1.0.1  bugfix release

TamaMcGlinn:
Fixed a really annoying bug where restart required a 1 minute wait for the socket to become available again.


TamaMcGlinn:
woops; added an unrelated commit. That will teach me, trying to use the web interface

PR #349, opened 2021-09-13
Port stable-1.0 into devel-1.1

mosteo:


mosteo:
Made obsolete by new submissions in the interim

PR #354, opened 2021-09-14
Build submissions with toolchains from Alire

mosteo:
Adds a new workflow that relies on the default toolchain from Alire.

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -1,0 +1,52 @@
+name: Build Crate with Alire Toolchain
+
+on:
+  pull_request:
+    paths:
+      - 'index/**.toml'
+
+jobs:
+
+  build:
+    name: ${{ matrix.os }}
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: 
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-native.yml
```diff
@@ -0,0 +1,69 @@
...
+
+    - name: Set up msys2 (Windows)
+      if: matrix.os == 'windows-latest'
+      run: ./alire_install/bin/alr --non-interactive version
+
+    - name: Install tar from msys2 (Windows) # Git tar in Actions VM does not seem to work)
+      if: matrix.os == 'windows-latest'
+      run: C:\Users\runneradmin\.cache\alire\msys64\usr\bin\pacman --noconfirm -S tar
+
+    - name: Test crate
```
Fabien-Chouteau, line 67: You don't do an `alr toolchain --select gnat_native gprbuild<2000`?

Fabien-Chouteau (approved):

.github/workflows/build-native.yml
mosteo: That's the default now for `setup-alire` in the devel branch. This way when the action finishes the compiler is already installed. Should we have an empty default?

Fabien-Chouteau: Ok, I forgot about the default :)

PR #355, opened 2021-09-14
audio_base 1.0.0

gusthoff:


mosteo (approved):

index/au/audio_base/audio_base-1.0.0.toml
```diff
@@ -0,0 +1,18 @@
+description = "Basic Ada packages for audio applications"
+long-description = """
+# Audio Base
+This crate contains basic Ada packages for audio applications.
+"""
+name = "audio_base"
+version = "1.0.0"
+website="https://github.com/Ada-Audio/audio_base"
+licenses = ["MIT"]
```
mosteo: The licenses are now a single SPDX string. While this is not a hard error (yet), it's probably safer for the future to fix it already. In this case, just replace it by `licenses = "MIT"`. Sorry for the inconvenience.

mosteo (changes requested):

index/au/audio_base/audio_base-1.0.0.toml
gusthoff: Thanks for the hint. I've just corrected it.

mosteo (approved):

PR #360, opened 2021-09-17
Rod chapman update sparknacl 3

rod-chapman:


mosteo (approved):
Tested successfully at https://github.com/mosteo/alire-index/pull/29 with the intended compiler, so merging

PR #371, opened 2021-09-27
Add septum

pyjarrett:


mosteo:
Hi @pyjarrett , I've seen the problem and will fix it.

pyjarrett:
Oh oops.  I had just bumped the version since I was thinking that `-beta3` might be causing the issue.

mosteo:
That's indeed the issue, but I prefer to fix it for good.

mosteo (approved):

PR #374, opened 2021-09-28
Added toml for specfun

riccardo-bernardini:


index/sp/specfun/specfun-1.0.0-rc.toml
```diff
@@ -0,0 +1,16 @@
+name = "specfun"
+description = "Native Ada library of special functions (e.g., beta, gamma)"
+version = "1.0.0-rc"
```
mosteo: Hello @fintatarta . Although the submission looks good, I have reservations about using a release candidate version since that difficults specifying the dependency for your users. For example, if they wanted to ensure compatibility with your `1.0` version, including this rc, they'd have to use `>=1.0.0-rc & <2.0.0-rc`. Since `1.0.0-rc` is less that `1.0`, `^1.0` won't work in this case.

In short, I think it would be better to use a `0.x` (up to you which minor number) for a pre-stable version, which can be narrowed down easily with e.g. `~0.x` or `^0`. (Or, for users wanting to move automatically to stable, `^0|^1` is still more intuitive.

@Fabien-Chouteau , I know we haven't enforced this before. I don't know if this is something we want to make into policy (no release candidates in the index besides `0.x` versions) or we leave publishers to their own devices.

riccardo-bernardini: OK, I can submit it again with a new version number.

fintatarta: I am going to close this PR

mosteo (approved):

mosteo:
Merged, thanks!

PR #380, opened 2021-10-13
simple_components=4.42.0: Change download url to sourceforge.net (#879)

jquorning:


mosteo:
Now the problem seems to be that the github action is pulling an old docker image, not sure how this is possible... I will investigate.

mosteo (approved):

mosteo:
All right, there were actually a few issues with this crate checks, to summarize:

- The tarball could not be used directly, so I imported it to our already existing mirror for the components.
- Some compiler versions reject the now old code in `4.42`.
- The community index was overriding the changes to the already indexed version in tests with the native toolchain.
- We were using the wrong docker image for the latest CE compiler edition.

But everything is deal with, tests pass, and I'm merging this now.

PR #383, opened 2021-10-21
Add spark_unbound to alire index

mhatzl:
Hello,

this package is my take on a formally proven unbound generic alternative to  `Ada.Containers.Vector`.
But the idea is to add more data structures based on the current base functionality at some later point. 

Feedback is very much welcome.

Thanks!

Manuel


Fabien-Chouteau:
Thanks @mhatzl,

If you don't mind, I want to suggest a change in your library.
It would be better for compatibility with other projects to put all your packages under the root package named like project/crate: ` spark_unbound`:
```
 - Spark_Unbound
 - Spark_Unbound.Safe_Alloc
 - Spark_Unbound.Array
```
This way you will avoid package name clashes with other libraries and projects.

mhatzl:
@Fabien-Chouteau thanks for your feedback. I am trying to rename the packages.

Is it ok to keep the procedure `Prove_Unbound` in `prove_unbound.adb` as is since it is no package?
Otherwise, I get some problems with GNATprove not detecting SPARK_Mode at all.

Fabien-Chouteau:
> Is it ok to keep the procedure `Prove_Unbound` in `prove_unbound.adb` as is since it is no package?
> Otherwise, I get some problems with GNATprove not detecting SPARK_Mode at all.

I would move it to a "tests/" crate that you will run GNATprove on.
Like I did here: https://github.com/Fabien-Chouteau/bbqueue-spark/tree/main/tests

mhatzl:
Ok thanks I will try that, but I intentionally kept it in the main crate for easier switch between build and prove over the GPS IDE.
Moving `prove_unbound` into the tests crate removes this convenient integration, or can I combine multiple projects in one GPS window?

Sorry for those basic questions, but I am really new to all this tooling.

mhatzl:
@Fabien-Chouteau I think i found a workaround for GPS. When opening tests.gpr, Spark_Unbound/src is under tests_config.
So I will move `prove_unbound.adb` into the tests folder.

Sidenote: I will also change the license to MIT, since GPL should not affect me as noted by @onox

mhatzl:
@Fabien-Chouteau I hope i got all the changes right

Fabien-Chouteau (dismissed):

mhatzl:
could it be that i need to set a minimum gnat version in alire.toml for big_numbers to work? did not know which version is required for that

mosteo:
> could it be that i need to set a minimum gnat version in alire.toml

@mhatzl It seems so indeed. You may have to exclude certain CE editions which gets a bit cumbersome:

`(>=11 & <2000) | >=2021`

(This is just an example, I don't know either which versions already include the bignum package.)

mhatzl:
Thanks for the tipp @mosteo 
Am I right, that I have to set the version you mentioned as `gnat = "(>=11 & <2000) | >=2021)"`.
I tried to search for an exact version of GNAT where this feature was added but I only found GNAT Pro 2021, which is like the CE, but I do not understand the dependent versioning of gnat 10, 11, ...

To be sure, I would use your version scheme as this should definitely work.

mosteo:
> Am I right, that I have to set the version you mentioned as gnat = "(>=11 & <2000) | >=2021)".

Yes, that's right. 

> I do not understand the dependent versioning of gnat 10, 11, ...

@Fabien-Chouteau will know better, but I don't think there is a particular relation besides whichever version was frozen at the time for use with the CE release. What I see though for CE 2021 is:
```
$ gcc --version
gcc (GCC) 10.3.1 20210520 (for GNAT Community 2021 20210519)
```
so you might get away with including the 10 version too.

mhatzl:
> ```
> $ gcc --version
> gcc (GCC) 10.3.1 20210520 (for GNAT Community 2021 20210519)
> ```
> 
> so you might get away with including the 10 version too.

Ah ok thanks again @mosteo 
Then version 9, 10, 11, ... refers to GCC which should be `v9.3.1` if I understand the [release notes of GNATpro](https://docs.adacore.com/live/wave/gnat-release-notes/html/gnat_release_note/release_notes_21.html) correctly.

mosteo (approved):

mosteo:
Everything seems OK now.

PR #388, opened 2021-10-26
samd51_hal 0.3.0

Fabien-Chouteau:


Fabien-Chouteau:
CI only works with toolchain because the cross compiler is available.

mosteo (approved):

PR #393, opened 2021-10-31
Add Matreshka 21.0 release

reznikmm:
Replace project archives with github URL and commits. Disable GNAT CE compilers for MacOS, because they miss TLS support.

### DIFF
--- a/index/ma/matreshka_amf/matreshka_amf-21.0.0.toml
+++ b/index/ma/matreshka_amf/matreshka_amf-21.0.0.toml
@@ -1,0 +1,22 @@
+description = "Implementation of OMG's Meta Object Facility (MOF)"
+name = "matreshka_amf"
+version = "21.0.0"
+website = "https://forge.ada-ru.org/matreshka"
+authors = ["Vadim Godunko"]
+licenses = "BSD-3-Clause"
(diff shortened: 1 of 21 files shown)


index/ma/matreshka_amf/matreshka_amf-18.1.0.toml
```diff
@@ -12,6 +12,7 @@ tags = ["uml", "meta", "mof"]
 [[depends-on]]
 matreshka_league = "18.1.0"
 matreshka_xml = "18.1.0"
+gnat = "<10|(>=2000 & <2020)"
```
mosteo: If these are the old compilers the PR alludes to, then this condition is the complementary of what is needed.

reznikmm: new compilers is unable to build old matreshka, because they have some "improvements". So matreshka 18 should be built with an old compiler. I'll try to split changes to separate editing of older crates from adding new ones.

mosteo: Ok, I see.

mosteo (approved):

PR #396, opened 2021-11-03
Add Libadalang and Langkit_Support 22.0.0

pmderodat:


pmderodat:
The checks seem to use an old gprbuild: `/usr/sbin/ld: -r and -pie may not be used together` is probably the issue fixed in https://github.com/AdaCore/gprconfig_kb/commit/4863a2068e3cdacfe8b225386772ec3508e4f0b7 But I can’t check for sure, as I’m not able to see from the logs which version is used exactly. What should I do?

mosteo:
If you're not in a hurry we can patch the checker to display the gprbuild version in use to be certain about it. 

Besides that, one possibility is to disable availability on the problematic platforms, or probably better to give explicit dependency on gprbuild with the versions that don't have the problem.

pmderodat:
It looks like even if the gprbuild version is displayed in the CI, we’ll have to restrict the gprbuild version to be correct…

Now I wonder why we have this problem now, while it worked fine for Langkit_Support 21.0 and nothing is supposed to have changed for this matter since then.

mosteo:
Maybe the failing platform (arch) wasn't yet in our CI tests?

Fabien-Chouteau:
@mosteo the problem here is between gprbuild and the Arch Linux GNAT (if I understand correctly).
It won't be fixed soon so let's merge this one.

mosteo:
All right, I'm merging but don't hesitate to amend it with an `available` property if appropriate.

pmderodat:
> @mosteo the problem here is between gprbuild and the Arch Linux GNAT (if I understand correctly). It won't be fixed soon so let's merge this one.

But which gprbuild is involved here? (there is no standard gprbuild on Arch Linux AFAIK) Recent versions should not have this problem.

mosteo:
It's being bootstrapped from here: https://aur.archlinux.org/packages/gprbuild-bootstrap/

pmderodat:
I see, thanks. I double checked: the fix landed in gprbuild_kb at the end of October 2020, so it’s out of gprbuild 21.0, which is what the AUR packages provide. So things will get fixed once they migrate to gprbuild 22.0.

PR #400, opened 2021-11-06
Add Markdown Highlighter for GNAT Studio v0.1.0

egilhh:
This is a simple syntax highlighter of github-style Markdown files for GNAT Studio

### DIFF
--- a/index/ma/markdown_highlighter/markdown_highlighter-0.1.0.toml
+++ b/index/ma/markdown_highlighter/markdown_highlighter-0.1.0.toml
@@ -1,0 +1,26 @@
+name = "markdown_highlighter"
+description = "Simple Markdown syntax highlighter for GNAT Studio"
+version = "0.1.0"
+
+authors = ["Egil H. Høvik"]
+maintainers = ["Egil H. Høvik <egilhh+dev@gmail.com>"]
+maintainers-logins = ["egilhh"]
+
+licenses = "MIT"
+
+website = "https://github.com/egilhh/markdown_highlighter"
+
+tags = ["markdown", "syntax", "highlighter", "gnat"]
+
+auto-gpr-with=false
+
+[environment]
+GNATSTUDIO_CUSTOM_PATH.append = "${CRATE_ROOT}/src"
+
+
+
+
+[origin]
+commit = "46c6ea33aff9b9c5278035e305eb720d8553b8a4"
+url = "git+https://github.com/egilhh/markdown_highlighter.git"
+


Fabien-Chouteau:
Hello @egilhh, looks like this is a Python project. I am not sure it belongs to the Alire ecosystem.

egilhh:
> Hello @egilhh, looks like this is a Python project. I am not sure it belongs to the Alire ecosystem.

Well, it provides a plug-in for GNAT Studio, and a .gpr-file that can be extended to tell GNAT Studio/gprbuild how to handle Markdown files, for example:

    project Docs extends "markdown_highlighter.gpr" is
       for Source_Dirs use (".");
    end Docs;

Are tools or plug-ins not allowed?

Fabien-Chouteau:
I am not fond of potentially having a dependency of my dependencies that changes the behavior of my editor without me knowing it. I had a look at your code and it seems fine, but we can imagine nasty things being done this way.

In my opinion this should be contributed to GNATstudio: https://github.com/AdaCore/gps 

In the meantime, if people want to use/try your plug-in they can:
```console
$ alr with markdown_highlighter --use=https://github.com/egilhh/markdown_highlighter
```
@mosteo what do you think?

mosteo:
Is the mere deployment of the dependency affecting GNATstudio behavior? If so that would be certainly problematic. If the user has to take extra steps to activate the plugin I think it could be OK. I'm afraid I have no knowledge of GNATstudio plugin system.

> In my opinion this should be contributed to GNATstudio: https://github.com/AdaCore/gps

This is also probably better to ensure maximum diffusion as it would benefit all future users of GNATstudio.

egilhh:
> I am not fond of potentially having a dependency of my dependencies that changes the behavior of my editor without me knowing it. I had a look at your code and it seems fine, but we can imagine nasty things being done this way.

Hmm, OK. It's not a permanent change, though, it just sets the plug-in search path with GNATSTUDIO_CUSTOM_PATH.append = "${CRATE_ROOT}/src". But I do see (and agree with) your point about inheriting this through dependencies.

> In my opinion this should be contributed to GNATstudio: https://github.com/AdaCore/gps

Maybe, in time 

> In the meantime, if people want to use/try your plug-in they can:
> 
> ```
> $ alr with markdown_highlighter --use=https://github.com/egilhh/markdown_highlighter

Yes, that seems like a good compromise, and it is what I used for testing. Maybe this is also a case for a sub-crate

PR #409, opened 2021-11-29
Add libadalang_tools crate

reznikmm:


mosteo:
Seems to be a problem with the style checks on Windows

reznikmm:
I can fix it on `master`, but I'm not sure how to fix it in 22 release.

pmderodat:
There was no such problem with Libadalang and Langkit_Support, whereas both are built with `-gnatyg` as well. Does someone understand why we have it with Libadalang-tools?

I’m not in favor of having releases use sources from forks, so I would like to make sure that there is no other possibility before doing this.

reznikmm:
@pmderodat 
I found that LAL and langkit are built from source archives. While I'm trying to build LAL_Tools from a git commit. Perhaps this is the reason of failure. 

Could you make a 22.0 release of LAL_Tools?

Forks are just for the issue investigation. I've mark PR as a draft for now.

.github/workflows/build-native.yml
```diff
@@ -21,6 +21,10 @@ jobs:
           - windows-latest
 
     steps:
+    - name: Initialize git autocrlf
+      shell: bash
+      run: |
+        git config --global core.autocrlf input
```
Fabien-Chouteau: I don't agree with this because it will hide from us issue that users will have on their setup.

Fabien-Chouteau (changes requested):

index/li/libadalang/libadalang-22.0.1.toml
```diff
@@ -0,0 +1,29 @@
...
+maintainers = ["Pierre-Marie de Rodat <pmderodat@kawie.fr>"]
+maintainers-logins = ["pmderodat"]
+licenses = "GPL-3.0-only WITH GCC-exception-3.1"
+website = "https://github.com/AdaCore/libadalang"
+tags = ["libadalang", "static-analysis"]
+project-files = ["build/libadalang.gpr"]
+
+[origin]
+url = "git+https://github.com/AdaCore/libadalang"
+commit = "112f4aec5ff6664d29c493c1f0fbb09facb12ae1"
```
Fabien-Chouteau: Looks like this commit doesn't correspond to any official release.

Fabien-Chouteau:
> Could you make a 22.0 release of LAL_Tools?

Yes, let's do that.

.github/workflows/build-native.yml
reznikmm: For now this is the only way (that I know) to make GitHub CI build LAL from github. We use this step in LAL and ALS builds on Windows. And now we know that it works for Alire too.

Fabien-Chouteau: The `-gnatyd Check no DOS line terminators present.` style check should be disabled. Especially for release build.

reznikmm:
The style issue is fixed by replacing git source origin with archive. This looks strange from my point of view.

pmderodat:
> @pmderodat I found that LAL and langkit are built from source archives. While I'm trying to build LAL_Tools from a git commit. Perhaps this is the reason of failure.

Ah you are right, and that rings a bell, now.
> 
> Could you make a 22.0 release of LAL_Tools?

Sure: I just did a pre-release <https://github.com/AdaCore/libadalang-tools/releases/tag/v22.0.0>. I’ll mark it as a regular release when the crate works if that’s ok for you.

.github/workflows/build-native.yml
pmderodat: Since there are too many projects out there which enable these checks, I don’t think it’s reasonable to expect all of them to be updated to disable this one, especially in this context: once it’s released, it’s too late to modify the project, and unfortunately, we have no official way to patch releases in Alire, which would allow some flexibility here.

Alternatively, I suppose we could switch defaults to use the “production” mode, where warnings are not supposed to be treated as errors. But here too, most projects use their “dev” mode by default out of convenience, so I’m not sure how that would play in the Alire ecosystem.

reznikmm:
@pmderodat  With this archive the crate is green!

But `alr run gnatpp` failes because `alr` auto-detects `gnatpp` twice. There are extra commands in `valgrind/` directory. I tried to delete it, but I'm unable to find a correct syntax for TOML:

```
[[actions]]

[actions.'case(os)'.windows]
type = "post-fetch"
command = ["cmd.exe", "/c", "del /Q valgrind"]

[actions.'case(os)'.'...']
type = "post-fetch"
command = ["rm", "-rf", "valgrind"]
```

pmderodat:
Thanks! The syntax looks fine, what’s the issue you get?

reznikmm:
```
ERROR: Loading release from /home/runner/work/alire-index/alire-index/index/li/libadalang_tools/libadalang_tools-22.0.0.toml:
ERROR:    actions:
ERROR:    actions:
ERROR:    action #1:
ERROR:    action type missing
ERROR at line 69 (code: 1)
```

And before this I've tried:
```
[[actions]]
type = "post-fetch"

[actions.'case(os)'.windows]
command = ["cmd.exe", "/c", "del /Q valgrind"]

[actions.'case(os)'.'...']
command = ["rm", "-rf", "valgrind"]
```

And error was:
```
ERROR: Loading release from /home/runner/work/alire-index/alire-index/index/li/libadalang_tools/libadalang_tools-22.0.0.toml:
ERROR:    actions:
ERROR:    actions:
ERROR:    action #1:
ERROR:    action command missing
, stderr ERROR at line 69 (code: 1)
```

pmderodat:
Thanks a lot for the details. At the very least, I think `alr` should have better error reporting for this… if there is indeed an error in the spec, instead of a bug in `alr`. @mosteo @Fabien-Chouteau what do you think?

reznikmm:
Let's push it as is, without removing `valgrind` directory, to unblock work for adding Ada Language Server 22.0.0.

mosteo:
@reznikmm, I believe the proper syntax is:
```
[[actions.'case(os)'.windows]]
type = "post-fetch"
command = ["cmd.exe", "/c", "del /Q valgrind"]

[[actions.'case(os)'.'...']]
type = "post-fetch"
command = ["rm", "-rf", "valgrind"]
```

Otherwise you have an empty action (in your first try) or a partially defined action (in the second one).

So the alire errors were actually accurate, fingering the first action, even if a bit obscure (but here I think TOML syntax is partially to blame).

Let me know if you want to make a last attempt or should I merge as-is. And sorry for the delayed feedback.

reznikmm:
@mosteo It works for me!

Fabien-Chouteau (approved):

PR #411, opened 2021-12-11
aflex 1.5.2021 - An Ada Lexical Analyzer Generator

stcarrez:


index/af/aflex/aflex-1.5.2021.toml
```diff
@@ -0,0 +1,42 @@
...
+
+- Aflex generates the spec and body files as separate files so that
+  there is no need to use gnatchop to split the DFA and IO files.
+- Aflex uses the lex file name to generate the package name and
+  it supports child package with the `%unit` directive.
+
+"""
+
+[gpr-externals]
+BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
```
Fabien-Chouteau: Hi @stcarrez, I will not block the PR for that but using such a common scenario variable as `BUILD` will lead to incompatibilities. If other projects use the same scenario variable name there is a high risk of incompatible values for the variable, for instance using `release` instead of `distrib`.

It is highly recommended to prefix your scenario variables with the project name, e.g. `AFLEX_BUILD`.

index/af/aflex/aflex-1.5.2021.toml
```diff
@@ -0,0 +1,42 @@
...
+
+[gpr-set-externals]
+BUILD = "distrib"
```
Fabien-Chouteau: It is also not recommended to fix the build mode here as project depending on this crate may want to change it to debug or optimize mode for instance. It's better to have the default mode set in the gpr file.

Fabien-Chouteau (dismissed):

index/af/aflex/aflex-1.5.2021.toml
```diff
@@ -0,0 +1,42 @@
...
+
+
+[origin]
+commit = "00967b9299f7e56ed10c0db6b893bb8948cdd943"
+url = "git+https://github.com/Ada-France/aflex.git"
```
mosteo, line 41: Another issue with these two submissions is that each includes a `config.gpr` project, so they can't possibly be built together in our current build methodology. The simplest solution would be to rename those to `config_aflex.gpr` and likewise, and disabling our own configuration generation:
```
[configuration]
disable = true
```

Another possibility (not tested) would be to include as a post-fetch action the building of the crate as a library, and providing an externally built project file in the manifest, instead of the current project file which I guess is used to generate the library.

But, as-is, I just tested that these two `aflex` and `ayacc` cannot be used together within Alire.

mosteo (changes requested):

Fabien-Chouteau (approved):

mosteo:
Hi @stcarrez, thanks for uploading the fixes. I found another issue in that both `aflex` and `ayacc` include a `parser.ads`/`parser.adb` package. Hence, compilation fails because GNAT complaints that a unit cannot belong to both projects. I guess renaming one would work, but I'm unsure if that can have other consequences in the public interface.

I don't know another solution for clashing top-level package names. What I find puzzling is that even using these both as pre-compiled libraries, the clash should happen anyway for the `.ads`/`.ali` files? Or am I misinterpreting something (like ayacc not really being intended to be used together with aflex?)

stcarrez:
aflex and ayacc are not libraries but final program. They were written in Ada 83 around the 90s.
I'm not going to rename the `Parse` package in any of these tools just to please Alire. Now, may be Alire is not appropriate to distribute a final program.

mosteo:
No, then is not a case of pleasing Alire, but precisely that I missed that these where final programs, so they cannot conflict with each other. I apologize for the noise.

PR #416, opened 2021-12-16
Gdalada, generated Ada bindings to GDAL

nbrk:


mosteo:
Hello, @nbrk, thanks for the submission. Given that the binding does not compile as-is in some platforms, you need to add explicit dependencies on the `gnat` versions that do compile it. See, for example: https://github.com/alire-project/alire-index/pull/412/files

mosteo:
This PR should be reopened against the `stable-1.2` branch, which is the current stable. In my experience it is not enough to rebase and change the base branch, due to some limitation in the Github actions framework, so I'm closing it now. Sorry for the inconvenience.

PR #418, opened 2021-12-28
Release 12 27 2022

pmunts:
Added support for locked antiphase motor drivers, which are controlled
by a single PWM output as follows:
    
100% duty cycle => Full speed clockwise.
50%  duty cycle => Stopped.
0%   duty cycle => Full speed counterclockwise.


pmunts:
Date was incorrect.  Will be superceded by another pull requiest.

PR #428, opened 2022-01-12
Add xoshiro 1.0.0

onox:


Fabien-Chouteau (approved):
Nice one @onox I was actually considering making this kind or pseudo RNG crate since I sometimes copy/paste the same crappy RNG code in my projects :+1: 

Did you consider making the package generic to instantiate with different unsigned types?
```ada
generic
   type T is mod <>;
package Xoshiro_Gen with SPARK_Mode => On is
```

onox:
> Did you consider making the package generic to instantiate with different unsigned types?

A generic package would be great, but how exactly should it be implemented? Depending on the bit size of the type (32 or 64 bits) you need either the 128++ or 256++ version. They have 3 slightly different constants in procedure `Next`.

Fabien-Chouteau:
> They have 3 slightly different constants in procedure `Next`.

I didn't see that...

onox:
Basically have different instantiations at compile time, something like:

```ada
generic
   type T is mod <>;
package Xoshiro is
  (case T'Object_Size is
     when 32 => new Xoshiro_Gen (T, 7, 9, 11),
     when 64 => new Xoshiro_Gen (T, 23, 17, 45));
```

But I don't think Ada can do this at the moment :slightly_smiling_face:

Fabien-Chouteau:
@onox Let's discuss in the repo :)

Fabien-Chouteau:
@onox Let's discuss in the repo :)

PR #430, opened 2022-01-15
brackelib 0.0.1 (was: Initial release of my first crate :-))

bracke:


index/br/brackelib/brackelib-0.0.1.toml
```diff
@@ -0,0 +1,17 @@
...
+long-description = "Ada library which contains various utility packages. Currently only a few implementations of abstract data types."
+version = "0.0.1"
+licenses = "CC0-1.0"
+website = "https://github.com/bracke/brackelib"
+tags = ["utility", "adt","stack", "queue"]
+authors = ["Bent Bracke"]
+maintainers = ["Bent Bracke <bent@bracke.dk>"]
+maintainers-logins = ["bracke"]
+[[depends-on]]  # Added by alr
+aunit = "^21.0.0"  # Added by alr
```
mosteo: Is `aunit` really needed by the library itself?

The proper way to do this if this is used only for testing, is to have a new crate in a folder, which indeed depends on "aunit", and also on the crate being tested via a pin. This crate remains "hidden" from your library dependents but you can use it locally for tests. See https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing

bracke: I have removed the aunit dependency and pushed a new version 0.0.2

mosteo: Looks good, @brake. However, since 0.0.1 hasn't made into the index and I don't think it's interesting anymore to publish it, I'd prefer if you remove its manifest from the PR so only 0.0.2 goes in. Thanks in advance.

bracke:
Done

mosteo (approved):

PR #434, opened 2022-01-22
Advanced Resource Embedder 1.2.0

stcarrez:
New version of Advanced Resource Embedder.

### DIFF
--- a/index/ar/are/are-1.2.0.toml
+++ b/index/ar/are/are-1.2.0.toml
@@ -46,7 +46,7 @@ end Config;
 [available.'case(os)']
 linux = true
 windows = true
-macos = false
+macos = true
 '...' = false
 
 [[depends-on]]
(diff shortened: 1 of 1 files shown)


index/ar/are/are-1.2.0.toml
```diff
@@ -0,0 +1,85 @@
...
+    return access constant String;
+end Config;
+```
+
+"""
+
+[available.'case(os)']
+linux = true
+windows = true
+macos = false
```
Fabien-Chouteau: Do you know why it is not working on macOS?

Fabien-Chouteau (dismissed):

index/ar/are/are-1.2.0.toml
stcarrez: No. It was failing when I submitted the 1.1.0 version.
It compiles correctly on MacOS for me with GNAT CE 2021.
I'll update the alire file and we will see.

mosteo (approved):

mosteo:
Looks good also on macOS; thanks, Stephane.

PR #436, opened 2022-01-26
EDC Client 1.3.2

hgrodriguez:
2nd attempt with correct line endings by adding .gitattribute

### DIFF
--- a/index/ed/edc_client/edc_client-1.3.2.toml
+++ b/index/ed/edc_client/edc_client-1.3.2.toml
@@ -1,0 +1,18 @@
+name = "edc_client"
+description = "Client library for: github.com/hgrodriguez/embedded-dashboard-console"
+version = "1.3.2"
+licenses = "BSD-3-Clause"
+
+authors = ["Holger Rodriguez"]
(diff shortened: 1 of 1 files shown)


mosteo:
Hi @hgrodriguez . I've been checking what's going on and your `.gitattributes` looks correct to me. I've narrowed down the problem to the clone being made by github natively (with crlf, even before you added `.gitattributes`) but gprbuild is still expecting lf, I'm not sure why.

Besides I realized your project file already contains the proper switches. The problem is with the `gnat.adc`, which is re-enabling the check again. To disable the check there, you must change line 5 to `pragma Style_Checks ("g-d");`. The `-d` cannot be added anywhere else, as it seems gprbuild is checking every line in `gnat.adc` with the switches up to the line being parsed, so line 5 fails otherwise.

Sorry for the confusions and let's see if next time is the charm. Just push the change here, no need to create another PR.

hgrodriguez:
@mosteo thanks a lot for your investigation. I am sorry, that this drags out so long. As I screwed up the last time re-using a branch, i decided to close this one and i will create a fresh one to assure, that nothing is lurking around.

PR #445, opened 2022-02-05
Driver for SH1107/OLED 128x128 monochrome display

hgrodriguez:
This is an implementation for:
Driver for SH1107/OLED 128x128 monochrome display

### DIFF
--- a/index/sh/sh1107/sh1107-1.2.2.toml
+++ b/index/sh/sh1107/sh1107-1.2.2.toml
@@ -1,0 +1,18 @@
+name = "sh1107"
+description = "Driver for the SH1107/OLED 128x128 monochrome display"
+version = "1.2.2"
+licenses = "BSD-3-Clause"
+
+authors = ["Holger Rodriguez"]
+maintainers = ["Holger Rodriguez <github@roseng.ch>"]
+maintainers-logins = ["hgrodriguez"]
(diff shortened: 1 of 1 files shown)


hgrodriguez:
it seems to fail, but this time at least not with line ending problems. 
What do I need to fix?

mosteo:
My first guess is that your configuration treats warnings as errors. I don't remember how to disable that off the top of my head, but you may try to look in that direction. Or you may try to fix the warnings, if it's practical.

hgrodriguez:
thanks for this. nevertheless, it builds on some platforms and not others. And yes, I do treat warnings as errors, I can promise you, that on my machine [macoS] there are no warnings which prevents the build.
I will check as much as I can.

mosteo:
It's normal to get different results in our tests because we use an assortment of compilers; when a Linux distro packages a GNAT we use that one to see which experience users of the platform will get, besides we check with both the latest Community Edition and the latest Alire-supplied toolchain.

Probably it's safer from a forward-looking PoV to not enable "warnings as errors" for packaging, as newer GNATs tend to uncover more issues.

hgrodriguez:
I am closing this pull request, as it needs modifications for the dependencies (embedded components -> adl_middleware).

PR #446, opened 2022-02-12
Arduino ItsyBitsy BSP for RP2040

hgrodriguez:


mosteo:
The problem with the "Check PR" test is in the javascript TOML parser that is outdated, so nothing to worry about.

The build failure in `ubuntu-latest::ubuntu-lts` is a known bug in our testing framework for cross-target crates (https://github.com/alire-project/alire/issues/900) that I should fix ASAP. 

As the crate passes the build in the three platforms with our bundled compilers, all is good.

mosteo:
Actually, there is a minor issue with the pin. See the review for that.

index/it/itsybitsy_bsp/itsybitsy_bsp-0.0.2.toml
```diff
@@ -0,0 +1,26 @@
...
+[[depends-on]]  # Added by alr
+hal = "~0.1"  # Added by alr
+[[depends-on]]  # Added by alr
+rp2040_hal = "^1"  # Added by alr
+
+[configuration.values]
+rp2040_hal.Flash_Chip = "w25qxx"
+
+[[pins]]  # Added by alr
+rp2040_hal = { url='https://github.com/JeremyGrosser/rp2040_hal' }   # Added by alr
```
mosteo, line 21: Indexed crates should not contain pins, as that overrides the dependency solver. As you already have the dependency on `rp2040_hal` in the `[[depends-on]]` array, you need just to remove this pin for the publishing.

mosteo (changes requested):

hgrodriguez:
Closing this pull request to avoid any crossed wires, will create a new one.

PR #455, opened 2022-02-24
lvgl_ada 0.2.0

Fabien-Chouteau:


mosteo:
Given the error [here](https://github.com/alire-project/alire-index/runs/5318744076?check_suite_focus=true#step:8:470) it seems the subrepo has to be switched to `https` protocol.

Fabien-Chouteau:
Good catch!

index/lv/lvgl_ada/lvgl_ada-0.2.0.toml
```diff
@@ -0,0 +1,27 @@
...
+maintainers = ["fabien.chouteau@gmail.com"]
+maintainers-logins = ["Fabien-Chouteau"]
+website = "https://github.com/Fabien-Chouteau/lvgl-ada"
+tags = ["embedded", "gui", "lvgl", "no-std"]
+
+[configuration.variables]
+Pixel_Bit_Depth             = {type = "Enum", values = ["Pix_1bit", "Pix_8bit", "Pix_16bit", "Pix_32bit"]}
+Horizontal_Resolution       = {type = "Integer", first = 1, last = 1000000}
+Vertical_Resolution         = {type = "Integer", first = 1, last = 1000000}
+Density_Per_Inch            = {type = "Integer", first = 1, last = 1000000}
```
mosteo, line 14: It seems the lack of default in some of those is causing trouble. I guess this is a shortcoming in our current testing, as it may be desirable to force clients to pick a value.

Fabien-Chouteau: Indeed, it doesn't make sense to set default values for many of the settings here.

mosteo: So I guess we can merge and think separately how to better test this kind of situation.

mosteo (approved):

PR #469, opened 2022-03-03
Install indexed gprbuild when necessary

### LINKED ISSUE #900: Install `gprbuild` too when a cross-compiler is automatically selected
Otherwise the native `gprbuild` is unable to locate/use the compiler.

See e.g. https://github.com/alire-project/alire-index/pull/417

mosteo:
Fixes https://github.com/alire-project/alire/issues/900

Tested in https://github.com/mosteo/alire-index/pull/44 and https://github.com/mosteo/alire-index/pull/45

### DIFF
--- a/scripts/gh-build-crate.sh
+++ b/scripts/gh-build-crate.sh
@@ -112,6 +112,12 @@ for file in $CHANGES; do
       echo No need to update system repositories
    fi
 
+   # Install an Alire-provided gprbuild whenever there is a non-external gnat in solution
+   if grep -iq 'gnat_' <<< $solution && ! grep -iq 'gnat_external' <<< $solution; then
+      echo "INSTALLING indexed gprbuild"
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau (approved):

Fabien-Chouteau:
I feel like this is showing a problem that we have with gprbuild.

Maybe we should pick a version of gprbuild, package it with Alire and that what will be used all the time.

mosteo:
The problem with that is that a gprbuild installed by Alire doesn't detect a system GNAT. Maybe we should automate from alr itself what this PR is doing.

Fabien-Chouteau:
> gprbuild installed by Alire doesn't detect a system GNAT

That might only be true for Debian/Ubuntu, but I am not sure. With a little bit of research we may be able to make it work for both cases/

mosteo:
That would simplify things, yep.

PR #470, opened 2022-03-06
Add fastpbkdf2_ada 2.0.0

AntonMeep:


index/fa/fastpbkdf2_ada/fastpbkdf2_ada-2.0.0.toml
```diff
@@ -0,0 +1,18 @@
...
+website = "https://github.com/AntonMeep/fastpbkdf2_ada"
+licenses = "ISC"
+tags = ["pbkdf2", "cryptography"]
+
+authors = ["Anton Fediushin"]
+maintainers = ["Anton Fediushin <imantonmeep@gmail.com>"]
+maintainers-logins = ["AntonMeep"]
+
+[[depends-on]]
+openssl = "~1.1.1"
```
mosteo, line 13: Shouldn't this one be `^1.1.1` ?

AntonMeep: This virtually doesn't make a difference, as OpenSSL LTS branch (1.1.1) does not follow semver specification and patch versions are indicated by a letter, say 1.1.1m. Even more than that, next major version is labeled 3.0.0, and now it will follow semver.
All in all, I am unsure what to put here, and thought that tilde will be good enough 🤷

mosteo: If it's not following semver until `3.0` I'd say the tilde is the safest choice here. So all is well.

mosteo (approved):

PR #475, opened 2022-03-09
new package: epoll

JeremyGrosser:


index/ep/epoll/epoll-0.1.0.toml
```diff
@@ -0,0 +1,15 @@
+name = "epoll"
+description = "Linux epoll binding"
```
mosteo: If this is linux-specific, better add:
```
[available.'case(os)']
linux = true
'...' = false
```

mosteo (changes requested):

index/ep/epoll/epoll-0.1.0.toml
JeremyGrosser: That's a very good point. Fixed.

mosteo:
Thanks for the fix, Jeremy. The remaining errors are not your fault, so merging.

mosteo (approved):

PR #480, opened 2022-03-22
Add sha1 v1.0.0

AntonMeep:


mosteo (approved):

mosteo:
Have you seen `GNAT.SHA1` (and `GNAT.Secure_Hashes.*` in general)?

AntonMeep:
> Have you seen `GNAT.SHA1` (and `GNAT.Secure_Hashes.*` in general)?

Yessir, I wouldn't bother rewriting basic algorithms if provided implementation was adequate, which GNAT.SHA* unfortunately are not - they're slow (not to say that my crate is fast, but I can work on that gradually), have weird API which is not consistent amongst all platforms (that's why #479 took me a while to get right) and worst of all are not Pure, Preelaborate which is nothing short of a crime in 2022.
There are a few other implementations I am aware of, but these usually are a part of bigger libraries which is not quite alire-friendly 🤷 Small crates is what we need!

mosteo:
Great! It was just in case, I have more than once failed to realize that something very useful is in `GNAT.*`

PR #482, opened 2022-03-24
Add sha2 v1.0.0

AntonMeep:


AntonMeep:
Nevermind, something has to be changed

PR #488, opened 2022-03-31
Add GNATprove 11.2.3 and gnat 11.2.4

Fabien-Chouteau:


mosteo (approved):

mosteo:
A new cross-compiler and gnatprove, now this is a PR!

Right now, to use gnatprove one has to add it as a dependency, which is a bit counterintuitive and probably not something we want in top-level/submitted crates. This can be work around by using a nested testing crate I guess.

However, it could be time to think about the possibility of having other crates installed to the shared config with either `alr toolchain` or some variant of `alr install`. The infrastructure is already there.

Another possibility would be to mark somehow crates as toolchain components (new manifest field?) or consider any binary crate a potential toolchain element.

Right now gnat and gprbuild are hardcoded as toolchain elements because they're the minimum required. I don't like the idea of updating alr itself to add other binaries as installable tools.

Thoughts?

Fabien-Chouteau:
That's a tricky issue indeed. For the moment I would advise to add `gnatprove` as a dependency of a test crate, like `aunit` or `gnatcov`.

PR #499, opened 2022-04-15
Update apdf-5.0.0.toml

zertovitch:
Upgrade to release #3 of Ada PDF Writer 005.
Using an Alire-friendly project file (with-ing the GID project), fixed for recent GNAT's incremental compilation.
Build scenario has "PDF_" prefixed in order not to mix up with other projects.

### DIFF
--- a/index/ap/apdf/apdf-5.0.0.toml
+++ b/index/ap/apdf/apdf-5.0.0.toml
@@ -1,15 +1,18 @@
-description = "Standalone, portable package for producing dynamically PDF documents"
+description = "Portable package for producing dynamically PDF documents"
 name = "apdf"
 version = "5.0.0"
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Hello @zertovitch,

Changing the sources of an existing release is not allowed. There is no way for us to check what kind of changes you introduced. See https://alire.ada.dev/docs/#release-immutability

You have to make a new release with those changes. Maybe `5.3.0`?

zertovitch:
No problem.
Rather maybe call it `5.0.3`. The changes are tiny: only on .gpr, comments, doc, etc.

zertovitch:
I've made a PR for a `5.0.3`, so I close this one.

PR #501, opened 2022-04-19
Create apdf-5.0.3.toml

zertovitch:
New crate descriptor points to 3rd release of the #005 version of APDF.
Using an Alire-friendly GNAT project file (with-ing the GID project), fixed for recent GNAT's incremental compilation.
Build scenario has "PDF_" prefixed in order not to mix up with other projects.

### DIFF
--- a/index/ap/apdf/apdf-5.0.3.toml
+++ b/index/ap/apdf/apdf-5.0.3.toml
@@ -1,0 +1,18 @@
+description = "Portable package for producing dynamically PDF documents"
+name = "apdf"
+version = "5.0.3"
+authors = ["Gautier de Montmollin"]
+licenses = "MIT"
+maintainers = ["fabien.chouteau@gmail.com"]
+maintainers-logins = ["zertovitch", "Fabien-Chouteau"]
+project-files = ["pdf_out_gnat_w_gid.gpr"]
+
+[gpr-externals]
+PDF_Build_Mode = ["Debug", "Fast"]
+
+[[depends-on]]
+gid = ">=9.0.0"
+
+[origin]
+url = "https://sourceforge.net/projects/apdf/files/apdf_005_r3.zip"
+hashes = ["sha512:dbe27598986b1744b024803348350e48b9fe14a14b4137796b3bf12fc98e400b45fd16dc3902a5ffbfa407552131bec072c287444889d5984ade6ba6d2d981cf"]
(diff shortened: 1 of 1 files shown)


index/ap/apdf/apdf-5.0.3.toml
```diff
@@ -0,0 +1,18 @@
+description = "Portable package for producing dynamically PDF documents"
+name = "apdf"
+version = "5.0.3"
+authors = ["Gautier de Montmollin"]
+licenses = "MIT"
+maintainers = ["fabien.chouteau@gmail.com"]
```
Fabien-Chouteau: You should put your email here instead.

Fabien-Chouteau (dismissed):

index/ap/apdf/apdf-5.0.3.toml
zertovitch: I guess I can do it online too, but how?

zertovitch: Found an ´Edit´ button... Let's see...

index/ap/apdf/apdf-5.0.3.toml
```diff
@@ -0,0 +1,18 @@
...
+licenses = "MIT"
+maintainers = ["gdemont@hotmail.com"]
+maintainers-logins = ["zertovitch", "Fabien-Chouteau"]
+project-files = ["pdf_out_gnat_w_gid.gpr"]
+
+[gpr-externals]
+PDF_Build_Mode = ["Debug", "Fast"]
+
+[[depends-on]]
+gid = ">=9.0.0"
```
mosteo, line 14: @zertovitch, should this be `"^9.0.0"` instead, or isn't `gid` following semantic versioning conventions?

zertovitch: With `">=9.0.0"` I mean any version equal to or above version 9.0.0. Could be the 10.0.0 for instance.
In the [doc](https://alire.ada.dev/docs/#release-information), I read:

> Available constraint operators are the usual Ada relationals (=, /=, >, >=, <, <=) plus caret (^, any upwards version within the same major point) and tilde (~, any upwards version within the same minor point).

Did I misunderstand it?

mosteo: Yes, that's correct, but the issue is not with the use of operators. The thing is that usually there is no guarantee that version 10 is back-compatible with version 9, so the caret operator is the safe upgrade route according to semver. That's what I mean with `gid` following or not semver conventions (major version changes are compatibility-breaking in semver).

zertovitch: Ah OK. In that sense `gid` does not follow the "semver" convention. A bit like the Windows API: compatibility is sacred!

mosteo: OK then!

mosteo (approved):

PR #506, opened 2022-04-26
Fix Licenses in ini_files

mosteo:


Fabien-Chouteau:
@mosteo  Should we reject old licenses in the next release?

mosteo:
> @mosteo Should we reject old licenses in the next release?

It's been a lot of time, so we might as well. But perhaps this is a change that we should have in the master branch for a longer while, just to see what happens. I hope to find the time to release the next version early next month, and that might be not enough exposure.

For the time being I'm preparing a better check for our index GH actions.

PR #509, opened 2022-04-27
json 5.0.2

onox:
Not sure if `stable-1.1` is ok since the toml file uses the `subdir` property.


mosteo:
> Not sure if `stable-1.1` is ok since the toml file uses the `subdir` property.

Indeed it isn't. We'll need to bump the index version and create a new branch, so it will have to wait a little.

mosteo:
Hi @onox, could you rebase on top of `devel-1.2`?

onox:
Replaced with PR #510

PR #512, opened 2022-05-02
rp2040_hal, pico_bsp, pico_examples 1.5.0

JeremyGrosser:
[1.5.0 Release notes](https://pico-doc.synack.me/#1_5_0)

The only significant change around packaging is that pico_examples now depends on chests.

### DIFF
--- a/index/pi/pico_bsp/pico_bsp-1.5.0.toml
+++ b/index/pi/pico_bsp/pico_bsp-1.5.0.toml
@@ -1,0 +1,22 @@
+name = "pico_bsp"
+description = "Board support package for Raspberry Pi Pico"
+version = "1.5.0"
+licenses = "BSD-3-Clause"
+
+authors = ["Jeremy Grosser"]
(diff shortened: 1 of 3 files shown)


mosteo (approved):

mosteo:
I saw Fabien's comment on gitter about this one maybe having to be `2.0`, waiting for confirmation on that.

JeremyGrosser:
As discussed on gitter, I think keeping the major version is fine... This release removes `RP.SysTick`, which has been broken for the last few releases and if anyone were using it, they'd have complained by now. So yes, the code is backwards incompatible, but it's a bit like a [tree falling in a forest](https://en.wikipedia.org/wiki/If_a_tree_falls_in_a_forest).

imo, we can release this as 1.5.0. But if you feel strongly about it, I can re-tag this as 2.0.0.

Fabien-Chouteau:
It's up to you @JeremyGrosser, so let's go.

PR #513, opened 2022-05-03
Update diff-release.yml

Fabien-Chouteau:


.github/workflows/diff-release.yml
```diff
@@ -23,17 +23,24 @@ jobs:
...
+      if: (matrix.os == 'ubuntu-latest') && (contains(github.base_ref, 'devel-'))
+      uses: ada-actions/toolchain@ce2020
+      with:
+        distrib: fsf # faster install?
+
+    - name: Set up GNAT toolchain (Community)
+      if: (matrix.os != 'ubuntu-latest') && (contains(github.base_ref, 'devel-'))
+      uses: ada-actions/toolchain@ce2020
+      with:
+        distrib: community
```
mosteo, line 42: I thought this wasn't necessary? I guess it is in some case?

Fabien-Chouteau: We are building the master branch below in case of `devel-` index branch so we need the toolchain for that.

mosteo: Ah right.

mosteo (approved):

PR #517, opened 2022-05-16
Add external crates for BLAS, LAPACK

simonjwright:
These only support linux; not needed on macOS, and the msys2 libraries can't
be installed because of key signing errors.

If called in on macOS there will be warnings about unsatisfied external dependency: is there a way of saying `null`? e.g. `"macos" = []`?

Tested by using a [personal index](https://github.com/simonjwright/local-alire-index) while making a crate for [GNAT Math Extensions](https://github.com/simonjwright/gnat_math_extensions).

PR #521, opened 2022-05-19
Correct msys2 externals for libblas, liblapack

simonjwright:
Following the very helpful comments on #517, this update resulted in a successful build & run with windows-latest.

I’m not sure about the "empty origin for macos" part (the point being, that anyone developing on macOS will have these libraries installed as part of the SDK, so I wanted to avoid warnings about unsatisfied external references).
   
  * index/li/libblas/libblas-external.toml: uncomment the msys2 origin
      and make it to lapack. The reason is that we require libblas.dll,
      and this isn't provided by openblas (lapack provides both).
      Also, add an empty origin for macos.
  * index/li/liblapack/liblapack-external.toml: uncomment the msys2
      origin.
      Also, add an empty origin for macos.

### DIFF
--- a/index/li/libblas/libblas-external.toml
+++ b/index/li/libblas/libblas-external.toml
@@ -9,5 +9,6 @@ maintainers-logins = ["simonjwright"]
 [[external]]
 kind="system"
 [external.origin."case(distribution)"]
+"macos" = []
 "debian|ubuntu" = ["libblas-dev"]
-# "msys2" = ["mingw-w64-x86_64-openblas"]
+"msys2" = ["mingw-w64-x86_64-lapack"]
--- a/index/li/liblapack/liblapack-external.toml
+++ b/index/li/liblapack/liblapack-external.toml
@@ -9,5 +9,6 @@ maintainers-logins = ["simonjwright"]
 [[external]]
 kind="system"
 [external.origin."case(distribution)"]
+"macos" = []
 "debian|ubuntu" = ["liblapack-dev"]
-# "msys2" = ["mingw-w64-x86_64-lapack"]
+"msys2" = ["mingw-w64-x86_64-lapack"]


mosteo:
The empty detectors for macOS aren't really necessary. This won't help dependents on these libs, as they won't exist in the index in the case of macOS, even if there is a external definition.

There are two approaches here: 

1. Document somewhere (where?) that these libs aren't needed on macOS as they're part of the standard macOS environment, so a dependency on them shouldn't be included on macOS. Problem with this is that is a distributed overhead imposed onto any project using these libs across OSes, as they would have to make dependencies conditional on the OS.
2. On macOS, have a "version command" external instead of a library one to detect somehow their version. Isn't there some command that would echo their version if they're installed?
3. Think about some new kind of external detector that could cover this kind of situation, if 2) doesn't cut it.

The purpose of external detectors is precisely to ascertain the version that is installed when it's impossible to known in advance. Their current limitation is that we cannot have a detector that says "they're there but I don't know the version" (nor I'm sure that would be desirable).

mosteo:
I'm thinking that a further solution would be to have "fake" real (but empty) crates for macOS that virtually `provide` that dependency. Although, this would probably be too heavy handed and we would still have a problem with versions being unknown. It would be simpler if we can think of a detector for macOS.

simonjwright:
> 2\. On macOS, have a "version command" external instead of a library one to detect somehow their version. Isn't there some command that would echo their version if they're installed?

Lapack has a procedural interface (ILAVER) to return the version. The Apple version reported by it is 3.2.1, the current Netlib one is at least 3.10.
Blas doesn’t have such a facility.

Also, it’s stated (somewhere) that Apple have done some heavy optimisation.

Maybe the best thing would be to detect the current version of the SDK? Actually, that seem to be the current release of the OS, so not helpful.

What about the SDK’s build version? on my machine, 21E226:
```
$ xcrun --show-sdk-build-version 2>/dev/null
21E226
```

We could of course go the Homebrew route, much though I personally dislike it, and in this particular instance it seem a waste. It would certainly fit with the Alire philosophy. And, if someone comes up with a binding to a C library that’s not available in a Mac SDK but is available via Homebrew ... I’ve almost persuaded myself! I could have a look, if you don’t have anyone lined up already?

mosteo (approved):

mosteo:
Thanks for the info, Simon. So there's nothing that we can already use with what's in Alire.

Homebrew sounds appealing for a system package manager for macOS but no need to rush to it.

Anyway, no need to delay this PR for the particular issue with macOS.

PR #524, opened 2022-05-21
Add Getopt, Linenoise_Ada, TLSAda, & URI_MIME

nytpu:
I added these all together into one pull request rather than deluging you all in several at once, but I'll split up the pull request for each individual crate if you'd prefer.

### DIFF
--- a/index/li/libtls/libtls-external.toml
+++ b/index/li/libtls/libtls-external.toml
@@ -11,7 +11,5 @@ maintainers-logins = ["nytpu"]
 kind = "system"
 [external.origin."case(distribution)"]
 "debian|ubuntu" = ["libtls-dev"]
+"arch|centos|fedora" = ["libretls"]
 msys2 = ["mingw-w64-libressl"]
-arch = ["libretls", "libressl"]
-centos = ["libretls"]
-fedora = ["libretls", "libressl"]
--- a/index/li/libtls/libtls-external.toml
+++ b/index/li/libtls/libtls-external.toml
@@ -1,5 +1,5 @@
 name = "libtls"
-description = "a new TLS library, designed to make it easier to write foolproof applications"
+description = "A new TLS library"
 website = "https://man.openbsd.org/tls_init.3"
 tags = ["tls", "ssl"]
 licenses = "ISC"
--- a/index/li/libtls/libtls-external.toml
+++ b/index/li/libtls/libtls-external.toml
@@ -1,0 +1,17 @@
+name = "libtls"
+description = "a new TLS library, designed to make it easier to write foolproof applications"
+website = "https://man.openbsd.org/tls_init.3"
+tags = ["tls", "ssl"]
+licenses = "ISC"
+
+maintainers = ["nytpu <alex@nytpu.com>"]
+maintainers-logins = ["nytpu"]
+
+[[external]]
+kind = "system"
+[external.origin."case(distribution)"]
+"debian|ubuntu" = ["libtls-dev"]
+msys2 = ["mingw-w64-libressl"]
+arch = ["libretls", "libressl"]
+centos = ["libretls"]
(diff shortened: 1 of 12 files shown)


mosteo:
Hi @nytpu , there's in principle no problem with submitting all at once, except that a failure in one of them can block the merging of others. 

The failing "Check PR" check is ours to fix. However, there are a couple of other problems that you can see in the details before these can be merged.

nytpu:
I believe I fixed most of the pending issues, however I don't understand what this issue could possibly be: <https://github.com/alire-project/alire-index/runs/6551652983?check_suite_focus=true#step:8:136>, since there *are* GPR externals defined, and the `alire.toml` is defined pretty much identically to all of my other crates' that seemed to have worked fine.

Fabien-Chouteau:
```
   compilation of linenoise.adb failed

ERROR: Command ["gprbuild", "-gnatwU", "-j0", "-p", "-P", "linenoise.gpr"] exited with code 4
ERROR: Compilation failed.
, stderr gnat1: bad -gnaty switch (invalid style switch)
```
One of the style option might not be supported by GNAT 9.

nytpu:
Hmm, taking another look at this again and I don't understand the errors still. For almost all of the builds it just says right in the middle “the task was cancelled”, oftentimes before it even finishes setting up the toolchain or downloaded any of my crates.

For the one that does give a concrete error, it just says `bad -gnaty switch (invalid style switch)`, which I don't understand why it doesn't just… tell me what switch is bad. I believe it's `-gnaty-D` so I'm going and disabling that in all my gprbuilds now.

index/tl/tlsada/tlsada-2.4.1.toml
```diff
@@ -0,0 +1,30 @@
...
+
+website = "https://git.sr.ht/~nytpu/tlsada"
+tags = ["tls", "ssl"]
+
+authors = ["nytpu"]
+maintainers = ["nytpu <alex@nytpu.com>"]
+maintainers-logins = ["nytpu"]
+
+[[depends-on]]
+libtls = "^3.5"
```
mosteo, line 14: The external definition for `libtls` is missing. Do you have one that you forgot to include, or do you need help with crafting it?

index/li/libtls/libtls-external.toml
```diff
@@ -0,0 +1,17 @@
...
+maintainers-logins = ["nytpu"]
+
+[[external]]
+kind = "system"
+[external.origin."case(distribution)"]
+"debian|ubuntu" = ["libtls-dev"]
+msys2 = ["mingw-w64-libressl"]
+arch = ["libretls", "libressl"]
+centos = ["libretls"]
+fedora = ["libretls", "libressl"]
```
mosteo: The error in the latest checks is coming from arch/fedora having two different packages specified here. The idea is that only one of these should be found, and supply libtls.

@nytpu, I need your help here: in arch, for example, what is the relationship between libretls and libressl? 

In the short term for Alire, the fix is to only have one system package listed per distro

nytpu: [LibreSSL](https://www.libressl.org/) is a fork of OpenSSL that natively provides (and originally created) libtls. [LibreTLS](https://git.causal.agency/libretls/about/) is a port of libtls from LibreSSL to OpenSSL. They are mutually exclusive (as are OpenSSL and LibreSSL) however both provide libtls. I assumed, being a list of packages, that it'd be a list of packages where any one of which provides the library but on second thought it definitely is meant to define multiple packages that would collectively provide the library. Since OpenSSL is the default on all those distros I'll switch it to solely specify LibreTLS.

mosteo: Understood. Thanks, @nytpu.

mosteo (approved):

PR #532, opened 2022-06-05
Add J2Ada translator to Alire.

Blady-Com:


Fabien-Chouteau:
@Blady-Com I think we see build error in the checks because of the ongoing transition to 1.2. Stay tuned,

mosteo:
It seems some compiler versions have trouble here:
```
insrc.adb:290:81: error: prefix of "Overlaps_Storage" attribute must be object
insrc.adb:688:29: warning: condition can only be True if invalid values present
insrc.adb:688:29: warning: condition is always False
insrc.adb:688:29: warning: (see test at line 687)
javaseq.adb:4457:30: warning: "Event" modified by call, but value might not be referenced
javaseq.adb:4457:37: warning: "Token" modified by call, but value might not be referenced
```

Fabien-Chouteau (approved):

PR #533, opened 2022-06-05
libkeccak 3.0.0

damaki:


index/li/libkeccak/libkeccak-3.0.0.toml
```diff
@@ -0,0 +1,32 @@
...
+    "kangarootwelve"
+]
+
+authors = ["Daniel King"]
+maintainers = ["Daniel King <damaki.gh@gmail.com>"]
+maintainers-logins = ["damaki"]
+
+[[depends-on]]
+gnat = ">=11.0.0"
+gnatprove = "^11.2.3"
```
mosteo: If this dependency on `gnatprove` is used to prove the own library, but is irrelevant to users of the library, it would be better not having it here, as this is a heavy dependency.

The recommended workflow is to have a nested crate that depends on your library and on gnatprove, that you use for the verification. This nested crate needs not to be explicitly published, but can be distributed normally with the main library sources. See here: https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing

mosteo (changes requested):

index/li/libkeccak/libkeccak-3.0.0.toml
damaki: Thanks for the recommendation. I have edited the file in commit fd1e907 to remove the dependency on `gnatprove`.

mosteo (approved):

mosteo:
thanks, @damaki

PR #537, opened 2022-06-09
3 releases

pjljvandelaar:
Did I forget this?

### DIFF
--- a/index/de/dependency_graph_extractor/dependency_graph_extractor-22.0.0.toml
+++ b/index/de/dependency_graph_extractor/dependency_graph_extractor-22.0.0.toml
@@ -1,0 +1,20 @@
+name = "dependency_graph_extractor"
+description = "Extract dependency information from Ada projects"
+version = "22.0.0"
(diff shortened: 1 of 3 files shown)


mosteo:
There seems to be a problem with some compiler versions:

`rejuvenation-utils.adb:4:18: warning: no entities of "GNATCOLL.GMP.Integers" are referenced [-gnatwu]`

Fabien-Chouteau:
It's probably the GNATcoll crate that should get a restriction on GNAT versions.

pjljvandelaar:
What is alire's best practice?
What is now the best way forward?

* Reject this pull request completely, and make new releases?
* Make a new release with the same release number (22.0.1)?
* Make a new release with an incremented number (22.0.2)?

Fabien-Chouteau:
@pjljvandelaar I don't think you need to make a new release.

Something I missed before is that you are contributing to the `stable-1.1` branch of the index, but your project is made with Alire 1.2. You should open the pull request against the `stable-1.2` branch instead.

PR #538, opened 2022-06-09
cortex_m 0.4.0

Fabien-Chouteau:


Fabien-Chouteau:
Builds will fail because of configuration variable without a default

mosteo:
The error in the checks is unrelated to the crate, I'm working on it separately. I've verified locally that the crate errs in the expected manner advanced by you, @Fabien-Chouteau .

mosteo (approved):

PR #539, opened 2022-06-09
Build submissions in release mode

mosteo:
As normally dependencies/executables are built in release mode, we better check any submissions in this mode instead of the default development. 

Musing: should we go overboard and check the three profile modes?


Fabien-Chouteau (approved):

Fabien-Chouteau:
Good catch!

> Musing: should we go overboard and check the three profile modes?

No because those profiles are more sensible to compiler versions as new warnings can be introduced for instance.

mosteo:
Tested at https://github.com/mosteo/alire-index/pull/51 before merging

PR #547, opened 2022-06-14
gnat_math_extensions 1.1.0

simonjwright:
This project started in 2010.

It provides additional matrix capabilities beyond those defined in the Standard Library (Annex G of the ARM).

The Standard defines real and complex matrix and vector operations. Not every possible operation is supported, so for example only symmetric real or [hermitian](http://en.wikipedia.org/wiki/Hermitian_matrix) complex matrices can be solved.

The implementations, which are only for GNAT, are in the form of bindings to the [LAPACK](https://www.netlib.org/lapack) and [BLAS](https://www.netlib.org/blas/") libraries which are widely available if not already provided on all operating systems.


mosteo:
Tested with the fix in our workflow in https://github.com/mosteo/alire-index/pull/53 and https://github.com/mosteo/alire-index/pull/54, so it can be merged.

simonjwright:
Great, thanks!

PR #548, opened 2022-06-17
TLSAda 2.4.1 -> 2.4.2

nytpu:


mosteo:
The tests are not finding `libtls` in any of the platforms. I wonder if the external definitions are wrong; for example, in Ubuntu I can install `libssl-dev` but there is no `libtls-dev` (which is the one in the external)

nytpu:
@mosteo 
`libtls` as a standalone library (as opposed to having to completely replace the system SSL/TLS library) is unfortunately very new to many package managers. Looks like it made it into Ubuntu Jammy but wasn't in any older version. I missed this when adding the pacakge but it's currently also only in Debian Bookworm, not Bullseye.

I don't understand why it's not finding it for Arch, Fedora, et al. at least though; it is assuredly present and the packages have existed for years.

- https://packages.ubuntu.com/jammy/libtls-dev
- https://packages.debian.org/bookworm/libtls-dev
- https://archlinux.org/packages/community/x86_64/libretls/
- https://packages.fedoraproject.org/pkgs/libretls/libretls/

Is it an issue with the version I'm depending on in TLSAda? I don't actually have a hard dependency on that version, I just set it to `^3.5` because 3.5 is the only version I've thoroughly tested it on; but I'm 99.9% sure specifying “any libtls version” would be safe (the library's been pretty static throughout its existence). Otherwise I don't see any reason why it's not installing the system package.

mosteo:
I went digging and for Arch at least the problem is our docker image is a bit outdated, and `libretls` there is 3.4. I will update it to at least have a tested platform.

mosteo:
Tested working at https://github.com/mosteo/alire-index/pull/55, so merging now, thanks for your patience.

PR #550, opened 2022-06-17
Gnat and GNATprove 12

Fabien-Chouteau:


index/gn/gnat_arm_elf/gnat_arm_elf-12.1.1.toml
```diff
@@ -0,0 +1,27 @@
...
+url = "https://github.com/alire-project/GNAT-FSF-builds/releases/download/gnat-12.1.0-1/gnat-arm-elf-linux64-12.1.0-1.tar.gz"
+hashes = ["sha256:a11c25f884a637ad48b2f5cde81eea7d2598c78ee5058bd8bdb90b2376b6fb47"]
+
+[origin."case(os)".macos."case(word-size)".bits-64]
+url = "https://github.com/alire-project/GNAT-FSF-builds/releases/download/gnat-12.1.0-1/gnat-arm-elf-darwin-12.1.0-1.tar.gz"
+hashes = ["sha256:c148f4ec11dcfb702e77f14fad5efc1ca54b76177628be336edabac738d569fb"]
+
+[origin."case(os)".windows."case(word-size)".bits-64]
+url = "https://github.com/alire-project/GNAT-FSF-builds/releases/download/gnat-12.1.0-1/gnat-arm-elf-windows64-12.1.0-1.tar.gz"
+hashes = ["sha256:7fb7c7b9689cad9901843f78187e9077f81f887aa0f79af6a6df78362be9d500"]
```
mosteo, line 27: Alire should be ready now to start using the appropriate `case(host-arch)` extra filters here.

mosteo:
@Fabien-Chouteau , is it OK to merge this one? Just double checking

Fabien-Chouteau:
I pushed a new version with updated binutils.
I also wanted to use the `host-arch` case but as you saw this is not available in `1.2`.

I still want some feedback on the 12.1.2 but we should be able to merge by the end of the week.

mosteo:
> I also wanted to use the `host-arch` case but as you saw this is not available in `1.2`.

Since using it will be breaking for `1.2.0` clients, we can dispense with it for the moment. At least while there aren't compilers not for x86_64 there's no possibility of confusion.

Fabien-Chouteau:
@mosteo I think we are good to go

mosteo:
I can't understand now why tar is giving us pain again on Windows. In any case the problem isn't with the manifest file, so I'm merging.

PR #552, opened 2022-06-17
Add a bit more macOS support for gnatcoll friends.

Blady-Com:
Add a bit more macOS support for gnatcoll, gnatcoll_iconv, gnatcoll_sql, gnatcoll_sqlite, gnatcoll_zlib, gnatcoll_syslog, gnatcoll_readline, gnatcoll_python3, gnatcoll_lzma.
Tested with MacPorts (Python 3.9, lzma, zlib, readline).

### DIFF
--- a/index/gn/gnatcoll_python3/gnatcoll_python3-21.0.0.toml
+++ b/index/gn/gnatcoll_python3/gnatcoll_python3-21.0.0.toml
@@ -17,8 +17,14 @@ LIBRARY_TYPE = ["static", "relocatable", "static-pic"]
 "debian|ubuntu" = { C_INCLUDE_PATH.append = "${DISTRIB_ROOT}/usr/include/python3.7/:${DISTRIB_ROOT}/usr/include/python3.8/"}
 msys2 = { C_INCLUDE_PATH.append = "${DISTRIB_ROOT}/mingw64/include/python3.8/" }
 
+[environment.'case(os)']
+macos = { C_INCLUDE_PATH.append = "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/include/python3.9" }
+
 [available.'case(distribution)']
 'debian|ubuntu|msys2' = true
+
+[available.'case(os)']
+'macos' = true
 '...' = false
 
 [[depends-on]]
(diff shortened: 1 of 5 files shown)


index/gn/gnatcoll_python3/gnatcoll_python3-21.0.0.toml
```diff
@@ -17,8 +17,14 @@ LIBRARY_TYPE = ["static", "relocatable", "static-pic"]
 "debian|ubuntu" = { C_INCLUDE_PATH.append = "${DISTRIB_ROOT}/usr/include/python3.7/:${DISTRIB_ROOT}/usr/include/python3.8/"}
 msys2 = { C_INCLUDE_PATH.append = "${DISTRIB_ROOT}/mingw64/include/python3.8/" }
 
+[environment.'case(os)']
+macos = { C_INCLUDE_PATH.append = "/opt/local/Library/Frameworks/Python.framework/Versions/3.9/include/python3.9" }
```
Fabien-Chouteau: I am not a huge fan of absolute paths like that. Is `/opt/local` coming from Homebew?

Fabien-Chouteau (changes requested):

index/gn/gnatcoll_python3/gnatcoll_python3-21.0.0.toml
Blady-Com: Well, so am I.
This path is the standard installation path of MacPorts.
There is also a Python version issue.
As Python version was hardcoded to 3.7 and 3.8 for debian, ubuntu, 3.8 for msys2, I thought the path could be too ;-)
For macOS, there are two questions in order to build the include path:
- how to get the relevant Python installation (Apple, MacPorts, Brew, adhoc...)?
- how to get the expected Python version?

Fabien-Chouteau: I don't think we want something that works on MacPorts but not on Homebew. So for me the first step is to add proper support of "system package managers" for macOS. And then have this in the same case statement as the others.

In the meantime I am not in favor of this patch.

mosteo:
@Blady-Com, for the time being maybe you can remove the release with the hard-coded paths and leave the external detectors. On that note, if those are intended for macOS only, you could add an
```
[external.available.'case(os)']
macos = true
'...' = false
```
definition.

index/gn/gnatcoll_python3/gnatcoll_python3-21.0.0.toml
Blady-Com: I have reverted too macOS specific changes.
How can I help with macOS support?

mosteo:
We discussed recently at gitter that using `pkg-config` cannot work as it still requires that the user use its output to set up environment. So crates depending on a version detected by `pkg-config` won't work as-is.

We may add support for that, but at this time what we do have is support in the master branch for Homebrew. So maybe that's a better approach for the time being.

PR #554, opened 2022-06-19
Create hac-0.11.0.toml

zertovitch:


zertovitch:
Some builds' fails can be explained: it is due to the lack of support for 128-bit integers.

zertovitch:
The reliance on 128-bit integers makes HAC too restrictive in respect to the compiler choice. I'll use something else for that computation...

PR #559, opened 2022-06-22
Fix xmlada 21 on Windows

Fabien-Chouteau:


mosteo:
Arch failure is because #557 is needed

Fabien-Chouteau:
Can we still merge because this is only for windows?

mosteo:
Sure, done.

Fabien-Chouteau:
Thanks

PR #565, opened 2022-07-02
Add a bit more macOS support for simple_components.

Blady-Com:


index/li/libgnutls/libgnutls-external.toml
```diff
@@ -9,3 +9,8 @@ kind = "system"
 [external.origin."case(distribution)"]
 "..." = []
 "debian|ubuntu" = ["libgnutls28-dev"]
+
+[[external]]
+kind = "version-output"
```
Fabien-Chouteau, line 14: @mosteo I don't know if the mix of external origin and external version-command is supported.

mosteo: It should pose no problem, although we don't have well-defined priorities for when both methods report the same exact version. Since the version output can only succeed for an already installed package, it should make little or no difference either way. I'll add a test for this case just to be sure.

mosteo: Actually the `make` crate is already working like this, with both methods.

mosteo (approved):

PR #567, opened 2022-07-07
Separate check scripts to their own repository

mosteo:
Why do this:

- Users will see less "empty" updates as changes to these scripts don't result on updates to the index repo.
- We can reuse these checks from several repos.
  - This is eyeing that, down the road, we may want a "staging" repo that accepts PRs fully automatically.
- The repo containing the checks can have its own self-check workflows without complicating things here.

Tested in https://github.com/mosteo/alire-index/pull/60


Fabien-Chouteau (approved):
Sounds like a nice improvement

PR #568, opened 2022-07-12
Library for SimH Standard magnetic tape images

SMerrony:
Simh_tape is an Ada package for handling magnetic tape images in the
[standard format](http://simh.trailing-edge.com/docs/simh_magtape.pdf) used by 
[SimH](http://simh.trailing-edge.com/) and many other computer simulators and emulators.


index/si/simh_tapes/simh_tapes-0.1.1-dev.toml
```diff
@@ -0,0 +1,20 @@
...
+maintainers = ["Stephen Merrony <merrony@gmail.com>"]
+maintainers-logins = ["SMerrony"]
+licenses = "AGPL-3.0-or-later"
+website = "https://github.com/SMerrony/simh_tapes"
+tags = ["legacy", "magtape", "magnetic", "tape", "emulation", "simulation"]
+executables = ["simhtapetool"]
+
+[build-switches]
+development.optimization = ["-O0"]
+
```
mosteo: Unfortunately there is a known limitation in current Alire 1.2 that precludes the use of build switches in the community index (you can still use them locally), which is why the checks are failing.

So your options at this time are to remove these lines from the manifest you're submitting, or wait for a while until we release a maintenance version (around the end of July most likely). I apologize for the inconvenience.

SMerrony: OK - the error message was quite misleading!
The build-switches were just for debugging - will try again with them removed.

mosteo: The error was quite obscure, yes, because it's an unexpected situation deep within alr itself. Sorry about that!

index/si/simh_tapes/simh_tapes-0.1.1-dev.toml
```diff
@@ -0,0 +1,16 @@
+name = "simh_tapes"
+description = "Library to handle SimH Standard magnetic tape image files"
+version = "0.1.1-dev"
```
mosteo: I think the `-dev` should be removed here, right? Which means that the manifest file should also be renamed to `simh_tapes-0.1.1.toml`. (Or perhaps you meant to publish `0.1.0`?)

Otherwise, tests are passing, so that should be the last issue.

mosteo (changes requested):

index/si/simh_tapes/simh_tapes-0.1.1-dev.toml
SMerrony: alr added the -dev, so I assumed it was needed...

mosteo: That's just to signal that the version is intermediate and before a firm release, according to semver conventions. But I see how can be confusing the first time.

mosteo (approved):

mosteo:
Merged, thanks!

index/si/simh_tapes/simh_tapes-0.1.1-dev.toml
Fabien-Chouteau: BTW @SMerrony, why do you want to set `-O0` instead of `-Og` here?

jquorning:
`alr test simh_tapes` says:
ERROR: Declared executable not found after compilation: simhtapetool
`simtapetool` belongs to a sub-crate in the `simtapetool` directory. If the `executables` declaration is removed all should be fine.

@mosteo The Build Crate action detects the missing executable but does not fail [1].

[1] https://github.com/alire-project/alire-index/runs/7342433593?check_suite_focus=true#step:10:285

mosteo:
Ouch, good catch

SMerrony:
Please note that I am coming from the Go(lang) world where it is normal for a package to contain a demo/sample executable alongside the actual library.  I really don't see the need for two crates to achieve this end.
In any case, could this common use-case be documented?

jquorning:
The sub-crate with the tool is all fine and there is no need for two crates. The executable is just not part of the main crate.

mosteo:
It's not a big problem as this doesn't break the use of the library by dependents. It will be enough to remove the `executables` declaration (which wouldn't need a version bump btw). 

And since the demo is already in a subcrate, that is already following our current recommendations.

There is already some documentation in the [pins section](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing) but you're right that it doesn't mention demos, although the idea is the same. I too think it'll be good having an entry about this on the best practices section.

PR #569, opened 2022-07-15
Add lmdb_ada 1.0.1

nytpu:
Add thick bindings to the [Lightning Memory Mapped Database](https://www.symas.com/lmdb).


index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
```diff
@@ -0,0 +1,47 @@
...
+project-files = ["lmdb.gpr"]
+
+website = "https://git.sr.ht/~nytpu/lmdb-ada"
+tags = ["database", "key-value"]
+
+authors = ["nytpu"]
+maintainers = ["nytpu <alex@nytpu.com>"]
+maintainers-logins = ["nytpu"]
+
+[available.'case(os)']
```
Fabien-Chouteau: Since all the entires are `true`, you can probably remove the `[available.'case(os)']` altogether.

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
```diff
@@ -0,0 +1,47 @@
...
+linux = true
+macos = true
+# lmdb supports windows' thread library as well as pthreads
+windows = true
+#'...' = false
+
+[gpr-externals]
```
Fabien-Chouteau: You don't have to list all the `gpr-externals` here.

BTW, did you check the compiler switches support in Alire 1.2?
It simplifies a lot the project files and provides a standard way to deal with compilation profiles.

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
```diff
@@ -0,0 +1,47 @@
...
+PLATFORM = ["unix", "windows"]
+LIBRARY_TYPE = ["dynamic", "relocatable", "static", "static-pic"]
+COMPILE_CHECKS = ["enabled", "disabled"]
+RUNTIME_CHECKS = ["enabled", "disabled"]
+STYLE_CHECKS = ["enabled", "disabled"]
+CONTRACTS = ["enabled", "disabled"]
+BUILD_MODE = ["debug", "optimize"]
+
+[gpr-set-externals]
+GNATCOLL_BUILD_MODE = "PROD"
```
Fabien-Chouteau: A crate is not supposed to change the build mode of its dependencies. Maybe we should change the GNATcoll crate to build in PROD mode by default, if it's not already the case.

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
```diff
@@ -0,0 +1,47 @@
...
+
+[gpr-set-externals.'case(os)']
+'linux|macos' = { PLATFORM = "unix" }
```
Fabien-Chouteau: Staring with Alire 1.2, we now have information about the host platform generated in the `config/<CRATE_NAME>_config.gpr`:

```
[...]
   Alire_Host_OS := "linux";

   Alire_Host_Arch := "x86_64";
[...]
```
You can use that in your project file. See https://alire.ada.dev/docs/#platform-specific-code

To use those new features, remove the `lmdb_ada.gpr` file and generate a new one with `alr init --lib --in-place lmdb_ada`.

Fabien-Chouteau (dismissed):
Hi @nytpu, thanks for the contribution!

You crate can be merged as is, however I have a few comments to help you better use the features of Alire. Let us know if you want to update.

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
nytpu: Ah yeah, I left that in accidentally when I was writing the library because it was originally just
```
linux = true
macos = true
'...' = false
```
But then I realized LMDB supported windows and added the windows line without thinking about removing the section altogether.

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
nytpu: Hmm, the documentation on `gpr-externals` is pretty sparse so I just assumed it was good to enumerate all possible external variables there. Is there a recommendation on what I should list there or not? Just custom build variables but not common ones like `LIBRARY_TYPE`?

And I'll check out the switches support too

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
nytpu: This was breaking the build of the package (GNATcoll built fine from AUR and from a normal source clone but the build would terminate on the first warning when building with alr). Again, probably should've removed it, just put it in so I could actually test my own package instead of trying to fix dependencies.

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
nytpu: Hmm, I'd like to avoid *requiring* alr to use the library which is why I avoided using Alire's config system; I've had my fair share of having to write really dumb patches for other things just because I want to install it with my system's package manager instead of the programming language's custom overlay package manager. Especially since the granularity of my platform switching is literally “POSIX or Windows”, and the only thing that configuration option changes is whether to link `libpthread` or not.

If this `gpr-set-externals` solution isn't acceptable, I'll likely keep the current GPR file as a “non-Alire project file” and have a separate “Alire project file”.

nytpu:
@Fabien-Chouteau I've prepared an update with some changes, however I can't yet publish it due to the strange build termination. It looks like `-Werror` is enabled on non-PROD GNATcoll builds, because the build terminates the instant there is a warning: https://paste.sr.ht/~nytpu/e0d3101f1c12d95db784e4509624edaffb47b502

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
Fabien-Chouteau: I just opened a PR to set the build mode for gnatcoll: #570.

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
Fabien-Chouteau: Indeed we have to clarify the use of `gpr-externals`. As far as I know this is only documentation of what's in the gpr file.
@mosteo what's the intent of this feature?

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
Fabien-Chouteau: Up to you, `gpr-set-extrenals` is acceptable of course. However you should prefix you scenario variables with the name of the crate, e.g. : `LMDB_ADA_PLATFORM` otherwise you are running a high risk of getting conflicts with other crates that will use the same name but with other values.

index/lm/lmdb_ada/lmdb_ada-1.0.1.toml
mosteo: Yes, the intent was to document what library clients can set via `gpr-set-externals` without going to the project files. In practice it does nothing.

mosteo:
Hi @nytpu, is this one ready after the latest pushes?

nytpu:
@mosteo Yep, it is!

mosteo (approved):

mosteo:
Merged, thanks.

PR #579, opened 2022-07-25
Update lvgl_ada_examples-0.1.0.toml

Fabien-Chouteau:
Fix dependencies.

~~@mosteo  This will still fail to build because of incompatibility with SDLada. I don't know why this code is not compatible and I don't have much time to fix it...~~


mosteo:
The manifests look good but there are still some failed checks. Waiting for instructions...

Fabien-Chouteau:
Seems like an issue with the sdlada crate. If that's ok for you I would say we can merge as this is still an improvement.

mosteo:
OK, right.

PR #590, opened 2022-07-31
utilada=2.1.0: Constraints on compiler

jquorning:
Also fix readthedocs url in long-description.


jquorning:
Actually I do not know if utilada compiles with CE2021 or later.

mosteo (approved):

mosteo:
> Actually I do not know if utilada compiles with CE2021 or later.

Given that CE is discontinued and we don't test with it anymore, let's go with it.

jquorning:
Guess it should have been "<11" instead of "<=10"..?

https://github.com/alire-project/alire-index/runs/7597046662#step:9:404

mosteo:
Oh right, as `<=10` will be interpreted as `<=10.0.0`... leaving out all of `^10`

PR #591, opened 2022-07-31
utilada=2.1.0: Fix compiler constraint

jquorning:


mosteo (dismissed):

index/ut/utilada/utilada-2.1.0.toml
```diff
@@ -15,7 +15,7 @@ UTIL_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
 UTIL_OS = ["win32", "win64", "linux32", "linux64", "macos64", "netbsd32", "netbsd64", "freebsd32", "freebsd64"]
 
 [[depends-on]]
-gnat = "<=10 | (>=2000 & <=2020)"
+gnat = "<11 | (>=2000 & <=2020)"
```
mosteo: And, just to be sure, should this `<=2020` be a `<2021`?

jquorning:
Why is build skipped on fedora? Isn't gnat-9 available?

https://github.com/alire-project/alire-index/runs/7598303553/#step:9:407
https://github.com/alire-project/alire-index/runs/7598303553/#step:5:99

mosteo:
> Why is build skipped on fedora? Isn't gnat-9 available?

That's because the actual GNAT in use is the one from the docker image, which is v12. The gnat installed by the workflow is used only for tests with a development `alr`, which requires a compiler to be built.

That's confusing (I had forgotten myself), yeah, I'll update the test script to show the actual GNAT in the environment.

But at least that means this PR is good to go.

mosteo (approved):

PR #594, opened 2022-08-02
Add crate for the new Ada Utility 2.5.0 library

stcarrez:
This is an update of Ada Utility Library. I've have not included yet the `utilada_lzma` because it needs to have the `lzmada` 1.1.2 integrated first.

### DIFF
--- a/index/ut/utilada/utilada-2.5.0.toml
+++ b/index/ut/utilada/utilada-2.5.0.toml
@@ -15,9 +15,6 @@ BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
 UTIL_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
 UTIL_OS = ["win32", "win64", "linux32", "linux64", "macos64", "netbsd32", "netbsd64", "freebsd32", "freebsd64"]
 
-[gpr-set-externals]
-BUILD = "distrib"
-UTIL_LIBRARY_TYPE = "static"
 [gpr-set-externals."case(os)".linux."case(word-size)".bits-32]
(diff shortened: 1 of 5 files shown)


index/ut/utilada/utilada-2.5.0.toml
```diff
@@ -0,0 +1,41 @@
...
+website = "https://gitlab.com/stcarrez/ada-util"
+
+[gpr-externals]
+BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
+UTIL_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
+UTIL_OS = ["win32", "win64", "linux32", "linux64", "macos64", "netbsd32", "netbsd64", "freebsd32", "freebsd64"]
+
+[gpr-set-externals]
+BUILD = "distrib"
+UTIL_LIBRARY_TYPE = "static"
```
mosteo: Hi @stcarrez. These externals look that should be left unset so clients of the library have the choice of them. The externals set by Alire are not defaults, and cannot be changed by clients. Defaults should be tackled in the GPR file.

If you already have defaults in the project files, it is enough to update this PR's manifest, there's no need to change the commit or republish.

(The ones setting OS, word size etc do make sense to be set already here.)

mosteo (changes requested):

index/ut/utilada/utilada-2.5.0.toml
stcarrez: Ah... Thanks, I'll checked my other Alire crates for that.

mosteo (approved):

PR #601, opened 2022-08-03
Upgrade Simple Components to version 4.62.

Blady-Com:


mosteo (approved):

mosteo:
It builds in the only platform with all required dependencies, thanks. In that regard, @Blady-Com, there's an effort being set up to repackage these in several crates according to dependencies in a way that's maintainable for the future. Just drop me an email at amosteo@unizar.es if you want to get involved.

PR #609, opened 2022-08-12
Updated crates to use configuration

onox:


mosteo:
Hi @onox, if you recreate this PR on top of branch `stable-1.2.1`, it should work now. It is not enough changing the base branch as the actions still use the old base branch.

PR #611, opened 2022-08-15
Drop setting of `LIBRARY_TYPE`

reznikmm:
This lets the user choose between static/relocatable versions of ALS and fixes
[an issue](https://github.com/AdaCore/ada_language_server/issues/1026)
with system compiler.

### DIFF
--- a/index/ad/ada_language_server/ada_language_server-22.0.0.toml
+++ b/index/ad/ada_language_server/ada_language_server-22.0.0.toml
@@ -9,9 +9,6 @@ maintainers-logins = ["reznikmm"]
 project-files = ["gnat/lsp_server.gpr"]
 tags = ["lsp", "vscode"]
 
-[gpr-set-externals]
-LIBRARY_TYPE = "static"
-
 [environment]
 ADA_PROJECT_PATH.set= "${CRATE_ROOT}/subprojects/stubs"
 


mosteo:
Hi @Blady-Com , maybe you can shed some light into the issue here with `gmp` for macOS (since you added the support). I'm seeing a failure [here](https://github.com/alire-project/alire-index/runs/7837323380?check_suite_focus=true) where `gmp` is properly detected by alr but then not found during compilation.

This makes me think that something else has to be done in some cases? Something like `pkg-config --libs --cflags gmp`? Clearly this is not the case in e.g. Ubuntu, but do you know about macOS?

Pinging @simonjwright as the resident macOS guru :-)

@reznikmm about the Arch failure, it seems warnings are being treated as errors?

reznikmm:
> @reznikmm about the Arch failure, it seems warnings are being treated as errors?

Yes, somewhere in libadalang-tools. It has been fixed after 22 release.  I'm not sure how to fix this in already released code. I can add a post-fetch action like `sed -i -e /gnatwae/d <project>.gpr`.

simonjwright:
> Hi @Blady-Com , maybe you can shed some light into the issue here with `gmp` for macOS (since you added the support). I'm seeing a failure [here](https://github.com/alire-project/alire-index/runs/7837323380?check_suite_focus=true) where `gmp` is properly detected by alr but then not found during compilation.
> 
> This makes me think that something else has to be done in some cases? Something like `pkg-config --libs --cflags gmp`? Clearly this is not the case in e.g. Ubuntu, but do you know about macOS?
> 
> Pinging @simonjwright as the resident macOS guru :-)

macOS doesn't come with GMP.

When I build GCC, I use `contrib/download_prerequisites` to get GMP, MPFR, MPC and others; the build process finds them in the build tree and builds a static version. This is only used inside the compiler, and so far as I can remember neither the library not the include files are installed. Instead, I separately build both static and shared versions using the new compiler and install over the top of it (the shared version is needed for one of the Python tools, I think - `libadalang`?)

In the compilers built for Alire (as was the case for all community macOS compiler releases), these libraries are pre-built outside the GCC tree. This means that there's no `gmp.h` (or `libgmp.*`).

I wonder whether @Blady-Com was using my compiler build?

I guess we could ask users to install libgmp using Homebrew. If we put a hint to that effect in the external toml, will it only show up if there’s no source stated for this OS? I’m not sure whether there’s an `aarch64-apple-darwin` version of libgmp.

simonjwright:
>>@reznikmm about the Arch failure, it seems warnings are being treated as errors?

>Yes, somewhere in libadalang-tools. It has been fixed after 22 release. I'm not sure how to fix this in already released code. I can add a post-fetch action like `sed -i -e /gnatwae/d <project>.gpr`.

`-gnatwn` cancels `-gnatwe`/`-gnatws`

Blady-Com:
> > > @reznikmm about the Arch failure, it seems warnings are being treated as errors?
> > Yes, somewhere in libadalang-tools. It has been fixed after 22 release. I'm not sure how to fix this in already released code. I can add a post-fetch action like `sed -i -e /gnatwae/d <project>.gpr`.
> `-gnatwn` cancels `-gnatwe`/`-gnatws`

Yes my command line is:
`./setup.py build --debug --gpr-opts=-gnatwn --prefix=$some_prefix`

Blady-Com:
> I wonder whether @Blady-Com was using my compiler build?

I did, but now I used the one from Alire in order to have same distro as for other Alire users.
 
> I guess we could ask users to install libgmp using Homebrew. If we put a hint to that effect in the external toml, will it only show up if there’s no source stated for this OS? I’m not sure whether there’s an `aarch64-apple-darwin` version of libgmp.

Well, first, yes, GMP needs to be installed whatever the origin, I used many.
But my inclination is for MacPorts.
Second, setup.py from GNATColl aren't aware of pkgconfig thus I told where GMP is installed with:
'C_INCLUDE_PATH=$some_prefix/include' and 'LIBRARY_PATH=$some_prefix/lib'

mosteo:
Well, my concern is that `pkg-config` is reporting `gmp` as installed in macOS, whereas compilation fails. If that's going to be the case by default (without user intervention), the externals relying on `pkg-config` are not useful for macOS and should be reverted.

mosteo:
On a more general note, can Homebrew and MacPorts be used intermixed?

simonjwright:
We _cannot_ rely on `pkg-config` being installed on macOS. Not, at any rate, unless we say it’s a prerequisite that `${package-manager}` be installed and have been used to install `${package}`.

If we decide we must have a package manager, I think I’d prefer Homebrew but people who use both (yes, it’s possible they say) have arguments both ways.

The log file says (line 32 in the "Test crate" log) `libgmp external: add Arch package (#586)` which I don’t understand - why `arch`?

reznikmm:
BTW, on MacOS gmp is installed on GitHub CI into /usr/local if I remember
correctly. In one of my build I do these to help the compiler find gmp.h

export CPATH=/usr/local/include:/mingw64/include
export LIBRARY_PATH=/usr/local/lib:/mingw64/lib
export DYLD_LIBRARY_PATH=/usr/local/lib

Blady-Com:
@mosteo 
> Well, my concern is that `pkg-config` is reporting `gmp` as installed in macOS, whereas compilation fails. If that's going to be the case by default (without user intervention), the externals relying on `pkg-config` are not useful for macOS and should be reverted.

I hope not, specifically speaking on gmp, it comes with pkg-config stuff. Thus the depends property checks both if pkg-config is installed / visible from PATH and if gmp is installed. If not, a futur macOS manager within Alire could install the missing items. For now, we can do that manually as GMP is an external dependency. If ok some variables for includes and lib have to be set in accordance with the actual GPR project or on the command line:

```
% PATH=/opt/local/bin:$PATH pkg-config --libs --cflags gmp 
-I/opt/local/include -L/opt/local/lib -lgmp
% PATH=/opt/local/bin:$PATH C_INCLUDE_PATH=/opt/local/include LIBRARY_PATH=/opt/local/lib alr build
Compile
...
   [C]            gnatcoll-gmp-mpz_odd_p.c
...
Build finished successfully in 7.12 seconds.
```

> On a more general note, can Homebrew and MacPorts be used intermixed?

Their installed location is different but I don't recommend to install both as they may interfere.

@simonjwright 
> We _cannot_ rely on `pkg-config` being installed on macOS. Not, at any rate, unless we say it’s a prerequisite that `${package-manager}` be installed and have been used to install `${package}`.

Even installed manually, that makes available a number of crates for macOS users while waiting for a integrated package manager.
I hope that it comes soon :-)

@reznikmm 
> BTW, on MacOS gmp is installed on GitHub CI into /usr/local if I remember correctly. In one of my build I do these to help the compiler find gmp.h

Great, could you please check if pkg-config is also available or how to check by myself?

simonjwright:
> BTW, on MacOS gmp is installed on GitHub CI into /usr/local if I remember correctly.

So it is. But that doesn’t help people to install on their own machines.

On the other hand, I see that the [macOS 11.6 runner](https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md) _includes Homebrew_ but not MacPorts! (and comes with Tcl/Tk already installed).  Maybe a hint there.

mosteo:
> We _cannot_ rely on `pkg-config` being installed on macOS.

This isn't a big concern as the external detector will silently skip it if unavailable. So for now, if it worked, it would be an improvement for users that do have `pkg-config` installed. Except that it isn't working as expected :-)

PR #612, opened 2022-08-16
Add crates for Ada Servlet 1.6.0

stcarrez:
I'll submit the servletada_ews crate when Simon publishes the ews crate

### DIFF
--- a/index/se/servletada/servletada-1.6.0.toml
+++ b/index/se/servletada/servletada-1.6.0.toml
@@ -1,0 +1,41 @@
+description = "Web Servlet Library following JSR 154, JSR 315 (Core)"
+tags = ["web", "servlet", "http", "json", "rest"]
+name = "servletada"
+version = "1.6.0"
+licenses = "Apache-2.0"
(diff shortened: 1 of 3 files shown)


simonjwright:
I’m about to publish ews, but there’s a little bit of an interface change; `ews-make_htdocs` has changed to `ews_generator` and has `-i`, `-o` switches to control input (web) source and output (Ada) directories.

index/se/servletada/servletada-1.6.0.toml
```diff
@@ -0,0 +1,41 @@
...
+
+"""
+
+[[depends-on]]
+security = "^1.4.1"
+utilada = "^2.5.0"
+elada = "^1.8.5"
+
+[gpr-externals]
+BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
```
mosteo, line 32: Suggestion for future releases is to rename `BUILD` to something like `SERVLET_BUILD` so it can be set individually.

mosteo (approved):

PR #613, opened 2022-08-16
Added EWS 1.11.0-1.

simonjwright:
I’ve uploaded this to `upload` in my fork, as instructed - did I get this right?

### DIFF
--- a/index/se/serverfaces/serverfaces-1.5.0.toml
+++ b/index/se/serverfaces/serverfaces-1.5.0.toml
@@ -1,0 +1,44 @@
+description = "Web Server Faces JSR 252, JSR 314 and JSR 344"
+tags = ["web", "servlet", "http", "json"]
+name = "serverfaces"
+version = "1.5.0"
+licenses = "Apache-2.0"
+authors = ["Stephane.Carrez@gmail.com"]
+maintainers = ["Stephane.Carrez@gmail.com"]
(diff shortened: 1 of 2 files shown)


upload/stable-1.2/index/ew/ews/ews-1.11.0-1.toml
```diff
@@ -0,0 +1,39 @@
...
+long-description = """
+
+EWS is a web server construction kit, designed for embedded
+applications using the GNAT Ada compiler.
+
+The project is hosted at https://github.com/simonjwright/ews.
+
+"""
+name = "ews"
+version = "1.11.0-1"
```
mosteo: Is this truly intended as a pre-release? (Asking because of the `-1` part in the version).

mosteo:
About these stuck checks, normally I would be able to restart them if they were hung after starting. This time however it seems they didn't event start.

I guess you can push a trivial commit (e.g. after merging or rebasing with the latest index changes) and that should do something. Otherwise you may have to close and reopen.

upload/stable-1.2/index/ew/ews/ews-1.11.0-1.toml
simonjwright: I think I’d best close & reopen as 1.11.0.

PR #617, opened 2022-08-20
Add crate for Dynamo 1.3.0 (code generator)

stcarrez:


mosteo:
It seems mariadb is a dependency?

stcarrez:
> It seems mariadb is a dependency?

Yes, you're right. I'll submit a libmariadb crate and update dynamo.

stcarrez:
> > It seems mariadb is a dependency?
> 
> Yes, you're right. I'll submit a libmariadb crate and update dynamo.

Now, the root issue comes from the `ado_mysql` library which should have a dependency to mariadb.
I'll submit an update of the `ado_mysql` crate when the libmariadb crate is integrated.

mosteo:
Just ping me here once everything is ready so I don't temporarily forget about this one.

stcarrez:
@mosteo since ado_mysql is now fixed with a dependency to libmariadb, it should work.

mosteo:
This can't succeed for the same reason as #619: some preinstalled packages in the ubuntu VMs conflict with libmariadb, and `alr` won't uninstall them. Testing locally everything seems OK, so I'm merging.

PR #619, opened 2022-08-22
Add crate for libmariadb

stcarrez:


mosteo:
The external definition looks OK to me, but it's unfortunate that it conflicts with something (in my system it's at least with `default-libmysqlclient-dev` and `libpoco-dev`). The user can deliberately do this, but it will show as failing in our tests as it won't succeed as long as removals are required.

I can only think of allowing removal with `--force`, and using that in tests so at least they can succeed. Meanwhile I will merge as the definition is fine.

Other ideas welcome.

PR #621, opened 2022-08-27
Create zipada-58.0.0.toml

zertovitch:
Moving on to Zip-Ada version 58.

### DIFF
--- a/index/zi/zipada/zipada-58.0.0.toml
+++ b/index/zi/zipada/zipada-58.0.0.toml
@@ -1,0 +1,34 @@
+name = "zipada"
+version = "58.0.0"
+authors = [ "Gautier de Montmollin", ]
+description = "Manage Zip Archives and raw LZMA streams"
+licenses = "MIT"
+tags = [ "zip" ]
+
+maintainers = [ "alejandro@mosteo.com" ]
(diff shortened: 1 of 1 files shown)


mosteo:
It seems linker switches for macOS are different: https://stackoverflow.com/a/24799865/761390

zertovitch:
In commit https://github.com/zertovitch/zip-ada/commit/70f273aca89b859ab5dbd2969c85e0414b299d5b
I make the linker use `-dead_strip` for Mac (scenario variable Zip_OS = "MacOSX").
But how to relate that to values known to Alire, through `os=mac`?

mosteo:
I see you found the way to use scenario variables.

In regard to [these warnings](https://github.com/alire-project/alire-index/runs/8050811060?check_suite_focus=true#step:9:565), maybe the Alire manifest requires updating to remove the not-built binaries?

zertovitch:
Do you mean the compilation diagnostics with `zip-compress-deflate.adb:361:05: info: in instantiation ` etc. ?
It's "infos", not real warnings. If it's disturbing, I can hide them via `-gnatw.N`.

mosteo:
Merged, thanks.

PR #626, opened 2022-08-30
Ada bindings for SDL2 0.1.0

leogermond:


onox:
sdl crate seems to be just generated thin bindings. Are you aware of the sdlada crate in alire?

index/gn/gnat_game/gnat_game-0.1.0-dev.toml
```diff
@@ -0,0 +1,15 @@
+name = "gnat_game"
+description = "Shiny new project"
```
onox, line 2: Description is missing.

index/gn/gnat_game/gnat_game-0.1.0-dev.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0-dev"
+
+authors = ["Leo Germond"]
+maintainers = ["Leo Germond <germond@adacore.com>"]
+maintainers-logins = ["leogermond"]
+
+licenses = "Apache-2.0"
```
onox, line 9: The LICENSE file says Apache 2.0, but some of the copyright headers in files in src/ are licensed under GPL2+.

Apache 2.0 is [compatible with GPL 3](https://www.gnu.org/licenses/license-list.html#apache2) (but only in the [Apache 2.0 -> GPL3 direction](https://www.apache.org/licenses/GPL-compatibility.html)), but not with 2, so if you have Apache 2.0 and GPL2+ licensed code in your project, then your binaries would fall under the GPL3 license.

You can use SPDX expressions though and use the `AND` keyword: `Apache-2.0 AND GPL-2.0-or-later`.

index/gn/gnat_game/gnat_game-0.1.0-dev.toml
```diff
@@ -0,0 +1,15 @@
+name = "gnat_game"
```
mosteo, line 1: The name "gnat_game" is concerning (unless the game is about a gnat), as it implies some kind of association with the GNAT compiler. Could you clarify about it? (Also the PR doesn't mention this crate at all?).

index/gn/gnat_game/gnat_game-0.1.0-dev.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "Shiny new project"
+version = "0.1.0-dev"
```
mosteo, line 3: Pre-release versions (with a dash part) should be exceptional. In particular, for pre-1.0 versions, they don't make much sense IMHO, so the version should be `0.1.0` at a minimum.

index/sd/sdl/sdl-0.1.0-dev.toml
```diff
@@ -0,0 +1,18 @@
+name = "sdl"
+description = "SDL Ada bindings"
+version = "0.1.0-dev"
```
mosteo, line 3: Same observation about the `0.1.0`.

index/sd/sdl/sdl-0.1.0-dev.toml
```diff
@@ -0,0 +1,18 @@
+name = "sdl"
```
mosteo, line 1: Using a name that's the same as the official project for an unaffiliated project can generate confusion. For a thin automatic binding, I would suggest something like `sdl_thin` or `sdl_autothin`. Then again, a purely automatic binding can be generated on the spot in a post-fetch action, so I'm not sure publishing it as a crate is necessary. Are you applying some further processing/edition to the generated files?

mosteo (changes requested):

PR #629, opened 2022-09-02
Add Alire support for Florist source version.

Blady-Com:


index/fl/florist/florist-6.0.0.toml
```diff
@@ -0,0 +1,32 @@
+name = "florist"
+description = "POSIX Ada Bindings"
+version = "6.0.0"
```
mosteo: Just curious, where's this version coming from?

index/fl/florist/florist-6.0.0.toml
```diff
@@ -0,0 +1,32 @@
...
+tags = ["florist", "posix", "bindings"]
+
+authors = ["Florida State University (FSU)", "AdaCore"]
+maintainers = ["Pascal Pignard <blady.net@orange.fr>"]
+maintainers-logins = ["Blady-Com"]
```
mosteo: Given the existence of https://github.com/AdaCore/florist I wonder if we could have some input from the current maintainers about whether they plan to publish this themselves or are they ok with this fork. I'm tentatively pinging @ptroja as the last person to commit to that repo.

ptroja: I think @ArnaudCharlet is the closest person to the "current maintainer" of the Florist repository. My recent commits were merely to fix some build issues with new GNAT compilers.

index/fl/florist/florist-6.0.0.toml
Blady-Com: Florist GPL 2017 was the last published public (GPL) source code by AdaCore since the show up on their Github recently. I started with this GPL 2017 version and considered this version as version 1 and incremented it along modifications from some users. I've informed AdaCore since end 2017 / beginning 2018.

index/fl/florist/florist-6.0.0.toml
Fabien-Chouteau: Since the original authors (FSU) are not maintaining this since long ago, and we effectively have two forks now. I think @Blady-Com you should add a suffix to your crate name (e.g. `florist_blady`), and if/when AdaCore publishes on Alire we will use the crate name `florist_adacore`. What do you think?

mosteo: With this idea, as long as the versioning doesn't overlap or the forks remain compatible, both forks could provide an abstract `florist` crate so users that don't care could use either of them transparently.

Blady-Com: @Fabien-Chouteau nice idea, ok for me.
@mosteo what is an abstract crate? How does it work?

mosteo: By using a `provides = ["florist=x.x.x"]` [property](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#release-information), you're saying that your crate is API-compatible with any other crate with a similar property. This is currently used for all gnat_xxx compilers to provide `gnat`, no matter for what architecture.

In practice, someone who doesn't care about the provider would depend on `florist` instead of `florist_blady` or `florist_adacore`, and either would be used to solve `florist`. But again, this requires care with versions/compatibility.

mosteo (approved):

PR #630, opened 2022-09-03
Add Run-Time-System and MCU definitions from AVR-Ada

RREE:
Provide a run time system for the AVR gnat compiler. It assumes that avr-libc is already provided together with the compiler . This is true for gnat_avr version 11, but not for gnat_avr version 12

### DIFF
--- a/index/av/avrada_mcu/avrada_mcu-2.0.0-pre_1.toml
+++ b/index/av/avrada_mcu/avrada_mcu-2.0.0-pre_1.toml
@@ -1,0 +1,20 @@
+name = "avrada_mcu"
+description = "Device (MCU) specific definitions for AVR microcontrollers"
+version = "2.0.0-pre_1"
+
+authors = ["Rolf Ebert"]
+maintainers = ["Rolf Ebert <rolf.ebert.gcc@gmx.de>"]
+maintainers-logins = ["RREE"]
+licenses = "GPL-2.0-or-later WITH GCC-exception-3.1"
+website = "https://sourceforge.net/projects/avr-ada/"
+tags = ["avr", "embedded", "rts"]
+
+[[depends-on]]
+gnat_avr_elf = "^11"
+avrada_rts = "^2.0.0"
+
+
+[origin]
+commit = "edcb6f3f773ea28dd55da50088aa6353a612b522"
+url = "git+https://github.com/RREE/AVRAda_MCU.git"
+
--- a/index/av/avrada_mcu/avrada_mcu-2.0.0.toml
+++ b/index/av/avrada_mcu/avrada_mcu-2.0.0.toml
@@ -1,6 +1,6 @@
 name = "avrada_mcu"
 description = "Device (MCU) specific definitions for AVR microcontrollers"
-version = "2.0.0-pre_1"
+version = "2.0.0"
 
 authors = ["Rolf Ebert"]
 maintainers = ["Rolf Ebert <rolf.ebert.gcc@gmx.de>"]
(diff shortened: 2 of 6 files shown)


mosteo:
> It assumes that avr-libc is already provided together with the compiler . This is true for gnat_avr version 11, but not for gnat_avr version 12

Is this the reason why these crates depend on `gnat_avr_elf ^11`?

index/av/avrada_mcu/avrada_mcu-2.0.0-pre_1.toml
```diff
@@ -0,0 +1,20 @@
+name = "avrada_mcu"
+description = "Device (MCU) specific definitions for AVR microcontrollers"
+version = "2.0.0-pre_1"
```
mosteo: Is this pre-release version intended? Same for `avrada_rts`

RREE: This is my first real Alire submission. I am not sure what is required. That's why I called it a pre-release to be on the safe side. I can resubmit all crates as V2.0.0 if that is preferred

RREE:
> > It assumes that avr-libc is already provided together with the compiler . This is true for gnat_avr version 11, but not for gnat_avr version 12
> 
> Is this the reason why these crates depend on `gnat_avr_elf ^11`?

Yes, gnat_avr_elf v12 currently does not work. It misses avr-libc. There was a discussion on Gitter a while ago that I should assume the presence of avr-libc in the compiler [releases](https://github.com/alire-project/GNAT-FSF-builds/releases). That's why I did not bother to provide my own

index/av/avrada_mcu/avrada_mcu-2.0.0-pre_1.toml
mosteo: We don't have hard rules about this kind of thing but it makes more sense to only publish final releases to the community index. For still experimental crates usually the `0.x` versions are used, but since you went straight to `2.0` I guess there's a `1.0` around. If you don't foresee major changes, plain `2.0` would be ok and you can always release updates if something comes up.

mosteo:
It looks like there's trouble with the line terminators on windows. The simplest fix would be to not treat warnings as errors in release mode build.

RREE:
Some remarks about that error. 

- first: I don't know where `-gnatwe` is set. It is certainly not in any of the files in AVRAda_RTS/MCU. 
- second, isn't that more of a problem of git and style check on Windows? Why does the style check on a Windows machine complain about CRLF line ending? That is the usual line end marker on Windows, 
- And then third, why does git generate CRLF ending in the first place when checking out on a Windows system? If we know that the GNAT style checker gripes we should switch off autocrlf

RREE:
Or are there CRLF line endings checked in?

mosteo:
git + cross-platform line endings is a great source of headaches, if you search about it. There's several git options to tweak how check-ins/outs convert (or not) line endings. But at this time I'm puzzled too by where is warnings-as-errors enabled.

Fabien-Chouteau:
> Yes, gnat_avr_elf v12 currently does not work. It misses avr-libc. There was a discussion on Gitter a while ago that I should assume the presence of avr-libc in the compiler [releases](https://github.com/alire-project/GNAT-FSF-builds/releases). That's why I did not bother to provide my own

Well that's a mistake on my part then. I thought it was included.

RREE:
Is there anything I can do to make the checks pass on Windows?

mosteo:
> Is there anything I can do to make the checks pass on Windows?

Do you have Windows machine on which to test? Ideally the source of the warnings-as-erros should be pinpointed and removed.

RREE:
I cannot reproduce the problem on my Windows computer. I freshly installed Alire, downloaded the toolchain and checked out AVRAda_RTS and AVRAda_MCU using the shell and terminal that is provided by Alire. The checked out sources only have LF line endings. 
As far as I understand the situation, the style error stems from the `-gnatg` option that is needed for compiling the run time system. The style error cannot be suppressed by `-gnatws`.
I'll try to add a `.gitattributes` that should force the git checkout to only use LF even on a Windows machine

mosteo:
Looks good now, thanks

mosteo (approved):

PR #631, opened 2022-09-03
crate for AVR standard library

RREE:
This crate requires the prior acceptance of avrada_RTS and avrada_mcu. The working examples (avrada_examples) will come next.

### DIFF
--- a/index/av/avrada_lib/avrada_lib-2.0.0-pre_1.toml
+++ b/index/av/avrada_lib/avrada_lib-2.0.0-pre_1.toml
@@ -1,0 +1,32 @@
+name = "avrada_lib"
+description = "Library of drivers for AVR microcontrollers"
+version = "2.0.0-pre_1"
+
+authors = ["Rolf Ebert"]
+maintainers = ["Rolf Ebert <rolf.ebert.gcc@gmx.de>"]
+maintainers-logins = ["RREE"]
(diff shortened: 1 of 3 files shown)


mosteo:
There seems to be some concerning warnings [here](https://github.com/alire-project/alire-index/actions/runs/3183534038/jobs/5190868756#step:11:701), [here](https://github.com/alire-project/alire-index/actions/runs/3183534004/jobs/5190868859#step:9:727), [here](https://github.com/alire-project/alire-index/actions/runs/3183534004/jobs/5190868965#step:9:748) (probably more).

Please let me know if that's OK or not.

RREE:
Thanks for remarking the warnings. All warnings can be ignored

- the warning about the failed assertion in I2c.Master shows because the configuration variable Clock_Frequency is not yet set (to be done in the application, not in the library)
- admittedly, I don't understand the second case. I use UC between two 32bit types and still get the warning. The generated code works as expected, though.
- the third is the same as the first one
- there are several intentional warnings "pragma NOT_Implemented". That is more visible during development than an exception Program_Error that simply restarts the processor 

I'd like to get the crate accepted despite the warnings, thanks.

mosteo:
All right then, thanks for clarifying.

mosteo (approved):

PR #632, opened 2022-09-09
Add Alire support for NCursesAda source version.

Blady-Com:


Blady-Com:
Build ubuntu-latest::fedora-latest error:
configure: WARNING: Ada95 binding required program m4 not found. Ada95 binding disabled
Does it correct that m4 is missing?

mosteo:
That's right, `m4` is not installed by default in the Fedora image. Even if it's by chance in the rest, it seems the proper thing is to have an external for m4 and depend on it.

Blady-Com:
M4 has been added with  m4_tool name (name longueur than 2 is required by alr).
But the macOS build crashed with a system like error:
`rm: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp.HhMXXhRD: Resource busy`
Is there a way to run again the checks?

mosteo:
I just relaunched the checks

Blady-Com:
@mosteo thanks.
But there is a Fedora error: `sudo: pacman: command not found`
What to do?
Do I suppress Fedora as available OS?

index/nc/ncursesada/ncursesada-6.3.0.toml
```diff
@@ -0,0 +1,37 @@
...
+project-files = ["Ada95/lib_adacurses.gpr"]
+
+[available.'case(os)']
+linux   = true
+macos   = true
+windows = false
+'...'   = false
+
+[[depends-on]]
+m4_tool = "*"
```
mosteo, line 24: With the current definition in `m4_tool`, this won't be resolved on most Linuxes

index/m4/m4_tool/m4_tool-external.toml
```diff
@@ -0,0 +1,17 @@
...
+[external.origin."case(distribution)"]
+"fedora" = ["m4"]
+
+[[external]]
+kind = "version-output"
+version-command = [ "m4", "--version" ]
+version-regexp = "m4.*M4. ([\\d\\.]+)"
+[external.available.'case(os)']
+macos = true
+'...' = false
```
mosteo: I think it's safe to leave this definition available for all OSes

mosteo (changes requested):
The check failure was a transient problem from Github. I think the only fix needed now is to make m4 generally available, so checks can succeed in more Linuxes.

mosteo (approved):

PR #636, opened 2022-09-16
Add index entry for Admpfr v4.1.0-0

thvnx:
Admpfr is an Ada library offering bindings for the MPFR C library. It's a complete interface to the C library.

Admpfr v4.1.0-* is compatible with MPFR 4.1.0.


thvnx:
Hi!

This is my first contribution to Alire, I wasn't able to use `alr publish` (it failed at step 4) so I wrote the configuration file by hand.

These bindings rely on the fact that the C MPFR library version 4.1.0 is installed on the system.

index/ad/admpfr/admpfr-v4.1.0-0.toml
```diff
@@ -0,0 +1,22 @@
+name = "admpfr"
+version = "4.1.0-dev"
```
mosteo: The `-dev` makes the version a pre-release, so it should be removed.

mosteo (changes requested):

index/ad/admpfr/admpfr-v4.1.0-0.toml
mosteo: And, in any case, the file name must match the version in the manifest.

mosteo:
> This is my first contribution to Alire

Thanks for your contribution!

> I wasn't able to use `alr publish` (it failed at step 4) so I wrote the configuration file by hand.

Could you elaborate a bit about the error you experienced?

> These bindings rely on the fact that the C MPFR library version 4.1.0 is installed on the system.

In that case, you should include the dependency in this manifest, and submit also an external definition for the C library. (See e.g. https://github.com/alire-project/alire-index/blob/25e277e47debcf6075646e913212882fb3c8b14f/index/li/libgmp/libgmp-external.toml , you only need to provide the "system" kind). Externals tell Alire how to install a system package.

thvnx:
Thank you for the review!

> > I wasn't able to use `alr publish` (it failed at step 4) so I wrote the configuration file by hand.
> 
> Could you elaborate a bit about the error you experienced?

Yes, it failed at step 4 with:

```
info: Publishing assistant: step 4 of 6: Build release
detail: Spawning: ["alr", "--non-interactive", "build"]
error:    Command ["alr", "--non-interactive", "build"] exited with code 1
```

But I don't know if I can enable some debug output to see why `alr build` fails here.

mosteo:
> But I don't know if I can enable some debug output to see why `alr build` fails here.

Strange because the output of `alr build` should be dumped to the screen.

index/ad/admpfr/admpfr-v4.1.0-0.toml
```diff
@@ -0,0 +1,29 @@
...
+macos   = true
+windows = true
+'...'   = false
+
+[[external]]
+kind  = "system"
+[external.origin."case(distribution)"]
+"debian|ubuntu" = ["libmpfr-dev"]
+"arch|macos"    = ["mpfr"]
+msys2           = ["mingw-w64-x86_64-mpfr"]
```
mosteo: This external definition should go in its own file, e.g., in `mpfr-external.toml` placed in `mp/mpfr`. In turn, `admpfr` should depend on this library through a
```
[[depends-on]]
mpfr = "^4.1" # or the version you're targeting
```
. This must appear in this manifest for admpfr.

index/ad/admpfr/admpfr-v4.1.0-0.toml
```diff
@@ -0,0 +1,29 @@
+name = "admpfr"
+version = "4.1.0-0"
```
mosteo: Two issues with this: the filename must match the version exactly: `admpfr-4.1.0-0.toml`. But, the `-0` seems suspicious to me. This is making this a [pre-release](https://semver.org/#spec-item-9) before 4.1.0 according to semver specs.

mosteo (changes requested):

index/ad/admpfr/admpfr-v4.1.0-0.toml
thvnx: Hope it's better now (named it `libmpfr` for consistency)!

index/ad/admpfr/admpfr-v4.1.0-0.toml
thvnx: It uses build metadata now, which is simply ignored according to the documentation.

mosteo: I'm afraid not, build metadata is introduced with `+`: `4.1.0+0`

> [Build metadata MAY be denoted by appending a plus sign](https://semver.org/#spec-item-10)

index/ad/admpfr/admpfr-v4.1.0-0.toml
mosteo: Yes, this looks good now, thanks!

thvnx:
Hello,

It fails on ubuntu-lts because `gnat2022` is not supported by gnat:
```
 Note: Building admpfr/admpfr.gpr...
Setup
   [mkdir]        object directory for project Admpfr
   [mkdir]        library directory for project Admpfr
Compile
   [C]            admpfr_stubs.c
   [Ada]          admpfr-custom_bindings.ads
```

Should I add a dependency on gnat_native 12 in the toml file?

mosteo:
> Should I add a dependency on gnat_native 12 in the toml file?

Yes, but not on `gnat_native` but on plain `gnat`. Assuming GNATs will be forward-compatible, you can rather add `gnat = ">=12"`.

thvnx:
One last thing I guess, how can I fix the gnat dependency on windows?

```
Dependencies (direct):
   gnat>=12
   libmpfr^4.1.0
Dependencies (solution):
   gnat=2020.0.0 (gnat_external) (installed) (origin: external)
   libmpfr=4.1.0 (origin: system)
Dependencies (graph):
   admpfr=4.1.0+20220920 --> gnat=2020.0.0 (gnat_external) (>=12)
   admpfr=4.1.0+20220920 --> libmpfr=4.1.0 (^4.1.0)              
```

EDIT: actually, it doesn't seem to work on other platforms too. As for example on ubuntu: `SKIPPING build for crate admpfr=4.1.0+20220920 with MISSING external dependencies`.

mosteo:
Ah, right. Since the discontinued Community Editions we use started in 2019, you can do:
 
`gnat = ">=12 & <2000"`

Kind of our little 2k problem.

mosteo (approved):

mosteo:
Thanks @thvnx and congrats on your first contribution

thvnx:
Thank you for the help! Keep going with Alire, it looks very promising!

I just have one last remark, I'm not sure build checks are actually OK:
```
Dependencies (graph):
   admpfr=4.1.0+20220920 --> gnat>=12 & <2000      
   admpfr=4.1.0+20220920 --> libmpfr=4.1.0 (^4.1.0)
Dependencies cannot be met
SKIPPING build for crate admpfr=4.1.0+20220920 with MISSING external dependencies
```
Is this expected?

mosteo:
Yes, it is for that particular Windows image as there's no suitable compiler available.

thvnx:
> Yes, it is for that particular Windows image as there's no suitable compiler available.

Not sure to understand, I can use gnat_native 12 on windows thanks to Alire on my side.

Moreover, I get the same result for all linux and macos images. Along with this message for the `libmpfr` part: ` SKIPPING undetected external crate libmpfr`.

That's just to be sure the crate definitions are OK. I tested it on my side with `alr get admpfr`, everything seems to work correctly though.

mosteo:
Yes, we have a Windows VM that checks using the Alire toolchains, and that one works as expected. There's however another one that builds using forcibly the latest Community Edition, because at the time it was the only "common" compiler to all platforms. This one is to be phased out now that it will be no longer supported.

As for other combos, as long as there is an external detector and GNAT 12 available, they pass; otherwise the build is skipped because one or the other is missing in the platform.

PR #644, opened 2022-10-02
Add Support for Fedora

JulienGioux:
Nomenclature of package name is different on Fedora, certainly the case for all RPM based distribution...

### DIFF
--- a/index/li/libpq/libpq-external.toml
+++ b/index/li/libpq/libpq-external.toml
@@ -8,5 +8,5 @@ maintainers-logins = ["reznikmm"]
 kind = "system"
 [external.origin."case(distribution)"]
 "fedora" = ["libqp-devel"]
-"debian|ubuntu" = ["libpq-dev"]
+"debian|ubuntu" = ["libqp-dev"]
 "msys2" = ["mingw-w64-x86_64-postgresql"]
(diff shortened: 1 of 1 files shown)


index/li/libpq/libpq-external.toml
```diff
@@ -7,5 +7,6 @@ maintainers-logins = ["reznikmm"]
 [[external]]
 kind = "system"
 [external.origin."case(distribution)"]
+"fedora" = ["libqp-devel"]
```
mosteo: Should this be `libpq-devel` instead of `libqp-devel`?

mosteo (changes requested):

JulienGioux:
Oooops... sorry !!!
You're right !
*Julien Gioux*
***@***.*** ***@***.***>*

index/li/libpq/libpq-external.toml
JulienGioux: You're right, i need to sleep i think ^^
dyslexic forever :/
See here :
[https://koji.fedoraproject.org/koji/buildinfo?buildID=1877923](https://koji.fedoraproject.org/koji/buildinfo?buildID=1877923)

mosteo:
Looks good now, thanks.

mosteo (approved):

PR #646, opened 2022-10-03
Add PowerJoular

adelnoureddine:
Add PowerJoular, a Linux tool to monitor the power consumption of multiple platforms and processes (x86, ARM RPi, etc.)


mosteo (dismissed):
Thanks for the updated PR, @adelnoureddine. The only thing I miss is tags, in case you wanted to make your crate easier to locate/classify. There is no need to update the commit or the sources, the tags can be added directly to the manifest in this PR.

See [here](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#release-information) for information on the tags property.

If you don't plan on adding them, just let me know so I can go ahead with the merge.

adelnoureddine:
Thanks. I added the tags to the toml file.

mosteo (approved):

PR #649, opened 2022-10-03
add support for fedora libmysqlclient & libgtk3

JulienGioux:
add support for fedora libmysqlclient & libgtk3


mosteo:
The problem here seems unrelated to your changes, @JulienGioux . I'll take a look at what may be happening.

JulienGioux:
> The problem here seems unrelated to your changes, @JulienGioux . I'll take a look at what may be happening.
Not sure, but the problem appeared when instead of Sync my fork i have merge and pull request. :/

mosteo:
> Not sure, but the problem appeared when instead of Sync my fork i have merge and pull request. :/

In any case you haven't touched anything about Arch so it must come from elsewhere. If I manage to test Fedora only I will merge already.

PR #651, opened 2022-10-05
Create task_coroutines-0.1.0.toml

Fabien-Chouteau:


mosteo (commented):
Neat, I wanted to take a stab at this in the past but I see I'm doubly late...

mosteo:
Mmm, don't know why checks aren't getting triggered...

Fabien-Chouteau:
I tried to re-open the PR... Looks like something's wrong on GitHub side.

mosteo:
We have a problem with the diagnostic message, but the root cause is that the new manifest should be filed under an extra folder `task_coroutines` ^_^'

PR #652, opened 2022-10-05
Add dependency on gnat 11 for libadalang_tools

dkm:
libadalang_tools crate does not build correctly with GNAT 12 because of new diagnostic message.


mosteo:
Arg, this is failing because the external for macOS is reporting gmp as installed but it's not really configured to be ready to use. This affects all crates in macOS. I'm going to merge as your change is correct. Still, this has to be fixed elsewhere...

mosteo (approved):

dkm:
Thanks for merging this !

Fabien-Chouteau:
There would be another solution which is to set `BUILD_MOD` to `prod`, but it might bring some incompatibility.

PR #656, opened 2022-10-08
Add Alire support to SparForte version 2.5.0.

Blady-Com:


index/sp/sparforte/sparforte-2.5.0.toml
```diff
@@ -0,0 +1,27 @@
+name = "sparforte"
+description = "An open source scripting lanquage for high integrity tasks"
```
mosteo: s/lanquaje/language

mosteo (changes requested):

mosteo:
Besides the typo, it seems there's missing dependency on `locate` external and some trouble with `sh`

mosteo:
I'll look into how to configure the GHA environment to use a `sh` that's compatible. The dependency on `locate` should be fixed in the crate though (and a external for it will be needed).

mosteo:
I don't think the current error from `configure`  at https://github.com/alire-project/alire-index/actions/runs/4286705968/jobs/7484303863#step:7:450 is our fault. Maybe there's some other package missing?

Blady-Com:
> I don't think the current error from `configure` at https://github.com/alire-project/alire-index/actions/runs/4286705968/jobs/7484303863#step:7:450 is our fault. Maybe there's some other package missing?

In fact for Linux, `configure` expects an actual ARCH value as `x86_64` for `--arch` option.
As for macOS it isn't needed, I put `native` in last option:
```
command = ["sh", "configure", "--without-bdb", "--without-mysql", "--without-opengl",
                              "--without-postgres", "--without-readline", "--without-sound",
                              "--without-pcre", "--without-sdl", "--arch=native"]
```
Is there a mean to set the relevant ARCH regarding the OS?

mosteo:
You could have conditional actions based on `os` and `word-size` to supply the value actually expected by configure.

See the entry on `actions` at https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#release-information, and on parameters at https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#parameters

You can use several conditions, e.g.,
`'case(os)'.linux.'case(word-size)'.bits-64 = ...`

Blady-Com:
Finally I suppress the `--arch` option that let configure to guess the ARCH (but it doesn't well for Mac: i386 instead of x86-64 thus I changed configure too).
But I don't understand why it gets `Could not run locate` in macOS log whereas it is ok on my my Mac.

Blady-Com:
In case of Ubuntu, it logs:
```
checking for tput... no
tput is sometimes included with the ncurses library (e.g. Cygwin).

```
Does not ncurses present by default in Ubuntu?
Else shall I create a ncurses dependance?

mosteo:
> Does not ncurses present by default in Ubuntu?
> Else shall I create a ncurses dependance?

The images provided by GitHub are quite bare. If you do depend on ncurses, yes, it should be made explicit through a dependency.

mosteo:
I need to check what's going on with this new error

index/nc/ncurses/ncurses-external.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "ncurses"
+authors = ["Thomas E. Dickey", "Juergen Pfeifer", "Eugene V Melaragno", "Nicolas Boulenguez"]
+
+maintainers = ["Pascal Pignard <blady.net@orange.fr>"]
+maintainers-logins = ["Blady-Com"]
+
+[[external]]
+kind = "system"
+[external.origin."case(distribution)"]
+"debian|ubuntu" = ["libncurses5-dev", "libncursesw5-dev"]
```
mosteo: The trouble is here, both packages are providing the same release. Also, both are transitional, so just `libncurses-dev` package should appear here.

mosteo (changes requested):

mosteo:
The new error at https://github.com/alire-project/alire-index/actions/runs/4419887013/jobs/7748891025#step:7:963 is puzzling because I just checked and `tput` is part of `ncurses-bin`, which is a dependence of `libncurses-dev`. Any ideas?

JeremyGrosser:
`tput` is supplied by the `ncurses-bin` package.
https://packages.debian.org/bullseye/amd64/ncurses-bin/filelist

PR #657, opened 2022-10-08
caught up with all new RP2040 changes required

hgrodriguez:
hal 0.3.0; rp2040 2.0.0 etc.


hgrodriguez:
found a missing entry so will re-raise

PR #658, opened 2022-10-08
caught up with newest changes for HAL/RP2040

hgrodriguez:
including flash chip selection and linker switches


mosteo (dismissed):
Hi, @hgrodriguez. I'm not sure why both `0.2.0` and `0.2.1` are being published? Is there any reason why `0.2.0` can't be skipped entirely?

hgrodriguez:
@mosteo : you are right, but i was not sure how to skip 0.2.0 re-using my branch. I am more than happy to learn, else I will *always* delete my fork, and then re-raise the pull request so your world is clean. more than happy to make your life easier.

mosteo:
It's enough to delete the file from the branch you're using to push the PR. I think in this case I can do it directly from the github interface so no need for you to worry.

mosteo (approved):

hgrodriguez:
¡Muchas gracias!

PR #659, opened 2022-10-08
first Alire release; required for Emacs ada-mode

stephe-ada-guru:


stephe-ada-guru:
This is my very first github pull request, as well as my first alire pull request. So I've got some things wrong. There are still a couple of crates to add; I'll post here again when everything is ready.

mosteo:
Feel free to tinker, I won't review until you request it via the Reviewers (top-right) option. Also don't hesitate to ask for guidance as needed.

stephe-ada-guru:
I believe this is now ready. According to github help, I should be able to click on "Reviewers" and select a reviewer, but that doesn't work (in firefox on debian); nothing happens when I click there.

Please allow the build tests to run before you review; there's probably something that needs fixing.

index/em/emacs_ada_mode/emacs_ada_mode-8.0.0.toml
```diff
@@ -0,0 +1,35 @@
...
+
+authors = ["Stephen Leake"]
+maintainers = ["Stephen Leake <stephen_leake@stephe-leake.org>"]
+maintainers-logins = ["stephe-ada-guru"]
+
+website = "https://www.nongnu.org/ada-mode/"
+
+project-files = ["emacs_ada_mode.gpr"]
+
+executables = ["ada_mode_wisi_lalr_parse.ads",
```
Fabien-Chouteau: Executable should not be source files, but actual executable name (without `.exe` extension).

Fabien-Chouteau (changes requested):

Fabien-Chouteau:
> I believe this is now ready. According to github help, I should be able to click on "Reviewers" and select a reviewer,

Only repo maintainers can do that.

index/em/emacs_ada_mode/emacs_ada_mode-8.0.0.toml
stephe-ada-guru: Ah. I assumed it was the same as the gpr setting. Other packages also have executables. 

Can I just edit the toml file in the index, or do I have to regenerate the tarball? I'm not clear on what's included in the hash.

stephe-ada-guru:
I assume I can't do anything about the canceled checks.

For the failing check on ubuntu, the error is:
DEPLOYING CRATE emacs_wisi=4.1.0-beta
stderr: Generating Outcome_Failure with message: release integrity test failed: expected [sha256:fb9fa8d4ef7402479830da0a9c49b2a1cbba5f65a2c30055d77a0e8eeca4180a] but got [sha256:d217716d4d6a68b2f79fb63c62b6cf1aa519a9f10447d1e1ca4f27cdbe1c976d]

When I run sha256sum on the .tgz file given by the url in emacs_wisi-4.1.0-beta.toml, it gives a hash that is neither of the ones in the error message, but it does match the one in the .toml file. So I don't understand the problem.

stephe-ada-guru:
I've verified that the sha256 of emacs_wisi-4.1.0-beta.tgz is the same in four places; the original archive in the alire workspace, savannah, the hashes field of emacs_wisi-4.1.0-beta.toml in my local checkout of my alire-index branch, and in the pull request.

stephe-ada-guru:
I figured out the problem, by searching thru alire-index for matching hashes. The expected hash is from stephes_ada_library-3.7.0.toml, and the found hash is from stephes_ada_library-3.7.0.tgz. The error message could be improved here.

So I need to redo that upload. Now I'm just waiting for savannah again.

mosteo:
> The error message could be improved here.

So the wrong crate is being reported?

stephe-ada-guru:
The log says "DEPLOYING CRATE emacs_wisi=4.1.0-beta", and then gives an error message about mismatched hash for the crate stephes_ada_library. emacs_wisi depends on stephes_ada_library, so it makes sense that it is being unpacked, but it is not clear from the error message which crate has the wrong hash.

stephe-ada-guru:
I committed a fix for the hash issue, and for the executables names; please trigger a build test.

mosteo:
> is not clear from the error message which crate has the wrong hash.

Yeah, then it's clearly in need of improvement.

mosteo:
> Please allow the build tests to run before you review;

This happens only for first-time contributors, imposed by github, I guess because it may potentially incur expenses. You won't have to endure it in future PRs.

mosteo:
This error

> gnat1: invalid switch: -gnat2022

tells me that you have to add dependency restrictions on gnat versions. I'm not sure which GNAT introduced that switch, but assuming it was 11 for the sake of example, it would be smthing like:

`gnat = ">=11 & <2000"`

The <2000 is because of the Community Edition versions, which use their year. (Incidentally, @Fabien-Chouteau , I've been thinking that maybe we could have the CE versions as separate crates that `provide` their "mainline" GCC version to avoid this confusing restriction.)

stephe-ada-guru:
> This error
> 
> > gnat1: invalid switch: -gnat2022

I don't see that in the test logs; where do you see it?

I do have that switch in my project files, so I should add that restriction.

mosteo:
In the Ubuntu LTS test, here: https://github.com/alire-project/alire-index/actions/runs/3252292865/jobs/5343881593#step:9:2103

stephe-ada-guru:
I pushed new versions to fix the problems (ie emacs_ada_mode-8.0.2), but it's compiling the old versions (emacs_ada_mode-8.0.0). Do I need to cancel this pull request and start another?

mosteo:
> Do I need to cancel this pull request and start another?

No; if you remove the old manifests it should be enough. The tests will go over any modified manifest, not just the latest ones.

stephe-ada-guru:
I fixed one problem on Windows, but now it's failing again. I don't understand the error. I'm trying to run 'make' in the 'wisitoken' crate build directory, so I use WISITOKEN_ALIRE_PREFIX. But then it complains that directory does not exist! What's going on?

This worked when I ran 'alr build' on Windows with pins in emacs_ada_mode.toml.

mosteo:
Inspecting the log it seems as if the pre-build action were being triggered even before retrieving dependencies, which would be very wrong. I will need some time to check if this is indeed a bug on Alire's side.

In the meantime, I suggest switching `pre-build` to `post-fetch` and see if that improves the situation (it may be necessary to do so in all crates using `pre-build`.

stephe-ada-guru:
I did "alr get --build emacs_ada_mode" on Windows, and it worked. So it must be something specific to the way the tests on github are run. Apparently it's running gh-build-crate.sh; where is that file? I'd like to run it on my Windows system, to see if I can reproduce the failure.

In addition, there are earlier errors in the Windows build, about not being able to delete a file, and not finding the "local" index.

mosteo:
The script that runs the tests is at https://github.com/alire-project/alire-index-checks/blob/stable-1.2.1/scripts/gh-build-crate.sh

The index "error" is expected I think but I should remove the cause as it may be misleading.

stephe-ada-guru:
So gh-build-crate.sh runs "alr get <crate>" and "alr build <crate>"; nothing special. So I don't know why it's failing in the github environment.

the reason I need to run this particular make command is to build an executable that the wisitoken crate should provide. It doesn't actually provide it, because of a limitation in gpr files; a library gpr file can't specify a main executable, and a normal gpr file that does specify an executable can't 'with' a library gpr file (it can only 'with' the installed library gpr file (hmm - maybe I can include both in the wisitoken source?)). An alternative solution to trying to run 'make' is to provide a separate crate wisitoken_exe that just provides the executable. Would that be preferable?

mosteo:
Taking another look at the log I see that the action is launching:

`C:/ProgramData/Chocolatey/lib/mingw/tools/install/mingw64/bin/mingw32-make.exe`

which seems not to be the mingw installed by Alire (as ours comes directly from its installer and not Chocolatey). Not sure if that could have any influence on the path resolving differently (but the path seems to be for an "out-of-mingw" location so it should work?)

The humongous PATH displayed [just before](https://github.com/alire-project/alire-index/actions/runs/3285101118/jobs/5424885639#step:9:450) has paths into mingw near the beginning and near the end, and for Chocolatey bin in the middle. I would need to check from where everything is coming from, in particular where Alire places its mingw path.

>  provide a separate crate wisitoken_exe that just provides the executable

I know some crates have gone this way. This is more natural if the executable is not always needed by the users of the library from which it would be segregated.

stephe-ada-guru:
I found a way to include the executable in the wisitoken crate; working on submitting that (savannah is slow :().

stephe-ada-guru:
Huzzah! All tests passed. Please do final review, and hopefully merge.

mosteo (approved):

mosteo:
It looks good on my side. I see that @Fabien-Chouteau requested changes so I leave up to him to take the final look and merge. I'm sorry it was a somewhat painful process but I think you have a particularly complex case... Let's hope for smoother next times. It helped us to identify points where to improve, though, so thanks for persevering.

index/em/emacs_ada_mode/emacs_ada_mode-8.0.0.toml
mosteo: I'm seeing this only now, but the authoritative source for your release is the manifest you submit to the index, so fixing that's enough in principle. The hash covers only the companion tarball. 

The bundled `alire.toml` with your sources only gets used in case someone directly pins to your repository or sources, so it's worth having it in sync, although it's less critical if the data used to configure the build is OK.

Fabien-Chouteau (approved):

PR #660, opened 2022-10-09
feat: add vaton 0.1.0-dev to alire

mhatzl:
Hello,
please add *vaton-0.1.0-dev* to Alire

Thanks!

### DIFF
--- a/index/va/vaton/vaton-0.1.0-dev.toml
+++ b/index/va/vaton/vaton-0.1.0-dev.toml
@@ -1,0 +1,37 @@
+name = "vaton"
+description = "Verified Ascii To Number conversion written in Ada/SPARK"
+version = "0.1.0-dev"
+
+long-description = """
+This library offers formally verified functions to convert character streams into the smallest standard type representation the resulting number may fit in.
+The allowed formats are based on the [JSON-Number format](https://www.json.org/json-en.html), with the addition to allow single underscores between digits.
+
+**Note:** Only decimal based numbers are supported!
+
+**Examples:**
+
+```
+-10_000 -> Standard.Integer
+1.0 -> Standard.Float
+1e4 -> Standard.Float
+```
+"""
+
+authors = ["Manuel Hatzl"]
+maintainers = ["Manuel Hatzl <hatzlmanuel@outlook.com>"]
+maintainers-logins = ["mhatzl"]
+tags = ["spark"]
+website = "https://github.com/mhatzl/vaton"
+licenses = "MIT"
+
+[[depends-on]]
+spark_unbound = "~0.2.1"
+[[depends-on]]
+# Needed for Ada.Numerics.Big_Numbers
+gnat = "(>=9.3.1 & <2000) | >=2021"
+gnatprove = "^12.1.1"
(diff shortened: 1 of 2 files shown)


mhatzl:
not sure why ubuntu-latest fails to build. Has there been a change to Ada.Numerics?

My GitHub action on ubuntu-latest worked.
Thanks!

index/va/vaton/vaton-0.1.0-dev.toml
```diff
@@ -0,0 +1,37 @@
...
+maintainers-logins = ["mhatzl"]
+tags = ["spark"]
+website = "https://github.com/mhatzl/vaton"
+licenses = "MIT"
+
+[[depends-on]]
+spark_unbound = "~0.2.1"
+[[depends-on]]
+# Needed for Ada.Numerics.Big_Numbers
+gnat = "(>=9.3.1 & <2000) | >=2021"
```
mosteo: One possibility for the numerics failure is that GNAT 10 were to old. You could try bumping the minimum version to 11.

index/va/vaton/vaton-0.1.0-dev.toml
```diff
@@ -0,0 +1,37 @@
...
+gnatprove = "^12.1.1"
```
mosteo: If gnatprove is used to verify the library, but is not needed by users of the library, the proving should be moved to a nested crate to avoid pushing this heavy dependency into users.

index/va/vaton/vaton-0.1.0-dev.toml
```diff
@@ -0,0 +1,37 @@
+name = "vaton"
+description = "Verified Ascii To Number conversion written in Ada/SPARK"
+version = "0.1.0-dev"
```
mosteo: In general, for submissions to the public index, we prefer that pre-release versions (the `-dev` part) are not used unless there's a strong reason. If this was by unawareness, simply remove it.

mosteo (changes requested):

index/va/vaton/vaton-0.1.0-dev.toml
mhatzl: have removed the "-dev" tag. thanks for the hint

index/va/vaton/vaton-0.1.0-dev.toml
mhatzl: I moved it to the tests sub-crate

mhatzl:
Ok good. most checks are passing. The first one fails, because I added three compile switches to `alire.toml`.
Do I just remove them from the index one?

The Ada2022 switch I added because I use Ada2022 features.
The style-checks I disabled because I do not like the enforced spacing errors, and I did not know how to selectively ignore them.
I also had to change the compiler checks for validation, because an instantiation warning of spark_unbound would prevent the build, but I do not care about warnings from other crates.

Hope this is ok

mosteo (approved):

mosteo:
Looks good now. The failing test is because the js TOML parser is somehow behind the times, nothing to worry about.

PR #663, opened 2022-10-11
Add tash 8.7.0 (first Alire release).

simonjwright:
Previously known as tcladashell.

### DIFF
--- a/index/ta/tash/tash-8.7.0.toml
+++ b/index/ta/tash/tash-8.7.0.toml
@@ -41,6 +41,6 @@ type = "post-fetch"
 command = ["make", "-C", "src", "tcl_record_sizes.ads"]
 
 [origin]
-commit = "17c7742bd420cbe773fc35cd26248685167a0088"
+commit = "e083b4f29b5df3388426145bbdd5e54fda74f743"
 url = "git+https://github.com/simonjwright/tcladashell.git"
 
--- a/index/ta/tash/tash-8.7.0.toml
+++ b/index/ta/tash/tash-8.7.0.toml
@@ -1,0 +1,46 @@
+name = "tash"
+description = "Binding to Tcl/Tk"
+long-description = """
+Tash (previously known as Tcl Ada Shell) is an Ada binding to Tcl/Tk.
+
+Its purpose is to
+
+* allow a Tcl program to use Ada in place of C to implement Tcl
+  commands where additional execution speed, more complex data
+  structures, or better name space management is needed, and
+
+* support the rapid development of Platform-Independent Graphical User
+  Interfaces via Tk.
+"""
+website = "https://github.com/simonjwright/tcladashell"
+version = "8.7.0"
+licenses = "GPL-2.0-or-later WITH GCC-exception-2.0"
+
+authors = ["Simon Wright"]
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Hi @simonjwright 
```
ERROR: A post-fetch action failed, re-run with -vv -d for details
, stderr debconf: delaying package configuration, since apt-utils is not installed
debconf: delaying package configuration, since apt-utils is not installed
/bin/sh: 1: tclsh: not found
make: *** [makefile:29: tcl_record_sizes.c] Error 127
stderr: Command ["make", "-C", "src", "tcl_record_sizes.ads"] exited with code 2
```

Does this need another external dependency?

simonjwright:
I’m _pretty_ sure that the issue is that Debian, at any rate in _bullseye_, doesn’t have `tclsh` but only `tclsh8.6` (`tclsh` is usually a symlink to this). `ubuntu-latest` doesn’t have a problem.

I’ll try that, though long-term there will be an issue when Tcl/Tk gets updated beyond 8.6 -- ideally sorted using `pkg-config`, but Homebrew has a problem with that -- is there an Alire way of setting an environment variable to the result of running a script?

simonjwright:
I was right about the tclsh[8.6] issue!

The Toolchain/windows-latest issue looks to be unrelated to this PR? the Build Crate/windows-latest was successful.

simonjwright:
@Fabien-Chouteau I want to request a review, but don’t see how to? The _Reviewers_ section at top right isn’t live

mosteo:
> @Fabien-Chouteau I want to request a review, but don’t see how to? The _Reviewers_ section at top right isn’t live

That's bizarre. Not sure what can be going on.

I've relaunched the Windows test in case it was transitory glitch.

mosteo (approved):

mosteo:
> ideally sorted using `pkg-config`, but Homebrew has a problem with that -- is there an Alire way of setting an environment variable to the result of running a script?

Not currently, but I'm also aware of this issue. We may need to add this capability to our external definitions.

> I've relaunched the Windows test in case it was transitory glitch.

And it was, so merging now.

simonjwright:
That seems to have worked! Good.

Don’t understand the difference between a 'toolchain' and a 'build crate' run, the log I just looked at went ahead and built the crate?

Could I have requested the re-run?

Anyway, :relaxes:

mosteo:
> Don’t understand the difference between a 'toolchain' and a 'build crate' run

This is just the naming of the workflows, that could be improved, but the essence is that a 'toolchain' build will use the compilers indexed by Alire, whereas the 'build crate' will use the native compilers of each distro/platform if defined, or the latest GNAT CE otherwise.

> Could I have requested the re-run?

I'm not sure about that. If the option is available to you, you'll see at the top of the log of the run: 

image

Fabien-Chouteau:
> > I've relaunched the Windows test in case it was transitory glitch.
> 
> And it was, so merging now.

Looks like our integrity check detected a bad download
```
ERROR: release integrity test failed: expected [sha256:3126cd6fe82b9428ba1bd50f7d7a5e108668a41da445184a6bbe13d2543fd2d0] but got [sha256:163df270842365bdceebcedcc53dad9b950d11e3463b4e461060866353acdd3d]
```

PR #667, opened 2022-10-26
Add VSS 23.0.0 release

reznikmm:


Fabien-Chouteau:
Seems like the failure on macOS is because of GNAT CE 2020:
```
vss-string_vectors.ads:32:08: "Aggregate" is not a valid aspect identifier
vss-string_vectors.ads:32:08: "Aggregate" is not a valid aspect identifier
```

You can specify `gnat < 2000`...

godunko (dismissed):

reznikmm:
> You can specify `gnat < 2000`...

Thank you, @Fabien-Chouteau. That is what I meant, but forgot to write indeed.

Fabien-Chouteau (approved):

PR #669, opened 2022-10-26
Add gtkada-23.0.0

reznikmm:


reznikmm:
Ubuntu CI fails because Utuntu 20.04 has an old Gtk3+:

> Requested 'gtk+-3.0 >= 3.24.24' but version of GTK+ is 3.24.20
> checking for GTK+ - version >= 3.24.24... no

I have tested it on Ubuntu 22.04 and it works fine.

index/gt/gtkada/gtkada-23.0.0.toml
```diff
@@ -0,0 +1,41 @@
...
+
+# Generate gtkada-intl.adb to be able to build with gtkada.gpr
+[[actions]]
+type = "post-fetch"
+command = ["make", "src/gtkada-intl.adb"]
+
+[[depends-on]]
+make = "*"
+pkg_config = "*"
+libgtk3 = "^3"
```
Fabien-Chouteau: You can specify a dependency on libgtk3 `3.24.24`.

reznikmm: Indeed! This works, thanks!

Fabien-Chouteau (approved):

PR #670, opened 2022-10-26
Add gtkada-22.0.0

reznikmm:


reznikmm:
My idea in this PR is to avoid `make install` to let the user change `LIBRARY_TYPE` on the fly.

Fabien-Chouteau (approved):
Nice!

PR #671, opened 2022-10-27
Disable pkg-config-detected externals

mosteo:
Proper use of these packages would require calling `pkg-config` to set up environment, which Alire does not do (yet?). As is, now Alire thinks they're available but clients don't find them at compilation time.

@simonjwright , @Blady-Com does your understanding of how `pkg-config` works on macOS match my interpretation?

Edit: rather that removing these externals, we keep them unavailable in case we support `pkg-config` in the future.

### DIFF
--- a/index/li/libgmp/libgmp-external.toml
+++ b/index/li/libgmp/libgmp-external.toml
@@ -15,3 +15,7 @@ msys2 = ["mingw-w64-x86_64-gmp"]
 kind = "version-output"
 version-command = [ "pkg-config", "--modversion", "gmp" ]
 version-regexp = "([\\d\\.]+)"
+
+# pkg-config requires further steps to make packages findable by
+# dependencies that Alire does not yet support
+available = false
--- a/index/li/libgnutls/libgnutls-external.toml
+++ b/index/li/libgnutls/libgnutls-external.toml
@@ -14,3 +14,7 @@ kind = "system"
 kind = "version-output"
 version-command = [ "pkg-config", "--modversion", "gnutls" ]
 version-regexp = "([\\d\\.]+)"
+
+# pkg-config requires further steps to make packages findable by
+# dependencies that Alire does not yet support
+available = false
--- a/index/li/libgtk3/libgtk3-external.toml
+++ b/index/li/libgtk3/libgtk3-external.toml
@@ -16,3 +16,6 @@ msys2 = ["mingw-w64-x86_64-gtk3"]
 kind = "version-output"
 version-command = [ "pkg-config", "--modversion", "gtk+-3.0" ]
 version-regexp = "([\\d\\.]+)"
+# pkg-config requires further steps to make packages findable by
+# dependencies that Alire does not yet support
+available = false


Fabien-Chouteau (approved):

simonjwright:
If you have `pkg-config` it will work like it does on other platforms! but there is sometimes an issue with what the package is called.

For instance, Homebrew supports a package called `gtk3+-3.0`, and if you want to use it with gprbuild it would (I’m pretty sure) look like
```
gprbuild -Pwhatever \
   -cargs:c $(pkg-config --cflags gtk+-3.0) \
   -largs $(pkg-config --libs gtk+-3.0)
```
which seems quite a lot to ask of Alire.

That said, what does `available = false` do? I just tried with Homebrew-installed GMP and Alire’s GCC 12.2.1 and `alr update` happily used `pkg-config` to find the version (6.2.1) and went ahead to do the build.

Fabien-Chouteau:
> For instance, Homebrew supports a package called `gtk3+-3.0`, and if you want to use it with gprbuild it would (I’m pretty sure) look like
> 
> ```
> gprbuild -Pwhatever \
>    -cargs:c $(pkg-config --cflags gtk+-3.0) \
>    -largs $(pkg-config --libs gtk+-3.0)
> ```
Maybe that's what Alire should do...

Cabal for haskel has support for "pkg-config dependencies": https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-pkgconfig-depends

I don't think this is the right way to go for Alire but having a better use of pkg-config in the external crates seems like a good idea.

We can either:
 - 1. Parse the output of `pkg-config --cflags-only-I --libs-only-L` to generate the corresponding `C_INCLUDE_PATH` and `LIBRARY_PATH`
 - 2. pass the output of `pkg-config` to gprbuild like you said Simon

In terms of interface in the manifest, we might want to de-correlate the detection of libs and their version with the installation from the system package manager (when there is one).

We may introduce a "pkg-config" field in the external table:
```
description = "The GNU Multiple Precision Arithmetic Library"
name = "libgmp"
licenses = "LGPL-3.0-only"
maintainers = ["derodat@adacore.com", "chouteau@adacore.com"]
maintainers-logins = ["pmderodat", "Fabien-Chouteau"]

[[external]]
kind = "system"
pkg-config = "gmp"
"debian|ubuntu" = ["libgmp-dev"]
arch = ["gmp"]
msys2 = ["mingw-w64-x86_64-gmp"]
```
This would:
 - Use `pkg-config --modversion gmp` to find if the lib is installed or not and which version
 - Use the system package mange to install if it is not installed
 - Get the required cflags and libs from `pkg-config --cflags-only-I --libs-only-L`

CC @mosteo @reznikmm

mosteo:
My initial thoughts went in the direction of doing something general, not pkg-config specific like:

```
[[external]]
# Usual means of installation/detection here

[[external.config]] # As many as needed
variable = "LD_FLAGS"
command = ["pkg-config", "--libs", "gmp"]
```

Maybe we could have alternatives like

```
[[external.config]]
gprbuild_option = "-largs"
command = # as before
```

Although if it's only ever going to be useful with pkg-config we might go for something simpler.

Fabien-Chouteau:
Interesting, I like it.

On the other hand, `pkg-config` is more or less ubiquitous so I would not mind something less generic. But this is simple enough...

Blady-Com:
@Fabien-Chouteau thanks for Cabal pointer :-)
@mosteo your syntax has the merit of addressing the 3 cases: embedded installation, external installation addressed with environment variables and external installation addressed with gprbuild arguments.
Environment variables will be useful for those crates which need configure script for instance.
Moreover, pkg-config is not the only tool, we may address python-config and so on.

On macOS many tools managers are available. Everyone has its one preference. My thinking is rather to let the end user to make the choice.
Therefore, the end user will choose the desired manager by defining the PATH variable to the corresponding pkg-config location for instance.

One question, will it be possible to write an external config specified only for an OS and an other external config for an other OS?

A suggestion:
```
[[external]] # As many as needed
kind = "variable"
variable = "LD_FLAGS"
command = ["pkg-config", "--libs", "gmp"]

[[external]]
kind = "gprbuild_option"
gprbuild_option = "-largs"
command = # as before
```

mosteo:
> If you have `pkg-config` it will work like it does on other platforms!

See, this is the thing I'm not 100% sure: from the CI runs I've seen, it seems that on Linux libs work as-is (C headers are found) but on macOS one should use the output of `pkg-config` or they aren't.

mosteo:
> On macOS many tools managers are available. Everyone has its one preference. My thinking is rather to let the end user to make the choice.

There's nothing precluding this at present; when we added Homebrew we left the door open for others to be used. Now, it's true that for Linux we are looking only for the native one...

Fabien-Chouteau:
> macOS one should use the output of `pkg-config` or they aren't.

On msys2 as well.

PR #676, opened 2022-10-31
Add gnatdoc (as a library, for now).

reznikmm:


index/gn/gnatdoc/gnatdoc-23.0.0.toml
```diff
@@ -0,0 +1,25 @@
+name = "gnatdoc"
```
Fabien-Chouteau: Maybe call this crate `libgnatdoc` if this contains only the lib?

reznikmm: And create `gnatdoc` crate when the binary will be ready? Ok, if you think that this is better.

Fabien-Chouteau: Yeah if you think it's good

Fabien-Chouteau (approved):

PR #677, opened 2022-10-31
AWS 23.0.0

Fabien-Chouteau:


reznikmm (commented):
I would propose to avoid `make build install` and use project files to build. So users could enable SSL if needed

Fabien-Chouteau:
If that works I am all for it.

reznikmm:
> If that works I am all for it.

I'm trying [this](https://github.com/alire-project/alire-index/pull/687)

What is `gcc -dumpmachine` on Mac OS X?

reznikmm:
can I select `gpr-set-externals` by compiler version? GNAT CE 2021 and alire GCC uses `TARGET=x86_64-pc-linux-gnu`, but system GCC uses `x86_64-linux-gnu`.

mosteo:
> can I select `gpr-set-externals` by compiler version?

No, this is not supported at the moment. Only the variables listed under SYSTEM in `alr version`.

reznikmm:
I've disabled all compilers except `gnat_native` to have a single `TARGET` per platform.

PR #686, opened 2022-11-14
Add gnatdoc 23.0

reznikmm:


Fabien-Chouteau (approved):

index/gn/gnatdoc/gnatdoc-23.0.0.toml
```diff
@@ -0,0 +1,35 @@
...
+maintainers-logins = ["godunko", "reznikmm"]
+licenses = "GPL-3.0-only WITH GCC-exception-3.1"
+website = "https://github.com/AdaCore/gnatdoc"
+
+project-files = ["gnat/gnatdoc.gpr"]
+
+executables = ["gnatdoc4"]
+
+[environment.PATH]
+append = "${CRATE_ROOT}/bin"
```
Fabien-Chouteau: Why are you adding this? Alire will find the executable automatically?

Fabien-Chouteau (changes requested):

index/gn/gnatdoc/gnatdoc-23.0.0.toml
reznikmm: Modifying `PATH` lets the user run `gnatdoc4` without full path specification. I'm not sure about `executables`, if this is used by alire or deprecated now.

Fabien-Chouteau: Alire uses `executables` field to find the binary for `alr run`. I understand what you want to do but I don't think we have that for the other crates. Ping @mosteo .

mosteo: I'm not sure how this helps? By using `alr printenv` to set the path? For `alr exec`?

reznikmm: I thought that having `gnatdoc4` in the `PATH` will help users find it and make usage easier (inside `alr exec`, `alr printenv` and with `gnatdoc` in dependency list).

Are you suggesting to drop these lines?

mosteo: OK, I see. I'm not against it but I expect it will become moot once we have a working `alr install`, which I guess would be the preferred option for this kind of binary crate.

reznikmm: Ok, so I drop them

Fabien-Chouteau (approved):

PR #695, opened 2022-12-03
Add BingAda to the index

mgrojo:


mgrojo:
I don't know why Fedora is failing with this:
https://github.com/alire-project/alire-index/actions/runs/3608397008/jobs/6080977757#step:9:5699
```
/usr/bin/ld: cannot find -lcsfml-system
/usr/bin/ld: cannot find -lcsfml-audio
/usr/bin/ld: cannot find -lcsfml-network
/usr/bin/ld: cannot find -lcsfml-audio
/usr/bin/ld: cannot find -lcsfml-graphics
/usr/bin/ld: cannot find -lcsfml-window
collect2: error: ld returned 1 exit status
gprbuild: link of bingada.adb failed
```
It seems to install it:
https://github.com/alire-project/alire-index/actions/runs/3608397008/jobs/6080977757#step:9:392
```
Installed:
  CSFML-2.5.1-4.fc36.x86_64                SFML-2.5.1-10.fc36.x86_64            
```

CSFML worked in Fedora for #635, and it installed the exact same package version:
https://github.com/alire-project/alire-index/actions/runs/3055374609/jobs/4935682916#step:9:241

The only strange thing I found is that it tries to run `pacman`. That is not happening in #635.
https://github.com/alire-project/alire-index/actions/runs/3608397008/jobs/6080977757#step:9:248

mosteo:
The pacman message is spurious -- something we have to remove but inoffensive.

AFAIS, the difference with #635 is that no linking is taking place there, as there is no executable. I'm unfamiliar with this particular library so I'm unsure right now if this is a matter of missing dependencies or with the library configuration.

If you're in a hurry to merge, I'd say to mark fedora as unavailable for the time being in your crate until we can look into the matter.

mosteo:
I understand here that the `-devel` is the equivalent of Debian's `-dev`, right? Not that this is some kind of unstable version.

mgrojo:
Yes, it's the equivalent of  `-dev` for Fedora. We need that to get the symbolic links to the libraries without version information, since this is required for `-l` switches.

I was thinking the Fedora package was already tested, but it was neither, as you say, in #635, nor it was in `play_2048` crate. In fact, I was fooled by its crate status in https://alire-crate-ci.ada.dev/crates/play_2048.html If you follow the status badge for  [fedora-33 10.2.1 success] it displays a [log](https://alire-crate-ci.ada.dev/status/pl/play_2048/1.0.0/linux/fedora-33/gnat=10.2.1/play_2048-1.0.0.yaml.log), which I'd say it's for a Debian based distribution (references to `apt` and deb packages). Maybe there's some bug in the crate status scripts.

mosteo:
I see, thanks for the detailed reply. I see the check that shows as OK is an older one, that may be related.

mgrojo:
> I understand here that the `-devel` is the equivalent of Debian's `-dev`, right? Not that this is some kind of unstable version.

Oh, I see now that I used inaccurate wording for my commit message. Yes, it's the development package for the library, not a "development (=unstable) version". But anyway, I suppose it's a minor wording problem.

mosteo (approved):

PR #697, opened 2022-12-08
Update basalt to 0.3.0

jklmnn:


index/ba/basalt/basalt-0.3.0.toml
```diff
@@ -0,0 +1,10 @@
+description = "Collection of formally verified building blocks"
+name = "basalt"
+version = "0.3.0"
+licenses = "AGPL-3.0-only"
+maintainers = ["kliemann@adacore.com"]
+maintainers-logins = ["jklmnn"]
```
mosteo, line 6: This would be a good opportunity to add some tags to the release.

jklmnn: I added the tags that I found relevant for this library.

jklmnn:
I added a separate version for legacy SPARK versions as the gnatprove currently in alire doesn't seem to support the `Always_Return` annotation.

mosteo (approved):

mosteo:
Merged, thanks.

PR #700, opened 2022-12-12
This are a tools to create a rom images für MOS 6502 systems

krischik:
This program can be used to create a ROM images for the MOS 6502 tutorial from Ben Eater. One can find the full tutorials here:

https://eater.net/

### DIFF
--- a/index/mo/mos6502_create_led/mos6502_create_led-1.0.0.toml
+++ b/index/mo/mos6502_create_led/mos6502_create_led-1.0.0.toml
@@ -18,6 +18,9 @@ executables         = ["create_led"]
 website             = "https://sourceforge.net/projects/tutorial-6502"
 tags                = ["mos-6502", "tools", "retrocomputing", "ada-2022"]
 
+[[depends-on]]
+gnat = ">12"
+
 # vim: set textwidth=0 nowrap tabstop=8 shiftwidth=4 softtabstop=4 expandtab :
 # vim: set filetype=toml fileencoding=utf-8 fileformat=unix foldmethod=diff :
 # vim: set spell spelllang=en_gb :
--- a/index/mo/mos6502_create_led/mos6502_create_led-1.0.0.toml
+++ b/index/mo/mos6502_create_led/mos6502_create_led-1.0.0.toml
@@ -1,0 +1,31 @@
+name                = "mos6502_create_led"
+description         = "Create ROM with little program"
+long-description    = """This is a tool to create a rom image with little program.
+
+This program can be used to create a ROM immage for the MOS 6502 tutorial
+from Ben Eater.
+
+Development versions available with:
+
+alr index --add "git+https://github.com/krischik/alire-index.git#develop" --name krischik
+"""
+version             = "1.0.0"
+licenses            = "GPL-3.0-or-later"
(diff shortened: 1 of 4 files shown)


krischik:
Not all backends are Ada 2022 capable. 

```
/home/runner/work/alire-index/alire-index/mos6502_create_led_1.0.0_54914b60/src/create_led.adb:57:31: illegal character, replaced by "("
```

That's a bit disappointing.

Fabien-Chouteau:
@krischik you have to specify a dependency on GNAT 12 because it is the first version with Ada 2022 support:

e.g.

```
[[depends-on]]
gnat = ">12"
```

krischik:
Not for these three. I could downgrade them. Of course this being part of a tutorial I just like to be at the forefront.

krischik:
@Fabien-Chouteau 

> ```
> [[depends-on]]
> gnat = ">12"
> ```

That worked better. Only one error. The strange part is that I use macOS myself.

Fabien-Chouteau:
The build on macOS is failing because GNAT Community 2020 is satisfying the ">12" dependency but still doesn't support Ada2022.

@krischik  you can set `gnat ">=12 && < 2000"`.

@mosteo Maybe we should stop testing with this 2+ years old compiler.

krischik:
@Fabien-Chouteau Can you restart the windows build? It hung on an _„Press any key to continue“_.

```
[59454] Writing maintenance tool: "D:\\a\\_temp\\gnat-2020-20200818-x86_64-windows-bin/maintenancetool.exe.new"
[59536] Wrote permissions for maintenance tool.
[60888] Maintenance tool restart: false.
Press any key to continue . . . 
Error: The operation was canceled.
```

mosteo:
> @mosteo Maybe we should stop testing with this 2+ years old compiler.

I've thought the same at times. If you deem it appropriate it's ok by me.

Failed tests restarted.

index/mo/mos6502_create_led/mos6502_create_led-1.0.0.toml
```diff
@@ -0,0 +1,34 @@
...
+alr index --add "git+https://github.com/krischik/alire-index.git#develop" --name krischik
+"""
+version             = "1.0.0"
+licenses            = "GPL-3.0-or-later"
+authors             = ["Martin Krischik <krischik@users.sourceforge.net>"]
+maintainers         = ["Martin Krischik <krischik@users.sourceforge.net>"]
+maintainers-logins  = ["krischik"]
+executables         = ["create_led"]
+website             = "https://sourceforge.net/projects/tutorial-6502"
+tags                = ["mos-6502", "tools", "retrocomputing", "ada-2022"]
```
mosteo: Given that we have already `ada1987` and `ada1995` tags, I suggest to change these `ada-2022` tags to `ada2022`. No need to repackage, just update these files in the PR for the index.

mosteo (changes requested):

index/mo/mos6502_create_led/mos6502_create_led-1.0.0.toml
krischik: Done.

mosteo (approved):

mosteo (approved):

PR #701, opened 2022-12-19
Update URL in Matreshka crates.

reznikmm:
The site https://forge.ada-ru.org isn't available any more, so replace source archives with GitHub release URLs.

### DIFF
--- a/index/ma/matreshka_amf/matreshka_amf-18.1.0.toml
+++ b/index/ma/matreshka_amf/matreshka_amf-18.1.0.toml
@@ -1,7 +1,7 @@
 description = "Implementation of OMG's Meta Object Facility (MOF)"
 name = "matreshka_amf"
 version = "18.1.0"
-website = "https://forge.ada-ru.org/matreshka"
+website = "https://github.com/godunko/matreshka"
 authors = ["Vadim Godunko"]
 licenses = "BSD-3-Clause"
 maintainers = ["Vadim Godunko <vgodunko@gmail.com>", "Maxim Reznik <reznikmm@gmail.com>"]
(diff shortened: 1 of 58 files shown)


index/ma/matreshka_amf/matreshka_amf-18.1.0.toml
```diff
@@ -1,18 +0,0 @@
-description = "Implementation of OMG's Meta Object Facility (MOF)"
```
Fabien-Chouteau, line 1: Deletion of a crate release is not allowed. Why do you want to remove those releases?

Fabien-Chouteau (changes requested):

index/ma/matreshka_amf/matreshka_amf-18.1.0.toml
reznikmm: We lost `ada-ru.org` domain and 18.1 source archive.

reznikmm: There is an [archive in Fedora](https://src.fedoraproject.org/repo/extras/matreshka/) but sha512 is different. I don't understand why :(

Fabien-Chouteau: I will way on @mosteo to comment here, maybe the best solution is to mark the release a unavailable in all cases.

reznikmm: We have found a missing file (`matreshka-18.1.tar.gz`).

Fabien-Chouteau (approved):
Even if the builds are failing I am approving this since it's only changing URLs.

mosteo (approved):

mosteo:
I guess this was submitted when we did fewer checks. Agreed with Fabien to go ahead since this preserves the current situation.

PR #704, opened 2022-12-31
Submission of the 'aShell' project crate for entry into the 'Crate of the Year' contest. 

charlie5:
Submission of the 'aShell' project for entry into the 'Crate of the Year' contest.


mosteo:
Sorry for the trouble with msys2, we need to push a fix about it. Since the failure is during alire's own setup, it happens before your disabling of the crate for `msys2` can make a difference (in case you want to re-enable it).

charlie5:
No problem, Mosteo :) .

Unfortunately neither of the 'florist's (my crates dependency) support windows/msys. I looked into trying to add msys support for florist but it is a major project in itself.

mosteo (approved):

PR #707, opened 2023-01-12
Bump alr version in use to 1.2.2

mosteo:


mosteo:
Merging this one so I can unclog the submissions.

PR #711, opened 2023-01-16
Maintenance work on libsimpleio and remoteio crates.

pmunts:
Fixed or eliminated many (but not all) compiler warning messages.

Major improvements to Microelektronika Click board support, especially for BeagleBone platforms.


pmunts:
Abandon this pull request.  Will try again with a work-around for the GNAT internal error.

PR #719, opened 2023-01-19
remoteio release 1.20663.1

pmunts:
Added support for MacOS, with libhidapi and libusb provided by homebrew.

Tested on a Mac Mini M1 running Ventura:

Darwin bethany.local 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:04:44 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T8103 arm64

It seems likely that the post-fetch action would need to be changed for an Intel Mac, but I don't have one to wring it out.


pmunts:
This should have been pulled from a different branch than stable, and will be superceded anyway.

PR #722, opened 2023-01-19
Added tags for gnavi / gwindows

zertovitch:


mosteo:
It seems the latest compiler is having an extra warning/error.

zertovitch:
Ouch, yeah it's a catch-22 with an intrinsic instruction. It seems to be an incompatible change in GCC...
https://github.com/zertovitch/gwindows/issues/9

PR #725, opened 2023-01-19
Added tags for MathPaqs

zertovitch:


zertovitch:
The failure on macos is due to linker option "--gc-sections". It has already been corrected in the project; issue doesn't stem from proposed change in the .toml file.

zertovitch:
New version in another PR; this PR is no more needed.

PR #734, opened 2023-01-22
gnatstudio 23.0.0-20220512

mosteo:
Windows has an installer with which we cannot work directly. macOS has no binary release on github.


Fabien-Chouteau:
I am not sure it's good to have a heterogeneous user experience depending on the platform.

mosteo:
Of course being an AdaCore product I'll defer to your criterion.

But, I'll advocate a little for the devil ;-) and say that current experience is homogeneous up to the point of reaching the GitHub releases page; then Windows users get an installer while Linux users get to untar and set-up manually.

krischik:
I remember when I tried to build gnatstudio myself and it was an absolute nightmare. I didn't even get the GtkAda sources to compile. Looking forward to you progress. Being a macOS User I would welcome a successful build.

simonjwright:
The gnatstudio [on Sourceforge](https://sourceforge.net/projects/gnuada/files/GNAT_GPL%20Mac%20OS%20X/2022-monterey/) is pretty good!

PR #735, opened 2023-01-23
Crates for AdaCore Ada Web Services for Debian 11 and MuntsOS Embedded Linux

pmunts:


Fabien-Chouteau (changes requested):
Hi @pmunts I am not sure to understand what you want to achive with those crates.

The archive only contain an alire manifest and a gpr file referencing pre-built system libraries using absolute paths.

This is not the correct way to declare dependencies on system libraries.

pmunts:
Abandoning

PR #736, opened 2023-01-26
MuntsOS Embedded Linux support crates

pmunts:
These crates make it as easy as possible to target a Linux microcomputer board running MuntsOS Embedded Linux, by just running something like `alr with muntsos_beaglebone` in the Alire program project directory.

I know these crates will fail the build tests.  Please merge them anyway.  I want to use them for Ada programming tutorial classes, including (hopefully) at Lisbon in June.  These crates hide most of the complexity of using Alire for embedded Linux targets.

### DIFF
--- a/index/mu/muntsos_beaglebone/muntsos_beaglebone-1.0.0.toml
+++ b/index/mu/muntsos_beaglebone/muntsos_beaglebone-1.0.0.toml
@@ -1,0 +1,33 @@
+name = "muntsos_beaglebone"
+description = "MuntsOS Embedded Linux support for BeagleBone targets"
+tags = ["muntsos", "embedded", "linux", "arm", "beaglebone"]
+version = "1.0.0"
+website = "https://github.com/pmunts/muntsos"
+
+authors = ["Philip Munts"]
+maintainers = ["Philip Munts <phil@munts.net>"]
+maintainers-logins = ["pmunts"]
+
+project-files = ["aws.gpr", "libsimpleio.gpr"]
+
+[available."case(os)"]
+"linux" = true
+"..." = false
+
+[configuration]
+disabled = true
+
+[[depends-on]]
+muntsos_dev_beaglebone = "*"
+
+[[actions]]
+type = "post-fetch"
+command = ["sh", "-c", "./postfetch"]
+
+[origin]
+hashes = [
+"sha256:64e2009c383d63cd85383e8f435c9a5c44ef4f7a48dfbb87698798db75f605d1",
+"sha512:6ad6de32f56c8fbe72c3d2b5e17b830f6c18f264f8d438e8714b1886e4448b20ae0df3d635308af7d0f3264daa5ff6aad1af490b01bc665c04c5676f06a5c03c",
+]
+url = "http://repo.munts.com/alire/muntsos_beaglebone-1.0.0.tbz2"
+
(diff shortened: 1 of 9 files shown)


mosteo:
Hi @pmunts. I'm trying to get a clear idea of what would be the best way to properly achieve what you want. For one thing, I see in the contents of these crates soft links to files that won't exist except (I guess) on MuntOS. So I can see a couple of options:

- You could have your own index for all MountOS-specific crates. It would take a single command to set it up in your tutorial: `alr index --add <url> --name muntos` and you would have complete control over that index contents. 
- Besides, at the bare minimum you should also submit an [external crate](https://alire.ada.dev/docs/#external-releases) that is detected only on MuntOS, and on which these crates depend. This way it would be clear that these crates cannot be installed otherwise. I suggest using the command-line tool kind, with an ad-hoc executable supplied with MuntOS.

pmunts:
My overall goal is to use Alire for teaching embedded systems programming using a Raspberry Pi as a target.  I need to hide absolutely as much as possible any magic necessary to get cross-compiling working.

For a student to just type "alr with muntsos_raspberrypi1" after "alr init --bin" seems reasonable, as it fits the alire paradigm nicely (principle of least surprise).  Manually editing the .gpr file or even copying the correct default.cgpr seems less reasonable/more difficult/more surprising.

I can define external crates for each of the MuntsOS cross-toolchain metapackages.  Each metapackage pulls in all of the required components for a single cross-toolchain.  Since I distribute the cross-toolchain components as Debian packages (though from my own repository), that works quite nicely.  It is not clear to me how to define a proper crate for an external cross-toolchain, though, because somehow a .cgpr file needs to be passed from the cross-toolchain crate to the superordinate program project, preferably without any manual intervention from the developer.

Adding "for Target use..." to the program project .gpr is rather sketchy, as the target identifier is just syntactic sugar for <target>.cgpr in a place gprbuild expects to find it, normally /usr/share/gpr/ for Linux distribution toolchains.  From my testing, I've concluded that the Alire gprbuild doesn't look in /usr/share/gpr but rather some share/gpr/ directory relative to the toolchain installation base.

Unfortunately, AdaCore didn't define "for Config use..." in a project file.  I'm actually thinking about submitting a feature request for that, as I had the same problem before outside the Alire context.

-------- Original Message --------
Subject: Re: [alire-project/alire-index] MuntsOS Embedded Linux support crates (PR #736)
From: "Alejandro R Mosteo" ***@***.***>
Date: Thu, February 2, 2023 9:22 am
To: "alire-project/alire-index" ***@***.***>
CC: "pmunts" ***@***.***>,"Mention" ***@***.***>

Hi @pmunts. I'm trying to get a clear idea of what would be the best way to properly achieve what you want. For one thing, I see in the contents of these crates soft links to files that won't exist except (I guess) on MuntOS. So I can see a couple of options:

- You could have your own index for all MountOS-specific crates. It would take a single command to set it up in your tutorial: `alr index --add <url> --name muntos` and you would have complete control over that index contents. 
- Besides, at the bare minimum you should also submit an [external crate](https://alire.ada.dev/docs/#external-releases) that is detected only on MuntOS, and on which these crates depend. This way it would be clear that these crates cannot be installed otherwise. I suggest using the command-line tool kind, with an ad-hoc executable supplied with MuntOS.

mosteo:
Thanks for the context, @pmunts. Frankly, in this case I still think you'd be better served by having your own index to which you can push updates without having to wait for our review; I guess it may be important to you to be able to pass hotfixes to your students rather than waiting for our review on any of the crates you're using with your students. Configuring a second index on top of the community one won't disturb your workflow in any way, I'd think.

That said, if you still want these `muntsos_*` crates to be in the community index, with the current addition of the `muntsos_dev_*` packages to clearly expose the dependency on those extra system packages I don't see a problem. Just confirm to me that you really don't want to use your own index.

pmunts:
Please do merge these crates to the index.

mosteo (approved):

PR #744, opened 2023-02-18
Euler Tools 1.0.1

rocher:
First public release.

### DIFF
--- a/index/eu/euler_tools/euler_tools-1.0.1.toml
+++ b/index/eu/euler_tools/euler_tools-1.0.1.toml
@@ -1,0 +1,15 @@
+name = "euler_tools"
+description = "Assortment of mathematical functions to solve Project Euler problems."
+version = "1.0.1"
+
+authors = ["Francesc Rocher"]
+maintainers = ["Francesc Rocher <francesc.rocher@gmail.com>"]
+maintainers-logins = ["rocher"]
+licenses = "MIT"
+tags = ["project-euler", "mathematics"]
+website = "https://github.com/rocher/euler_tools"
+
+[origin]
+commit = "0c2dd7793ad1927ec8425035fefd9c75eb31a2df"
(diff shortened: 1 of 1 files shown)


mosteo:
Your dependency on gnat is only half-complete because it also admits the CE editions with versions 2019 and up. You can fix it as `gnat>=12 & gnat<2000` for example.

rocher:
> Your dependency on gnat is only half-complete because it also admits the CE editions with versions 2019 and up. You can fix it as `gnat>=12 & gnat<2000` for example.

Yes, I saw this problem.
Also, I was using the `-gnat2022` switch, which looks like is only supported in `ubuntu-latest`. Changed to `-gnat2020` and run unit tests in all platform and now it works.

rocher:
Don't understand why [this workflow run passed](https://github.com/rocher/euler_tools/actions/runs/4223138022) and not the one in the current PR.
What I want is to use the currently implemented Ada 2022 extensions: the `-gnat2020` switch worked fine for me. Why not here?

mosteo:
That's because the failing workflow is using a GNAT Community 2020 compiler, which doesn't support the switch. That's why you can avoid the problem with the extra restriction on versions in the gnat dependency.

rocher:
Ok, fixed dependencies but failed again: [was this a system error](https://github.com/alire-project/alire-index/actions/runs/4225834266/jobs/7338591104#step:6:47)?
Will push again, just in case.

mosteo (approved):

mosteo:
> Ok, fixed dependencies but failed again: [was this a system error](https://github.com/alire-project/alire-index/actions/runs/4225834266/jobs/7338591104#step:6:47)?

Yes, I've seen that before on macOS instances; it's some kind of disk failure. Re-running as you did is enough to fix it.

PR #752, opened 2023-03-02
Mage 0.3.0

leogermond:
I reworked most of the API, so that it is both easier to use and
has new modules:

`Mage.Input` for uniform handling of user inputs
`Mage.Event` for all events
`Mage.Log` for tiny logs

There's still a lot to do, but we're getting slowly there


mosteo:
The error is in our checks. Will fix shortly.

mosteo (approved):

PR #765, opened 2023-03-16
AdaCL module adacl_test release 5.9.9

krischik:
# Rationale:

In the „Using pins for crate testing“ chapter it is suggested that unit test are moved to they own test project so that the main project doesn't depend on AUnit (or similar). This makes perfect sense.

However if as suggested the test is a places as subdirectory inside the main project the test are never actually run inside the [alire-index](https://github.com/alire-project/alire-index) environment. Which I consider desirable if only to make sure that that the test will run outside my own build environment.

For now I add this pull request as a draft only to see if the test create actually builds and the test are actually executed.

### DIFF
--- a/index/ad/adacl_test/adacl_test-5.9.9.toml
+++ b/index/ad/adacl_test/adacl_test-5.9.9.toml
@@ -1,0 +1,56 @@
+name                        = "adacl_test"
+description                 = "Ada Class Library - AUnit test"
+long-description            = """AUnit tests for adacl
+
+Development versions and testsuite available using the follwowing index:
+
+```sh
+alr index --add "git+https://github.com/krischik/alire-index.git#develop" --name krischik
+```
+
+Source code including AUnit tests available on [SourceForge](https://git.code.sf.net/p/adacl/git)
+"""
+version                     = "5.9.9"
+licenses                    = "GPL-3.0-or-later"
+authors                     = ["Martin Krischik"]
+maintainers                 = ["Martin Krischik <krischik@users.sourceforge.net>"]
+maintainers-logins          = ["krischik"]
+website                     = "https://sourceforge.net/projects/adacl/"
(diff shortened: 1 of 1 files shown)


krischik:
The test where executed which is good. Sadly the windows test failed yet the windows build is considered succesfull:

```
Link
   [link]         adacl_test-main.adb
Note: Running post_build actions for adacl_test=5.9.9...
OK AdaCL.Strings : Test_Append_All_01 : simple append
OK AdaCL.Strings : Test_Append_All_02 : skip append
OK AdaCL.Strings : Test_Append_All_03 : skip and append
OK AdaCL.Trace : Test_Enable_Trace_01 : enable trace
OK AdaCL.Trace : Test_Write_To_File_01 : enable trace
OK AdaCL.Trace : Test_Write_String_01 : string output
OK AdaCL.Trace : Test_Function_Trace_01 : function trace single string
FAIL AdaCL.Trace : Test_Function_Trace_02 : function trace two strings
    Trace output output file is 195 bytes.
    at adacl-trace-test.adb:105
FAIL AdaCL.Trace : Test_Function_Trace_03 : function trace two level
    Trace output output file is 195 bytes.
    at adacl-trace-test.adb:133
Total Tests Run:   9
Successful Tests:  7
Failed Assertions: 2
Unexpected Errors: 0
```

Not quite what I was hoping for.

mosteo:
> Not quite what I was hoping for.

And rightly so. Does the aunit testsuite ends with exit code /= 0 in case of failures? If, so, the post-build action should fail.

krischik:
> And rightly so. Does the aunit testsuite ends with exit code /= 0 in case of failures?

Good point. Most unit testing frameworks will handle the return code automatically however AUnit needs a lot of manual work and it seems handling the return value is one of them. I'll fix that.

krischik:
Release and Pull request abandoned

Reason: Problems which need code changes.

PR #770, opened 2023-03-19
On macOS, get libgmp from Homebrew.

simonjwright:
This is stage 1 of what may be a tricky process!

While working on [libadalang2xml](https://github.com/simonjwright/libadalang2xml), I found that both langkit_support and libadalang require libgmp to build their relocatable libraries but can’t find it; gnatcoll_gmp manages this by supporting LDFLAGS, but the two packages above don’t. I’ll raise an issue (or two?), but that’ll take some time to work through and is really not ideal.

### DIFF
--- a/index/li/libgmp/libgmp-external.toml
+++ b/index/li/libgmp/libgmp-external.toml
@@ -10,6 +10,7 @@ kind = "system"
 "debian|ubuntu" = ["libgmp-dev"]
 arch = ["gmp"]
 msys2 = ["mingw-w64-x86_64-gmp"]
+homebrew = ["gmp"]
 
 [[external]]
 kind = "version-output"


mosteo:
Do you need this merged already? Or can it wait until the rest is ready?

simonjwright:
> Do you need this merged already? Or can it wait until the rest is ready?

At the moment, I’ve abandoned any attempt to get libadalang, langkit built under Alire on macOS, because of this problem and (more tedious) AdaCore’s insistence on building those static libraries as SALs using a technique which is _impossible_ on Darwin’s mach-o object format.

Instead, libadalang2xml on macOS forces its build using an external compiler (i.e. one of the builds I’ve been releasing on Github, which includes prebuilt gmp, langkit, libadalang).

```
[[depends-on]]

[depends-on.'case(os)'.macos]
# The gnat_native compilers don't provide libgmp, and there's no
# current way of making it visible to the Libadalang/Langkit_Support
# relocatable library builds.
gnat_external = ">=12.2.0"

[depends-on.'case(os)'.'...']
libadalang = "^23.0.0"
```

That said, I still think this change would be worth while, for other contexts.

mosteo (approved):

mosteo:
Alright, merging then!

PR #771, opened 2023-03-19
Elevator 1.0.0

tshoang:
A simple model of an elevator system. This is used for teaching Ada/SPARK together with Event-B modelling.

### DIFF
--- a/index/el/elevator/elevator-1.0.0.toml
+++ b/index/el/elevator/elevator-1.0.0.toml
@@ -1,0 +1,19 @@
+name = "elevator"
+description = "A Simple Elevator System"
+version = "1.0.0"
+
+authors = ["Thai Son Hoang"]
+maintainers = ["Thai Son Hoang <T.S.Hoang@soton.ac.uk>"]
+maintainers-logins = ["tshoang"]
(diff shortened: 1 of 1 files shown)


mosteo:
I understand this is an executable, and its use of SPARK is integral to its usefulness? So it's OK to have a dependency on `gnatprove` in this case.

tshoang:
Dear @mosteo,
 Thank you very much for looking at this. I'm teaching SPARK/Ada at the University of Southampton as a part of a 2nd Year module. This is an example for reasoning formally about a control system (an elevator system) using SPARK, hence the dependency on gnatprove.
 This is more a teaching material rather than a library, so I don't know if it is something that feasible with alire. I really like the concept of distributing various materials via alire and hope that it works with teaching materials as well.
 Best regards,
Son

mosteo:
Thanks for the explanation, @tshoang . As a professor myself, I don't see a problem with educative projects making into the index. On the contrary, they can make for very good learning examples and use cases.

The thing is that we don't want libraries to depend on gnatprove directly, as that would be cumbersome to users of the library, so I just wanted to be sure that having that dependency in this case was proper.

mosteo (approved):

PR #778, opened 2023-03-31
Add esp_idf crate

jklmnn:
This is the crate for https://github.com/jklmnn/esp32c3-ada-bindings. I decided to call it `esp_idf` here as the binding will likely work for all devices that are supported by the ESP-IDF, not just the Esp32C3.


index/es/esp_idf/esp_idf-0.1.0.toml
```diff
@@ -0,0 +1,11 @@
+description = "Ada binding for the ESP-IDF SDK"
+name = "esp_idf"
+version = "0.3.0"
```
mosteo: This is mismatched with the file name and the tarball.

mosteo (changes requested):

index/es/esp_idf/esp_idf-0.1.0.toml
jklmnn: Typical copy paste error, I fixed it.

mosteo:
It seems there's some missing dependency or misnaming wrt a certain `config/esp32c3_lib_config.gpr` file

jklmnn:
I fixed the renaming. To prevent changing a release I updated the version to 0.1.1 (and omit publishing the broken 0.1.0 on alire).

mosteo (approved):

PR #784, opened 2023-04-11
6502 tutorial module atari_atr_tools release 1.1.0

krischik:
* Update to current AdaCL
* Add AUnit Tests

### DIFF
--- a/index/at/atari_atr_tools/atari_atr_tools-1.1.0.toml
+++ b/index/at/atari_atr_tools/atari_atr_tools-1.1.0.toml
@@ -51,12 +51,12 @@ website                     = "https://sourceforge.net/projects/tutorial-6502"
 tags                        = ["atari", "tools", "retrocomputing", "ada-2022"]
 
 [build-switches]
-development.runtime_checks  = "Overflow"
-release.runtime_checks      = "Default"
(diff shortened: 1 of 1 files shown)


krischik:
Why where the build canceled?

mosteo:
Builds are cancelled when another check in the same workflow fails. You can push fixes and they'll all retrigger.

krischik:
> Builds are cancelled when another check in the same workflow fails. You can push fixes and they'll all retrigger.

Line 54 is correct:

```
[build-switches]
development.runtime_checks  = "Overflow"
release.runtime_checks      = "Default"
validation.runtime_checks   = "Everything"
development.contracts       = "Yes"
release.contracts           = "No"
validation.contracts        = "Yes"
```

But it's also the only option with a `.` in the name.

krischik:
The pins from the test cases need to be removed manually. Close at this needs a hot-fix with new tar-file.

mosteo:
> Line 54 is correct:

Yes, that's a different problem. The check-author workflow can't cope with some more recent TOML syntax, so that workflow failing shouldn't be a concern. It also won't cause the cancellation of the rest. The other jobs that do build the crate are aborted when other in the same workflow batch fails.

PR #786, opened 2023-04-11
6502 tutorial module mos6502_create_nop release 1.1.0

krischik:
* Update to current AdaCL
* Add AUnit Tests


krischik:
Also cancelled builds.

krischik:
The pins from the test cases need to be removed manually. Close at this needs a hot-fix with new tar-file.

PR #789, opened 2023-04-12
6502 tutorial module mos6502_create_reset release 1.1.1

krischik:
* Update to current AdaCL
* Add AUnit Tests
* Fix Alire configuration for the AUnit test.


krischik:
`make cleanup` didn't clean up well enough so an actual bug in the unit tests made it into the release.

@mosteo Well, at least we now know the execution of the the unit tests work as expected.

krischik:
`make cleanup` didn't clean up well enough so an actual bug in the unit tests made it into the release.

@mosteo Well, at least we now know the execution of the the unit tests work as expected.

PR #801, opened 2023-04-29
Adding libadalang2xml-1.0.0.

simonjwright:
This tool generates an XML representation of the structure of an Ada source file.


simonjwright:
The error in Check PR is in the `CFLAGS.set` line below:
```
[environment.'case(os)'.macos]
# This code was originally developed on macOS, which is why the CFLAGS
# setting specifies where the includes for external libraries used by
# GNATCOLL are to be found (Homebrew places items it manages in
# /usr/local or /opt/homebrew on Intel or Apple silicon respectively).
CFLAGS.set = "-I/usr/local/include -I/opt/homebrew/include"
```
which works just fine in the builds.

mosteo:
Yes, I must fix that test which keeps on giving false negatives and confusing people.

mosteo (approved):

PR #802, opened 2023-05-03
Added base package for Ada ISO Library

AJ-Ianozi:
This is my first time publishing to alire, so please let me know if I made any mistakes. This is the base (mostly empty) package but I'm hoping for it to be a dependency for two other alire items I'll be publishing: the almost-completed [iso_countries](https://github.com/ada-iso/iso_countries) and (the soon-to-be worked) on iso_curriences.  Before I attempt to publish iso_countries I'd like to have iso_base already in there.  Please let me know if you have any questions!

### DIFF
--- a/index/is/iso_base/iso_base-1.0.0.toml
+++ b/index/is/iso_base/iso_base-1.0.0.toml
@@ -1,0 +1,13 @@
+name = "iso_base"
+description = "Base package for Ada ISO Project"
+version = "1.0"
+licenses = "MIT"
+website = "https://github.com/ada-iso/iso_base/"
+authors = ["AJ Ianozi"]
+maintainers = ["AJ Ianozi <aj@ianozi.com>"]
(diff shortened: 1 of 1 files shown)


AJ-Ianozi:
I'm guessing this failed because my GitHub username is in all lowercase, which I'm guessing I messed up when I first configured Alire on that machine way back when.

AJ-Ianozi:
Going to re-do this to fix the name in alire.toml projects, then I'll re-submit.

mosteo:
Hi @AJ-Ianozi , thank you for becoming a new contributor. Just so you know, you can incrementally fix your PR without closing and reopening new ones.

Also, you can submit several manifests in the same PR (e.g., your base dependency and the dependent one). Although not generally recommended, in some cases it may make sense (you won't have to wait for the base to be accepted and you can check that both crates work together as intended in one go.)

PR #803, opened 2023-05-03
Added base package for Ada ISO Library

AJ-Ianozi:
This is my first time publishing to alire, so please let me know if I made any mistakes. This is the base (mostly empty) package but I'm hoping for it to be a dependency for two other alire items I'll be publishing (the almost-completed [iso_countries](https://github.com/ada-iso/iso_countries) and the soon-to-be worked on iso_curriences).  Before I attempt to publish iso_countries I'd like to have iso_base already in there.  Please let me know if you have any questions!

### DIFF
--- a/index/is/iso_base/iso_base-1.0.0.toml
+++ b/index/is/iso_base/iso_base-1.0.0.toml
@@ -1,0 +1,13 @@
+name = "iso_base"
+description = "Base package for Ada ISO Project"
+version = "1.0"
+licenses = "MIT"
+website = "https://github.com/ada-iso/iso_base/"
+authors = ["AJ Ianozi"]
+maintainers = ["AJ Ianozi <aj@ianozi.com>"]
+maintainers-logins = ["AJ-Ianozi"]
+
+[origin]
+commit = "bc3b5c304b83f2a622f12e5af7bc26211ed806f5"
+url = "git+https://github.com/ada-iso/iso_base.git"
+


Fabien-Chouteau:
Hi @AJ-Ianozi, what's the reason for splitting `iso_base` and `iso_countries`? Having a crate with just a 3 lines spec file brings a lot of overhead for you and the users.

AJ-Ianozi:
> Hi @AJ-Ianozi, what's the reason for splitting `iso_base` and `iso_countries`? Having a crate with just a 3 lines spec file brings a lot of overhead for you and the users.

I was going by the [Best practices](https://alire.ada.dev/docs/#best-practices)  section on splitting:

> If your project is a collection of components (like GNATcoll for instance) and each component has different dependencies, you should consider splitting the collection into multiple Alire crates. The components can still be maintained in the same repository and use the same release archive/commit (e.g. [gnatcoll_sqlite](https://alire.ada.dev/crates/gnatcoll_sqlite), [gnatcoll_sql](https://alire.ada.dev/crates/gnatcoll_sql), [gnatcoll_postgres](https://alire.ada.dev/crates/gnatcoll_postgres)).

In this case, right now I only have `iso_countries` but later I'll also have `iso_currencies` and `iso_languages`. All three of these will require `iso.ads`, so my choice is to maintain and ship `iso.ads` with all three or have an `iso_base` that can be a dependency of all three.

I have a feeling that `iso_currencies` will actually have a dependency of `iso_countries` (if I want to map countries with currencies), and then I'll have two `iso.ads` files if I did that unless I had `iso_base` and `iso_countries` as separate dependencies.

I also considered if I should just make the `iso.ads` in `iso_countries` the source of record (the Wayland library seems to do that) but I would have to either have a second `iso.ads` in every library that doesn't utilize `iso_countries` or have `iso_countries` a dependency of everything else, even if it's not needed (which it very will could be, but I'm not sure yet).

In the long term, once I notice commonality that I'll be using between all of the ISO line , I can just put that in `iso_base`.

Pleases let me know if I missing something obvious, I just wanted a single source of truth. Any other ways to deal with this I'm missing?

Edit: If you guys think it's better I can just cross that road when I reach it, ship `iso.ads` with `iso_countries`, and keep doing so until I encounter an iso library that doesn't require countries (and then submit `iso_base`?). I think that `iso_currencies` will rely on `iso_countries` but that may be it.

Fabien-Chouteau:
> > Hi @AJ-Ianozi, what's the reason for splitting `iso_base` and `iso_countries`? Having a crate with just a 3 lines spec file brings a lot of overhead for you and the users.
> 
> I was going by the [Best practices](https://alire.ada.dev/docs/#best-practices) section on splitting:

In this paragraph "each component has different dependencies" means external dependencies.

For instance with GNATcoll, `gnatcoll_lzma` and `gnatcoll_gmp` are separate because they both have different external dependencies, respectively `liblzma` and `libgmp` which are C projects. External dependencies are more difficult to deal with so it's best to split the project here.

For your ISO project I don't see any external dependencies, and I am guessing that this will be a pure Ada project will only Ada code and using only packages from the standard Ada library.  So you don't have any benefit for yourself or the users to split the project.

> Edit: If you guys think it's better I can just cross that road when I reach it, ship `iso.ads` with `iso_countries`, and keep doing so until I encounter an iso library that doesn't require countries (and then submit `iso_base`?). I think that `iso_currencies` will rely on `iso_countries` but that may be it.

I strongly suggest that you just make one crate called `iso` and put all your `ISO` packages in this unique crate.

AJ-Ianozi:
> I strongly suggest that you just make one crate called `iso` and put all your `ISO` packages in this unique crate.

Alright, I'll work towards doing that.

PR #804, opened 2023-05-04
Releasing The Ada ISO Library

AJ-Ianozi:
This is the initial release for the [Ada ISO Library](https://github.com/ada-iso/ada_iso), which at the moment contains the implementation of ISO 3166-1 (Country Codes) with other implementations planned next (such as currencies).  This is the full library, with the countries implementation (as discussed on #803 ).

### DIFF
--- a/index/is/iso/iso-1.0.0.toml
+++ b/index/is/iso/iso-1.0.0.toml
@@ -1,0 +1,18 @@
+name = "iso"
+description = "ISO Standard references for Ada such as ISO 1366 country codes."
+version = "1.0.0"
+licenses = "MIT"
+website = "https://github.com/ada-iso/ada_iso/"
+tags = [ "countries", "iso-1366" ]
+
+authors = ["AJ Ianozi"]
+maintainers = ["AJ Ianozi <aj@ianozi.com>"]
+maintainers-logins = ["AJ-Ianozi"]
+
+[build-switches]
+"*".ada_version = "Ada2022"
+
+[origin]
+commit = "8c458179e6bc957f5219c0f9474123407cf947b4"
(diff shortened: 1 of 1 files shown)


AJ-Ianozi:
It doesn't seem to like
```toml
[build-switches]
"*".ada_version = "Ada2022"
```

Second one failed here:
```
Note: Building iso/iso.gpr...
Setup
   [mkdir]        object directory for project ISO
   [mkdir]        library directory for project ISO
Compile
   [Ada]          iso_config.ads
   [Ada]          iso-countries.adb
gnat1: invalid switch: -gnat[202](https://github.com/alire-project/alire-index/pull/804/checks#step:6:203)2

   compilation of iso-countries.adb failed
   compilation of iso_config.ads failed
```

I am using ada2022 features in this library, let me see what other packages are doing to handle this.

EDIT:
Since I'm not really utilizing Ada 2022 in this library yet (aside from an array aggregate), I opted to just make the library compatible with Ada2012 (the aggregate was changed into an extended return statement).  In the future if the library must utilize Ada2022, I'll be sure to opt for `pragma Ada_2022;` and specify the gcc requirement rather than use a compiler switch, which is probably the right way of doing it anyway.

AJ-Ianozi:
Since I'm not really utilizing Ada 2022 in this library yet (aside from an array aggregate), I opted to just make the library compatible with Ada2012 (the aggregate was changed into an extended return statement).  In the future if the library must utilize Ada2022, I'll be sure to opt for `pragma Ada_2022;` and specify the gcc requirement rather than use a compiler switch, which is probably the right way of doing it anyway.

Fabien-Chouteau (approved):

PR #817, opened 2023-05-29
Use gcc version instead or "year" version for GNAT CE external crate

mosteo:
I'm not entirely sure about this one because of legacy, although I guess this should have been like this from the beginning.

With this change, community versions report the gcc version instead of the year, which will always be a problem for things like `"gnat>=11"`. 

With the CE no longer being released, this may be a non-issue already...

### DIFF
--- a/index/gn/gnat_external/gnat_external-external.toml
+++ b/index/gn/gnat_external/gnat_external-external.toml
@@ -6,8 +6,8 @@ maintainers-logins = ["mosteo"]
 
 [[external]]
 kind = "version-output"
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau (commented):
Won't this detect something even when just GCC is installed and not GNAT?

mosteo:
> Won't this detect something even when just GCC is installed and not GNAT?

It shouldn't because of the "for GNAT" in the regular expression (so it would result in no detection). But anyway, this won't work until the next alr release, because the current stable raises in exactly the case you describe (a gcc that's not from gnat) and that requires a code fix and index version bump. So in the freezer it goes...

PR #818, opened 2023-05-30
AdaCL module adacl release 5.13.0

krischik:
* use `gnatpp` to format the source.
* use `gnatdoc` to create documentation.
* Add unique pointer implementation based on the C++ `unique_ptr`
* Add shared pointer implementation based on the C++ `shared_ptr`
* Cleanup the old refcounted pointer so that the calls match.


krischik:
Needs a fix.

PR #827, opened 2023-06-03
Create lea-0.87.0.toml

zertovitch:
Added the LEA application

### DIFF
--- a/index/le/lea/lea-0.87.0.toml
+++ b/index/le/lea/lea-0.87.0.toml
@@ -1,0 +1,43 @@
+description = "LEA is a Lightweight Editor for Ada "
+name = "lea"
+version = "0.87.0"
+
+authors = ["Gautier de Montmollin"]
+maintainers = ["gdemont@hotmail.com"]
+maintainers-logins = ["zertovitch"]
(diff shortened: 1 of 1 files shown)


mosteo (approved):

mosteo:
It seems it is not building on Windows because of the compiler dependency. Should I still merge it like this?

zertovitch:
On a local machine, the crate proposal builds seamlessly. I had to set a maximum to the gnat version because of a subtle incompatibility in gcc on an intrinsic function (see https://github.com/zertovitch/gwindows/issues/9 ). But recent gcc versions seem to have removed this glitch (notably, GNAT Pro 23.0).
I will refine the criterium and perhaps move it to the gwindows crate later, after probably a couple of iterations and when I have a clearer picture which versions are impacted.

zertovitch:
Short answer: go for it!

zertovitch:
Probably the issue with the compiler version will disappear from itself in an upcoming GNAT version in Alire since LEA builds already OK with GNAT Pro 23.

mosteo:
Alright, merging then!

PR #830, opened 2023-06-14
Mage v0.6.0: new Apps API

leogermond:
- mage 0.6.0
- mage_hat v0.3.0

This adds a new API to the mage lib, to simplify creating apps, on the format of a processing or arduino main loop definition.

This also provides a new version for the mage hat demo, which makes use of this new API.


mosteo:
The error is caused by the release of the new Debian stable. On it.

mosteo (approved):

PR #833, opened 2023-06-21
makefile release 1.0.0

pmunts:
This crate adds a useful `Makefile` to your Alire program project.

```
alr init --bin foo
cd foo
alr with makefile
```

Then:

```
make build # Build program and optionally scp executable file(s) to a target computer
make clean  # Clean the program project but not dependencies
make reallyclean # Clean program project and dependencies
make distclean # Return program project to pristine state
```

### DIFF
--- a/index/ma/makefile/makefile-1.0.0.toml
+++ b/index/ma/makefile/makefile-1.0.0.toml
@@ -1,0 +1,24 @@
+name = "makefile"
+description = "Add Makefile to your Alire project directory"
+tags = ["make", "makefile", "muntsos"]
+version = "1.0.0"
+website = "https://github.com/pmunts/libsimpleio"
+
+authors = ["Philip Munts"]
+maintainers = ["Philip Munts <phil@munts.net>"]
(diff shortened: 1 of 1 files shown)


index/ma/makefile/makefile-1.0.0.toml
```diff
@@ -0,0 +1,24 @@
+name = "makefile"
+description = "Add Makefile to your Alire project directory"
+tags = ["make", "makefile", "muntsos"]
```
mosteo, line 3: Hi Philip, is this specific to MuntsOS? (asking because I see the tag). If that's the case, I guess the crate name would be better `muntsos_makefile`.

krischik:
@pmunts Interesting. I too use makefiles to ease up the usage of Alire. Will this work with any operating system?

My makefiles tend to have a few more command like

* `make run`     # run the executable
* `make test`    # build and run the AUnit tests
* `make pretty` # use gnat pretty to pretty print
* `make doc`    # use gnat doc to create documentation.

PR #835, opened 2023-06-25
nael 0.1.0

Fabien-Chouteau:


Fabien-Chouteau:
On Arch the issue is with the configuration of GtkAda:
```
checking for GTK+ - version >= 3.24.24... no
*** Could not run GTK+ test program, checking why...
*** The test program failed to compile or link. See the file config.log for the
*** exact error that occurred. This usually means GTK+ is incorrectly installed.
```
However the arch package `gtk3 2.24.38-1` is installed so I don't know what is going on there. Anyway the issue is not with the `nael` crate,

On Fedora the issue is missing C++ compiler:
```
gprconfig: can't find a native toolchain for language 'c++'
```
We don't have ways to express that at the moment. For Alire provided toolchain it is not an issue since g++ is included.

mosteo:
I've updated our Fedora image so it has g++ like the rest. There's a new error at https://github.com/alire-project/alire-index/actions/runs/5368657575/jobs/9970419503#step:6:2446 

It might be a missing dependency, but I'm not sure, so if you can check it out...

PR #847, opened 2023-07-15
Honki Tonks Zivilisationen update to Version 0.05.3220

HonkiTonk:


mosteo:
Has anything changed that might be causing the linking error on [Arch](https://github.com/alire-project/alire-index/actions/runs/5560037444/jobs/10210140497#step:6:1260)?

HonkiTonk:
I think this is an Arch problem. Recently SFML 2.6 was released, but there is no CSFML 2.6. yet. Therefore Arch seems to install SFML 2.6 and CSFML 2.5 and then CSFML ends up looking for the 2.5 versions of SFML. But this is not available.

PR #853, opened 2023-07-20
Update ini_files-10.0.0.toml

zertovitch:
Added an executable, so that `alr run` doesn't end in `ERROR: Executable "ini_files.exe" not found`


mosteo (approved):

mosteo:
Just commenting that libraries don't necessarily have to provide any executable, so `alr run` not working would be expected.

PR #855, opened 2023-07-25
sqlite_ada 1.0.1

nytpu:
Add thick(-ish) bindings to SQLite: https://git.sr.ht/~nytpu/sqlite-ada

One note: I could change `SQLITE_USE_EXTERNAL` to default to `true` instead of `false` and add [libsqlite3](https://github.com/alire-project/alire-index/blob/stable-1.2.1/index/li/libsqlite3/libsqlite3-external.toml) as a dependency if you all would prefer, but the primary motivation behind writing these bindings from scratch was to allow customizing [SQLite compile options](https://sqlite.org/compile.html) and statically link extensions, which makes the statically linked SQLite amalgamation more appealing than using the system's dynamic library (also makes the library completely free of external dependencies from an Alire perspective)

### DIFF
--- a/index/sq/sqlite_ada/sqlite_ada-1.0.1.toml
+++ b/index/sq/sqlite_ada/sqlite_ada-1.0.1.toml
@@ -1,0 +1,39 @@
+name = "sqlite_ada"
+description = "Bindings to the SQLite 3 RBDMS"
+version = "1.0.1"
+licenses = "MPL-2.0"
+
+project-files = ["gnat/sqlite_ada.gpr"]
+
+website = "https://git.sr.ht/~nytpu/sqlite-ada"
+tags = ["database", "sql", "sqlite"]
+
+authors = ["nytpu"]
+maintainers = ["nytpu <alex@nytpu.com>"]
+maintainers-logins = ["nytpu"]
+
+[gpr-externals]
+SQLITE_LIBRARY_TYPE = ["dynamic", "relocatable", "static", "static-pic"]
+SQLITE_COMPILE_CHECKS = ["enabled", "disabled"]
+SQLITE_RUNTIME_CHECKS = ["enabled", "disabled"]
+SQLITE_STYLE_CHECKS = ["enabled", "disabled"]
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Thanks @nytpu 

> One note: I could change `SQLITE_USE_EXTERNAL` to default to `true` instead of `false` and add [libsqlite3](https://github.com/alire-project/alire-index/blob/stable-1.2.1/index/li/libsqlite3/libsqlite3-external.toml) as a dependency if you all would prefer,

I don't really have an opinion on this, external may also bring security updates instead of a fixed version of libsqlite, on the other hand I like the fact that this is self-contained.

What I would recommend however is to re-generate the `alire.toml` and `sqlite_ada.gpr` with the latest version of Alire:
`alr init --lib sqlite_ada`. 
We moved away from all the all the external variables (`SQLITE_COMPILE_CHECKS`, `SQLITE_RUNTIME_CHECKS`, `SQLITE_STYLE_CHECKS`, etc.) and now use a more simple and clear build switch system. 

I also recommend to avoid extra project files such as `example.gpr` because if another crate does the same `gprbuild` will complain about multiple projects having the same name and that makes the crate unusable. To provide examples it's better to make a "sub-crate" in your repo, like this for instance: https://github.com/wee-noise-maker/nael/tree/main/examples

nytpu:
@Fabien-Chouteau
I'll work on making those changes this weekend.

I like to keep my libraries trivially usable outside of Alire but most of the changes recommended every time I submit a new library here are very incompatible with that goal; so I might look into doing a whole separate Alire release from a different branch…

Fabien-Chouteau:
You can have a checked-in version of the config project file that will then be replaced by the Alire generated one for people using Alire. Like we do in here: https://github.com/alire-project/clic/blob/main/config/clic_config.gpr

Best of both worlds.

mosteo:
Is this still in draft mode on purpose?

nytpu:
@mosteo yeah, I've been busy with some projects and haven't been able to do the (rather substantial) changes needed. I'll competely close it and make a new one once it's ready.

PR #863, opened 2023-08-03
Mosquitto 1.0.2

persan:
Bindings to mosquitto


mosteo:
Shouldn't there be some dependency on `libmosquitto`?

mosteo:
Hi @persan , I've taken a look and indeed attempting to use this library will fail at link time if `libmosquitto` isn't installed. To make this PR mergeable, an external definition for `libmosquitto` is needed, you can base it on e.g. https://github.com/alire-project/alire-index/blob/stable-1.2.1/index/li/libcsfml/libcsfml-external.toml

It's not necessary to include all distros, it's enough to start with the one you use, and people can contribute the rest once they need them. Also, you can add this new file to this PR instead of making a separate one.

mosteo:
@persan, if you don't have the time to add the missing external, I'll do when I have some time. Commenting just to remove the stale countdown.

PR #864, opened 2023-08-04
Honki Tonks Zivilisationen update to Version 0.05.3685

HonkiTonk:


mosteo:
I tried updating the Arch image but nothing changed.

HonkiTonk:
If I see this correctly, Arch has SFML 2.6, but there is no CSFML version for SFML 2.6 yet. So I guess I will have to wait with further updates for Alire until the CSFML is updated.

mosteo:
If the dependencies are properly set, instead of a failure we should see Arch being skipped due to unfulfilled dependencies. But according to https://github.com/alire-project/alire-index/actions/runs/5767438049/job/16072066304?pr=864#step:6:289 Arch has everything needed?

HonkiTonk:
Yes, but here Arch installs [SFML 2.6](https://github.com/alire-project/alire-index/actions/runs/5767438049/job/16072066304?pr=864#step:6:399) and here [CSFML 2.5](https://github.com/alire-project/alire-index/actions/runs/5767438049/job/16072066304?pr=864#step:6:412).
[At the end it also searches for libsfml 2.5, but cannot find it.](https://github.com/alire-project/alire-index/actions/runs/5767438049/job/16072066304?pr=864#step:6:1239) So in my opinion the problem is not that Arch doesn't have everything it needs, but that Arch has a newer version of SFML which is not compatible with the current version of CSFML. Or am I getting something completely wrong?

mgrojo:
You're right, @HonkiTonk; something is broken in Arch Linux. I don't know how Arch works, but I don't see any dependency to specific versions in the [CSFML package](https://archlinux.org/packages/extra/x86_64/csfml/). If they are supposed to maintain all the available versions working together without relying on version specifications, they have failed for CSFML (the C binding), because the latest version, 2.5.2 (which they have), is [not compatible to the latest version of SFML](https://github.com/SFML/CSFML/issues/187) itself (the original C++ library), which is already in version 2.6.0 (the version they're also using).

mosteo:
Thanks for the feedback, @HonkiTonk, @mgrojo. Then I would suggest for the time being to disable Arch in the manifest and go ahead.

ChrisThrasher:
SFML maintainer here. You can use CSFML 2.5 with SFML 2.6.

mgrojo:
> SFML maintainer here. You can use CSFML 2.5 with SFML 2.6.

Thanks for letting us know.

Then the problem is not related to dependency management, but build management. I don't know how Arch works, but in some way, CSFML 2.5 is not being built with SFML 2.6, since it still references 2.5 libraries. 

https://github.com/alire-project/alire-index/actions/runs/5767438049/job/16072066304?pr=864#step:6:1239
```
/usr/sbin/ld: warning: libsfml-system.so.2.5, needed by /usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../lib/libcsfml-system.so, not found (try using -rpath or -rpath-link)
/usr/sbin/ld: warning: libsfml-audio.so.2.5, needed by /usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../lib/libcsfml-audio.so, not found (try using -rpath or -rpath-link)
/usr/sbin/ld: warning: libsfml-network.so.2.5, needed by /usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../lib/libcsfml-network.so, not found (try using -rpath or -rpath-link)
/usr/sbin/ld: warning: libsfml-graphics.so.2.5, needed by /usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../lib/libcsfml-graphics.so, not found (try using -rpath or -rpath-link)
/usr/sbin/ld: warning: libsfml-window.so.2.5, needed by /usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/../../../../lib/libcsfml-graphics.so, not found (try using -rpath or -rpath-link)
```

It might be solved already, though, but [no bugs found](https://bugs.archlinux.org/?project=1&string=csfml).

PR #868, opened 2023-08-10
Release/adacl 5.14.1

krischik:
* More generic support assertions
    * generic  for arrays types
    * generic  for discrete types
    * generic  for access types
* Debug and fix the hexdump feature of the trace module


krischik:
Why did macOS not build?

krischik:
Now macOS as successful and arch and fedora failed.

mosteo:
It seems warnings are enabled as errors? And the compiler on Arch/Fedora is emitting some:

https://github.com/alire-project/alire-index/actions/runs/5916858950/job/16044593546?pr=868#step:6:512

krischik:
> It seems warnings are enabled as errors? And the compiler on Arch/Fedora is emitting some:

I certainly didn't enable warnings as errors. That's in the Alire system. And quite honestly I hate it. Warnings are warnings for a reason and they should not be treated as errors. And I also have no idea how to switch that off. It's very difficult to switch something off you did not switch on in the first place.

krischik:
Interesting:

```ada
   ---
   --  Trace the given exception details and then raise the exception.
   --
   --: @param Raising  Exception which is raised
   --: @param Message  Free form Message
   --: @param Entity   Location descriptor. Suggested content: AdaCL.Trace.Entity
   --: @param Source   Location descriptor. Suggested content: AdaCL.Trace.Source
   procedure Raise_Exception
      (Raising : in Ada.Exceptions.Exception_Id := Ada.Assertions.Assertion_Error'Identity;
       Message : in String                      := "No Message given";
       Entity  : in String                      := GNAT.Source_Info.Enclosing_Entity;
       Source  : in String                      := GNAT.Source_Info.Source_Location)
   is
      use Ada.Exceptions;
   begin
      Write ("Raise Exception " & Exception_Name (Raising));
      Write ("   with Message " & Message);
      Write ("   for Entity   " & Entity);
      Write ("   in Source    " & Source);

      Raise_Exception (E => Raising, Message => Message & " Entity :" & Entity & "." & " Source :" & Source & ".");
      --
      --  GNAT designer forgot to add pragma No_Return to Ada.Exceptions.Raise_Exception.
      --
      raise Constraint_Error;
   end Raise_Exception;
```   

Was this finally fixed after almost 20 years and Linux got a newer compiler then macOS?

mosteo:
It's not enabled by default in our test builds, as everything should be built in release mode. It may be coming from some legacy-generated project files. I will need some more time to check this in detail if you don't find it first.

krischik:
Tried:

```ada
   pragma Warnings (Off, Raise_Exception);
```

But that didn't work either. I try removing the extra raise. It's an old workaround any way.

krischik:
Always arch-rolling and fedora-latest. Why are they creating larger log files then anybody else? Now I have to weaken my unit tests because those OS can't behave like the other Linux systems.

simonjwright:
From the point of view of the dead code warning, GCC 13 recognises Ada.Exceptions.Raise_Exception as having No_Return and reports the warning, while GCC 11 & 12 don’t report the warning.

However, a different part of 11 & 12 recognises that Raise_Exception doesn’t actually return, because if you put No_Return on your procedure and delete the final CE it doesn’t complain.

That said, I was bitten a long time ago by someone who turned off style warnings on his own machine; when checked in for a real build, the blizzard of warnings masked real issues in the code from reviewers, so I support the Alire project’s position on this.

krischik:
> That said, I was bitten a long time ago by someone who turned off style warnings on his own machine; when checked in for a real build, the blizzard of warnings masked real issues in the code from reviewers, so I support the Alire project’s position on this.

I don't mind the warnings and my code is usually warning free. But having warnings as error when the CI system uses different compiler is a pain.

mosteo:
Should this one be merged even with `5.15.1` being already merged?

krischik:
> Should this one be merged even with `5.15.1` being already merged?

Doesn't hurt and I already have the tar file to download on SourceForge. And the project looks more active. 😉 But at the end it is your decision as all the code is in 5.15.1 as well.

mosteo (approved):

PR #870, opened 2023-08-21
Release/adacl 5.15.1

krischik:
String utilities now also in Wide and Wide_Wide.


krischik:
Arch und Fedora — again 😞

krischik:
Always arch-rolling and fedora-latest. Why are they creating larger log files then anybody else? Now I have to weaken my unit tests because those OS can't behave like the other Linux systems.

mosteo (approved):

PR #886, opened 2023-08-26
mold_lib 0.2.0-dev

rocher:
Template processor library.

### DIFF
--- a/index/mo/mold_lib/mold_lib-0.2.0-dev.toml
+++ b/index/mo/mold_lib/mold_lib-0.2.0-dev.toml
@@ -1,0 +1,82 @@
+name = "mold_lib"
+version = "0.2.0-dev"
+tags = ["template", "template-engine", "template-repo"]
+
+description = "Meta-variable Operations for Lean Development (lib)"
+long-description = '''
+[![Alire](https://img.shields.io/endpoint?url=https://alire.ada.dev/badges/mold_lib.json)](https://alire.ada.dev/crates/mold_lib.html)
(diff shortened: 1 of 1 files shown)


mosteo (approved):

index/mo/mold_lib/mold_lib-0.2.0-dev.toml
```diff
@@ -0,0 +1,82 @@
+name = "mold_lib"
+version = "0.2.0-dev"
```
mosteo, line 2: Is the `-dev` on purpose? Usually it's unexpected in submissions to the community index.

rocher: It's a kind of pre-release, but I can remove it if not appropriate.

mosteo: It's up to you, really. The thing is, the pre-release nature is already kind of implied for all `0.x` versions.

Just tell me if I wait or I go ahead with the merge.

rocher: In case that another crate want to use it, the dependency `mold_lib = ">=0.2.0"` would find `0.2.0-dev` as a valid version?  Anyway, I prefer to change it.
Will send another PR soon, thanks for the review.

mosteo: That's one of the things, per semver definition, all pre-release versions are earlier than the corresponding regular one. So no, neither `>=0.2` nor `~0.2` would find a `0.2-something`.

For the future, you don't need to close the PR, you can just update the files you want to tweak if you prefer.

PR #888, opened 2023-08-27
Ada Utility Library AWS 2.6.0 crate

stcarrez:


stcarrez:
See https://github.com/alire-project/alire/issues/1394 for the explanation of the issue on Debian and Fedora: issue in AWS `TARGET` definition.

PR #904, opened 2023-09-30
Add OpenAI 1.0 crate

stcarrez:
The library provides an Ada binding to make requests on OpenAI and query GPT-4, GPT-3, DALL.E and other AI models.


mosteo (approved):

mosteo:
Nice! Out of curiosity, does it support getting the answers as they're generated (I think they call it chunked mode)?

stcarrez:
> Nice! Out of curiosity, does it support getting the answers as they're generated (I think they call it chunked mode)?

In fact no. This could be some improvement. If I'm not wrong, this is based on server-sent-events. It's not supported yet on OpenAPI and I don't know whether it is supported by the AWS client.

mosteo:
I know it's supported by the Python client, but the documentation was really scarce on it. I don't know how it works internally (I was playing with this a while ago):

```python
for chunk in openai.ChatCompletion.create(
            model=model,
            max_tokens=max_answer_tokens,
            messages=messages,
            stream=True
        ):
            token = chunk["choices"][0].get("delta", {}).get("content")
            if token is not None:
                print(token, end='')
```
So it looks like regular iteration on the client side.

PR #905, opened 2023-10-05
Release ada-mode 8.1.0, gpr-mode 1.0.4 and dependencies

stephe-ada-guru:


stephe-ada-guru:
The windows build is failing with: 

s-memory.adb:1:79: (style) incorrect line terminator

<rant>
It seems gnat 13 is now trying to enforce "correct" line terminators on files, which is just ridiculous; always accepting either Unix or DOS is far more friendly, and not at all harmful.
</rant>

According to the gnat 13 user guide gnat_ug.info, there is no way to turn off this check. 

I searched for this issue, and found https://github.com/AdaCore/Ada_Drivers_Library/issues/43, which says that the solution is a git config setting to get the line endings right. That has to be done in the github task that is running this check.

stephe-ada-guru:
I tried to reproduce the issue on my Windows box by deleting "autoclrf = false" from my ~/.gitconfig, but that had no effect; wisitoken crate built fine either way.

stephe-ada-guru:
I managed to reproduce the compiler error; it is reported when s-memory.adb has DOS line endings, which happens when ~/.gitconfig has autocrlf = true. Apparently the gnat style switch -gnatyd is set when compiling that file (it has different compiler switches than the other files). But that's not new in gnat 13, so this must have changed in the .gitconfig used for this test. I'll try adding a .gitattributes setting for that file.

stephe-ada-guru:
I managed to reproduce the compiler error; it is reported when s-memory.adb has DOS line endings, which happens when ~/.gitconfig has autocrlf = true. Apparently the gnat style switch -gnatyd is set when compiling that file (it has different compiler switches than the other files). But that's not new in gnat 13, so this must have changed in the .gitconfig used for this test. I tried adding .gitattributes to prevent git from messing with the line termination, but that did not work. So I added a pre-build action.

stephe-ada-guru:
Now the prebuild action that runs dos2unix is failing on non-Windows, probably because it is not installed? How do I make that action only on Windows?

stephe-ada-guru:
Fixed the line ending problem. Now Windows is failing when it runs the ada-mode pre-build function, which runs wisitoken-bnf-generate, which can take a large amount of memory. So I'm guessing the failure is due to limited memory. The other builds succeed; do they have more memory, or is Windows just less efficient? In either case, the best solution is to increase the memory for that machine. I could do a workaround where I store the results of that run on Savannah, and change the pre-build action to just copy them, but that's a kludge.

mosteo:
If you're confident that things are OK, we can merge as-is, let me know if so. 

As for a permanent solution, I will check if we're able to increase memory of Windows VMs.

mosteo:
The free Windows machines have 7GB of RAM. Anything beyond requires pay-per-use :S

stephe-ada-guru:
Please merge; everything works fine on my Windows machine

stephe-ada-guru:
Please merge; everything works fine on my Windows box.

PR #913, opened 2023-11-01
Don't look for gprbuild from system provider.

simonjwright:
Because the manifest currently asks for Alire to try to download gprbuild from the system package manager, alr <anything> spends several fruitless seconds trying to find gprbuild. This PR removes the 'system' external from gprbuild-external.toml.


simonjwright:
I see that Debian and Msys _do_ provide gprbuild. Should I just make the load conditional on the "os" _not_ being macOS? or the "distribution" not being HomeBrew, MacPorts? (not sure how to do either)

mosteo:
For the time being I think it's OK to use the OS for the conditional.

mosteo (approved):

PR #919, opened 2023-11-12
Create globe_3d-2023.11.12.toml

zertovitch:


zertovitch (commented):
Fixed directory of project file

zertovitch:
The checker borks for some reason... At least the issue seems to be OS-neutral:
ALIRE.CHECKED_ERROR
Windows: Unexpected contents where a single directory was expected: D:\a\alire-index\alire-index\alr-cvqj.tmp
Ubuntu: Unexpected contents where a single directory was expected: /home/runner/work/alire-index/alire-index/alr-kvvp.tmp

Fabien-Chouteau:
@zertovitch at the root of your tarball there should be only a directory with the sources in it. Here you have multiple files and directories at the root.

zertovitch (commented):
New Zip file, new hash.

zertovitch:
Looks definitely better!
Current state: Windows, Mac: OK, Linuxes: KO.
On Linuxes, Alire seems to build imported projects GID and Zip-Ada, but not to find them after that.

PR #924, opened 2023-11-23
Match more version of gnat/gprbuild when using external versions

rowan-walshe:
This expands the list of GNAT/GPRBUILD versions that can be matched by the external crates. For example the version of gprbuild that is installed by apt on Ubuntu 22.04


mosteo (approved):

mosteo:
I see the problem, thanks @rowan-walshe

PR #934, opened 2023-12-13
Ada Binding to Gnu Scientific Library

RajaSrinivasan:
Thinnish binding to gsl 2.7 with examples

### DIFF
--- a/index/gs/gsl/gsl-0.1.0-dev.toml
+++ b/index/gs/gsl/gsl-0.1.0-dev.toml
@@ -1,5 +1,5 @@
 name = "gsl"
-description = "Shiny new project"
+description = "Ada binding to Gnu Scientific Library 2.7"
 version = "0.1.0-dev"
 
 authors = ["R Srinivasan"]
--- a/index/gs/gsl/gsl-0.1.0-dev.toml
+++ b/index/gs/gsl/gsl-0.1.0-dev.toml
@@ -1,0 +1,13 @@
+name = "gsl"
+description = "Shiny new project"
+version = "0.1.0-dev"
+
+authors = ["R Srinivasan"]
+maintainers = ["R Srinivasan <rajasrinivasan@hotmail.com>"]
+maintainers-logins = ["rajasrinivasan"]
+
+[origin]
+commit = "3b92de5827013378f45d92a47463efb9bdb16cea"
+subdir = "./gsl/"
+url = "git+https://gitlab.com/ada23/adagsl.git"
+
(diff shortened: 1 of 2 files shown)


index/gs/gsl/gsl-0.1.0-dev.toml
```diff
@@ -0,0 +1,13 @@
+name = "gsl"
+description = "Shiny new project"
```
Fabien-Chouteau: Hi @RajaSrinivasan, you should have a short description of your project here instead of the default `Shiny new project`.

Fabien-Chouteau (changes requested):

RajaSrinivasan:
Thanks. Not even sure how that came about! Fixed it. Hope it was
resubmitted. regards, srini

index/gs/gsl/gsl-0.1.0-dev.toml
```diff
@@ -0,0 +1,13 @@
+name = "gsl"
+description = "Ada binding to Gnu Scientific Library 2.7"
+version = "0.1.0-dev"
```
mosteo: This most likely wants to simply be `0.1.0` (with the filename fixed accordingly).

Separate issue: doesn't this depend on `libgsl`?

mosteo (changes requested):

index/gs/gsl/gsl-0.1.0-dev.toml
RajaSrinivasan: should I re start the whole process?  

yes. this does require libgsl. The gpr file(s) indicate appropriate switches for Linux and MAC. I don't test this on windows.

some guidance will be good. regards, srini

mosteo: No need to restart. Just amend the manifest in this PR and push the changes.

Since `libgsl` is already indexed, you need only to add the dependency on the proper version:
```
[[depends-on]]
libgsl = "^2.7"
```

mosteo:
Thanks, for the fixes, @RajaSrinivasan . This is almost ready, the single failure is due to the compiler in Ubuntu LTS being to old to recognize 2022 features. You need to add a compiler restriction such as
```
[[depends-on]]
gnat = ">=12"
```
It's possible GNAT 11 could work too, but I'm unsure. I saw that GNAT 12 succeeds in other checks so that's a safe choice. Up to you to dig a bit further.

mosteo (approved):

mosteo:
Thanks, @RajaSrinivasan

PR #935, opened 2023-12-13
pla_util 2.1.1

serock:
Initial release of pla_util crate.

### DIFF
--- a/index/pl/pla_util/pla_util-2.1.1.toml
+++ b/index/pl/pla_util/pla_util-2.1.1.toml
@@ -1,0 +1,64 @@
+#  SPDX-License-Identifier: GPL-3.0-or-later
+#----------------------------------------------------------------------
+#  pla-util - A power line adapter utility
+#  Copyright (C) 2016-2023 John Serock
+#
+#  This file is part of pla-util.
+#
+#  pla-util is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  pla-util is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


index/pl/pla_util/pla_util-2.1.1.toml
```diff
@@ -0,0 +1,64 @@
+#  SPDX-License-Identifier: GPL-3.0-or-later
```
Fabien-Chouteau: Hi @serock, we don't expect copyright headers in release manifests. The authors and licenses of the projects are already documented in the manifest.

Fabien-Chouteau (changes requested):

index/pl/pla_util/pla_util-2.1.1.toml
serock: Hi @Fabien-Chouteau, thank you for reviewing; the header has been removed.

index/pl/pla_util/pla_util-2.1.1.toml
```diff
@@ -0,0 +1,44 @@
...
+"linux" = true
+"..." = false
+
+[[depends-on]]
+libpcap = "^1.8.1"
+
+[configuration]
+generate_c = false
+
+[build-profiles]
```
Fabien-Chouteau: Sorry I missed that the first time, but you must not set build profiles in a release manifest.

@mosteo I am surprised that the CI didn't catch that.

Fabien-Chouteau (changes requested):

index/pl/pla_util/pla_util-2.1.1.toml
mosteo: I don't think we have checks for that, as it is conceivably desirable for non-library crates. Making a note of it.

serock: Hi @Fabien-Chouteau , thanks again for reviewing; I removed the build profiles.

Fabien-Chouteau (approved):

PR #940, opened 2023-12-15
Added portable_minimal_ada_rts crate

jgrivera67:


mosteo:
The dreaded trouble with wrong line terminators strikes again. At first sight, you have the proper compiler flags, so I'm not sure what's going on right now.

index/po/portable_minimal_ada_rts/portable_minimal_ada_rts-1.0.0.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "1.0.0"
+licenses = "GPL-3.0-or-later WITH GCC-exception-3.1"
+website = "https://github.com/jgrivera67/portable_minimal_ada_rts"
+tags = ["rts"]
+authors = ["J. German Rivera"]
+maintainers = ["J. German Rivera <jgrivera67@gmail.com>"]
+maintainers-logins = ["jgrivera67"]
+
+[[depends-on]]
+gnat_arm_elf = "^12.1.2"
```
mosteo: Is this dependency on ARM intended?

jgrivera67:
> The dreaded trouble with wrong line terminators strikes again. At first sight, you have the proper compiler flags, so I'm not sure what's going on right now.

Thanks for catching this @mosteo. @Fabien-Chouteau kindly told me how to fix it:

"You are using the -gnatg switch here:
https://github.com/jgrivera67/portable_minimal_ada_rts/blob/a66019438515d482f8d12d6c9fe4003fe8f4327f/portable_minimal_ada_rts.gpr#L30

That's good because this switch is required to build a run-time project, however -gnatg also enables other switches (see here), in particular style checks that revert what you have set in ADAFLAGS. That's why you have build errors on Windows with the line termination warning. So you have to set -gnatg before the ADAFLAGS switches."

index/po/portable_minimal_ada_rts/portable_minimal_ada_rts-1.0.0.toml
jgrivera67: The code of this crate itself should be platform independent, but I needed to specify a compiler to be able to pass the alire-index PR checks. The client code of this crate, which lives in another crate, does depend on ARM.  However,
this crate should be compilable for any platform.

What would be a cleaner way to specify a compiler dependency for this minimal Ada RTS crate, given the fact it should compile on any platform?

jgrivera67:
> The dreaded trouble with wrong line terminators strikes again. At first sight, you have the proper compiler flags, so I'm not sure what's going on right now.

Thanks for catching this @mosteo. @Fabien-Chouteau kindly told me how to fix it:

"You are using the -gnatg switch here:
https://github.com/jgrivera67/portable_minimal_ada_rts/blob/a66019438515d482f8d12d6c9fe4003fe8f4327f/portable_minimal_ada_rts.gpr#L30

That's good because this switch is required to build a run-time project, however -gnatg also enables other switches (see here), in particular style checks that revert what you have set in ADAFLAGS. That's why you have build errors on Windows with the line termination warning. So you have to set -gnatg before the ADAFLAGS switches."

Now, my PR passes all 15 checks

index/po/portable_minimal_ada_rts/portable_minimal_ada_rts-1.0.0.toml
mosteo: A compiler dependency shouldn't be necessary, and given what you tell me, it isn't desirable as clients of your RTS may want to specify a different target. What was the problem without it?

jgrivera67: I get this error:
"Note: Building portable_minimal_ada_rts/portable_minimal_ada_rts.gpr...
gprconfig: can't find a toolchain for the following configuration:
gprconfig: language 'ada', target 'arm-eabi', runtime '/Users/runner/work/alire-index/alire-index/portable_minimal_ada_rts_1.0.0_99f278cb'
Setup
portable_minimal_ada_rts.gpr:11:25: warning: libraries are not supported on this platform
   [mkdir]        object directory for project Portable_Minimal_Ada_Rts
portable_minimal_ada_rts.gpr:4:09: no compiler for language "Ada", cannot compile "portable_minimal_ada_rts_config.ads"
   [mkdir]        library directory for project Portable_Minimal_Ada_Rts
ERROR: Command ["gprbuild", "-s", "-j0", "-p", "-P", "/Users/runner/work/alire-index/alire-index/portable_minimal_ada_rts_1.0.0_99f278cb/portable_minimal_ada_rts.gpr"] exited with code 4
stderr: Command ["gprbuild", "-s", "-j0", "-p", "-P", "/Users/runner/work/alire-index/alire-index/portable_minimal_ada_rts_1.0.0_99f278cb/portable_minimal_ada_rts.gpr"] exited with code 4
ERROR: Compilation failed.
stderr: ALIRE.CHECKED_ERROR"

I wonder if when building an RTS (-gnatg switch), an explicit toolchain needs to be specified.

jgrivera67: I tried adding a dependency on `gnat`, but the build for `macos-latest` fails with the following error:
```
Error: nodename nor servname provided, or not known (api.github.com:443)
```

mosteo: Mm. An explicit `gnat` shouldn't also be needed. I guess it's possible that, when compiling an RTS, things are a little out of the ordinary , but I lack the expertise wrt to cross-compiling. @Fabien-Chouteau will problably know.

> gprconfig: language 'ada', target 'arm-eabi',

That suggests to me that, even if you removed the dependency in the Alire manifest, there's a target set in some `.gpr`, which defeats the purpose.

If, once you make sure no references to targets are anywhere, the tests still fail, what you can do is create a test crate inside a `test` folder, that uses this crate through a ".." pin. That crate may use an explicit target. Then, CI checks will simply build that crate instead of this one. Although I'd like to see the fail log in that case.

jgrivera67: Thanks, you were right. I had a target clause in the .gpr file, which did not need to be there. I removed it and cleaned some warnings for Windws buids and I'm passing all 15 checks without needing the depedency on a specific compiler.

mosteo (approved):

mosteo:
Thanks, merged.

PR #942, opened 2023-12-18
Add elogs crate to the alire index

kevlar700:


upload/stable-1.2.1/index/el/elogs
```diff
@@ -0,0 +1,24 @@
...
+
+authors = ["Kevin Chadwick"]
+maintainers = ["Kevin Chadwick <kc-ada@elansys.co>"]
+maintainers-logins = ["kevlar700"]
+licenses = "ISC OR BSD-2-Clause"
+website = "https://github.com/kevlar700/elogs"
+tags = ["spark", "embedded", "zfp", "logging", "log"]
+
+[[depends-on]]
+gnatprove = "^13"
```
mosteo, line 13: To avoid pulling in the dependency on `gnatprove` for clients of your library, we recommend to move the proving to a nested crate that depends on the main crate. See e.g. the `tests` nested crate in the `vaton` crate [here](https://github.com/mhatzl/vaton).

mosteo (changes requested):

upload/stable-1.2.1/index/el/elogs
kevlar700: I could just remove the gnatprove dependency? 
I only left it in as I figured it would ensure that the SPARK menu would show up in Gnat Studio CR.

kevlar700: In fact my tests are not in Spark Mode. Isn't gnatprove cached in general anyway?

Fabien-Chouteau:
@kevlar700 I think you have to do another push to restart the checks.

PR #944, opened 2023-12-19
bme280 1.0.0

reznikmm:
Created via `alr publish` with `alr 2.0-beta1`

### DIFF
--- a/index/bm/bme280/bme280-1.0.0.toml
+++ b/index/bm/bme280/bme280-1.0.0.toml
@@ -1,0 +1,30 @@
+name = "bme280"
+description = "Driver for BME280 Humidity sensor"
+version = "1.0.0"
+
+authors = ["Max Reznik"]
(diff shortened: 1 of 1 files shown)


index/bm/bme280/bme280-1.0.0.toml
```diff
@@ -0,0 +1,34 @@
+# SPDX-FileCopyrightText: 2023 Max Reznik <reznikmm@gmail.com>
```
Fabien-Chouteau: Hi @reznikmm , we don't expect copyright headers in release manifests. The authors and licenses of the projects are already documented in the manifest.

Fabien-Chouteau (changes requested):

index/bm/bme280/bme280-1.0.0.toml
reznikmm: Thanks! Fixed.

Fabien-Chouteau (approved):

PR #945, opened 2023-12-19
Portable minimal rts fix

jgrivera67:
Updated sources to match gnat_arm_elf 13.2.1

### DIFF
--- a/index/po/portable_minimal_ada_rts/portable_minimal_ada_rts-1.3.0.toml
+++ b/index/po/portable_minimal_ada_rts/portable_minimal_ada_rts-1.3.0.toml
@@ -1,0 +1,15 @@
+name = "portable_minimal_ada_rts"
+description = "Portable Minimal Ada Runtime Library"
+version = "1.3.0"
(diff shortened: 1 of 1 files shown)


jgrivera67:
The following three PR required builds are failing due to older native compilers being used for the corresponding build platforms:

* The ubuntu-lts build seems to be failing with the following error:
```
system.ads:75:49: unrecognized attribute "Max_Integer_Size"
```
The problem here seems to be that the ubuntu-lts build is using an older GNAT compiler (10.5.0) that does not support the `Max_Integer_Size` attribute.

* The centos-fsf-stream-latest and the debian-stable builds are failing with the error below:
```
[Ada]          ada.ads
g-souinf.ads:42:07: error: state "Source_Code_Information" requires refinement
g-souinf.ads:42:07: error: package body should have Refined_State for state "Source_Code_Information" with constituents
```
Both the centos-fsf-stream-latest and the debian-stable builds use GNAT 12.2.0, which does not seem to support the construct causing the error.

mosteo:
Please resubmit to the new index base branch (`stable-1.3.0`) if still relevant.

PR #947, opened 2023-12-23
Honki Tonks Zivilisationen update to Version 0.05.5305

HonkiTonk:


mosteo:
There's some trouble with the linking of `asfml` on Arch. You can investigate the issue or disable availability for Arch.

mgrojo:
Arch includes now [CSFML v2.6.0](https://archlinux.org/packages/extra/x86_64/csfml/), and I've updated ASFML now to version v2.6.0, so there's a chance that the Arch version links again. If I don't understand incorrectly, this dependency `asfml = "^2.5.3"` will take 2.6 where csfml=2.6 is available.

mosteo:
Indeed it has worked, merging. Thanks for the heads-up!

mosteo (approved):

PR #957, opened 2024-01-05
elogs 1.1.0

kevlar700:


mosteo:
Hi @kevlar700 , the tests are not running because the submitted file is at the wrong path: `upload/stable-1.2.1/index/el/elogs`

It should be at `index/el/elogs-1.1.0.toml`.

You can move the file to the proper location, commit and push in this same PR.

mosteo:
Now the file has the correct name, but is still in the wrong location. The `upload/stable-1.2.1` should not be there, it should be directly start at `index`.

index/el/elogs-1.1.0.toml
mosteo, line 1: This file should be removed

upload/index/el/elogs-1.1.0.toml
mosteo, line 1: This file should be removed

upload/stable-1.2.1/index/el/elogs-1.1.0.toml
mosteo, line 1: This file should be removed

kevlar700: Sorry, the delete button was hidden. I didn't see the menu scroll bar for ages.

kevlar700:
What is the issue now @mosteo ?

mosteo:
Compilation is failing at the new `[]` arrays, you should enable Ada 2022 features with
```
[build-switches]
"*".ada_version = "Ada2022"
```
in the manifest.

kevlar700:
> in the manifest.

Do I need to remove -gnat2022 from here now? Does that require a new PR?
https://github.com/kevlar700/elogs/blob/main/config/elogs_config.gpr

Edit: I tried removing it but I see now that the failed LTS checks use a distro toolchain that doesn't support 2022. Not sure what is happening with alr dev.

A pretty cool set of checks :)

mosteo:
> Do I need to remove -gnat2022 from here now? Does that require a new PR?

The `config` dir is regenerated by `alr` on any changes to the manifest, so you can leave it as is. You could even leave it out of version control, although leaving it in allows the use of your project without Alire.

>  toolchain that doesn't support 2022

This means you need a minimum compiler version dependency. I'll add that one because I also have to deal with the error with compiling the master branch; bizarrely I suspect a PR intended to ensure wide compiler compatibility may have introduced that regression.

So I think you're done, thanks for the patience

mosteo:
Remaining checks should succeed after https://github.com/alire-project/alire/pull/1566 is merged

PR #966, opened 2024-01-18
Septum 0.8.0

pyjarrett:
Version bump for "atomic" dependency.


pyjarrett:
I made a mistake and actually made another (correct) PR instead.

PR #968, opened 2024-01-18
Use new Windows cache location for direct `pacman` call

mosteo:
Prompted by changes in https://github.com/alire-project/alire/pull/1528


mosteo:
In tandem with https://github.com/alire-project/alire/pull/1533, this should fix the issues introduced in https://github.com/alire-project/alire/pull/1528, which compounded with GHA cache management.

mosteo:
Merging already to reduce impact to submitters.

PR #976, opened 2024-01-25
Add macOS support to libreadline.

simonjwright:
"brew info readline" shows that the library is "keg-only", i.e. it's not symlinked into Homebrew's normally-visible include/, lib/ folders. This doesn't matter for the main usage in Alire (gnatcoll_readline), because it's only requirement is to link against "-lreadline". On macOS this is provided by BSD libedit.

In that case, why bother installing it? To stop alr warning "Generating possibly incomplete environment because of missing dependencies".


simonjwright:
I think the (alr dev) / macos_latest failure was a momentary glitch.

mosteo (approved):

PR #977, opened 2024-01-26
Update gid-11.0.0.toml (only changed description fields)

zertovitch:
Added long description


zertovitch (commented):
Added missing footnote.

Fabien-Chouteau (approved):

PR #978, opened 2024-01-27
rp2040_hal 2.2.1

JeremyGrosser:


JeremyGrosser:
I don't understand why the CI runs are still failing. It looks like it's trying to use gnat_native when this crate can only be built with gnat_arm_elf, which is listed in the dependencies.

mosteo:
This is likely related to https://github.com/alire-project/alire/pull/1533. Sorry for the trouble, I'll take a look ASAP.

Fabien-Chouteau:
I can see a warning during the build of tests:
```
aunit-memory.adb:35:06: warning: renamed predefined unit is an obsolescent feature (RM J.1) [-gnatwj]
```
I don't know if that's the problem.

mosteo:
I can reproduce this problem locally, working on the fix.

mosteo:
There are two issues here: one is some mixing of github actions caches, which was causing the apparent mixing of tools. I'm working on that.

The other one is the one pointed by Fabien, which seems legit, see e.g. this run: https://github.com/alire-project/alire-index/actions/runs/7681838264/job/20986832061?pr=978#step:6:825

Fabien-Chouteau:
My bad, the warning is the issue here [1]. The problem is that the check script is trying to run the test, which is an ARM Cortex-M application for the RP2040. 

The scripts assumes that the tests are native applications and that `alr run` should work.

[1] Aunit is rightfully in not compiled with warnings as error.

mosteo:
So, IIUC, these tests are not intended for execution here, and this release can be merged?

JeremyGrosser:
Correct, these tests can only run on an RP2040. I do think there is value in compiling the tests as a part of CI, but we don't have a good simulator for running them outside of real hardware.

This release can be merged, imo.

mosteo:
Alright, then. After 2.0 the CI script can be updated to rely only on `alr test`, and that should get rid of the issue.

PR #984, opened 2024-02-05
Create ulid-20240205.0.0.toml

zertovitch:


zertovitch (commented):
Fixed glitches in tags

zertovitch:
Don't worry about the failed Ubuntu test, it's just that that compiler version doesn't support 128 bit integers.
"modulus exceeds limit (2 ** 64)"

The ULID crate depends only on pure Ada stuff:

```
with Ada.Calendar.Formatting;
with Ada.Calendar.Time_Zones;
with Ada.Numerics.Discrete_Random;
with Ada.Text_IO;
with Interfaces;
```

mosteo:
> it's just that that compiler version doesn't support 128 bit integers.

Should there be a minimum compiler version dependency then?

zertovitch:
Good point! Any suggestion?
I am a bit rusty regarding how to formulate correctly the GNAT version number.
I sort of remember there was a debate of "year" versus "gcc version".
Anything including GNAT CE 2021 (but not too much more in the past time) should be good...

mosteo:
From the test runs it seems GNAT 10 is the trouble. We are using GNAT versions, not gcc versions, so that should be:

`gnat = "(>=11 & <2000) | >=2021"`

Edit: I've added it.

mosteo (approved):

PR #987, opened 2024-02-09
Add A0B Base&ARMv7M and update light-startup crates.

godunko:


godunko:
Not good for common index... Sorry.

PR #997, opened 2024-02-15
si_units 0.2.2

mosteo:
Created via `alr publish` with `alr 2.0-beta1`


mosteo:
Closed using `alr 2.0-beta1` with reason: Failed own testsuite

PR #1008, opened 2024-03-09
Add ada_toml 0.4.0

pmderodat:


index/ad/ada_toml/ada_toml-0.4.0.toml
```diff
@@ -0,0 +1,110 @@
...
+rules](https://github.com/pmderodat/ada-toml/tree/master/CONTRIBUTING.rst) if
+you consider doing so.
+"""
+
+[gpr-externals]
+ADA_TOML_BUILD_MODE = ["dev", "prod"]
+LIBRARY_TYPE = ["static", "relocatable", "static-pic"]
+
+[gpr-set-externals]
+ADA_TOML_BUILD_MODE = "prod"
```
mosteo: Hi @pmderodat, this should be left unset so clients of the library can define it. You can of course use that as default in your gpr file, so clients are not forced to choose either. I realize now this is inherited from previous releases, so I apologize for previous oversights.

mosteo (changes requested):
See inlined comment.

index/ad/ada_toml/ada_toml-0.4.0.toml
pmderodat: Hello, thank you for the review! No problem :-)

mosteo (approved):

mosteo:
Merged, thanks.

pmderodat:
Great, thank you very much!

index/ad/ada_toml/ada_toml-0.4.0.toml
Fabien-Chouteau: Hi, I don't agree here ^^

The default value in `ada_toml.gpr` is `dev`.

Most people will just want to use this library and not work on it, so it's better to default to performance and lower binary size. 

Also the "dev" mode enables warnings as errors, so there's a good chance this library won't compile with future compilers that will introduce new warnings.

So for me, it's important to either:
 - Change the default in `ada_toml.gpr`
 - Set the build mode in the release manifest
 - Use the Alire crate config system

mosteo: Mmm, so you mean that users should not have the option to even enable `dev`? Because if set to `prod` with `gpr-set-externals`, any attempt to change it will cause a conflict.

I didn't check the default, that makes sense that it be `prod`.

I don't see why the config system is needed here.

pmderodat: FTR I would prefer that the default in the GPR file stays `dev`, because it’s the most convenient in developer mode, which is what I would expect most `git clone` consumers to do. It would be really nice if Alire was able to have its own default in its spec, and then let Alire users override it if necessary.

Fabien-Chouteau: I that case I strongly recommend to bring back the "prod" build mode in the 0.4.0 release manifest
```toml
[gpr-set-externals]
ADA_TOML_BUILD_MODE = "prod"
```

PR #1009, opened 2024-03-10
alr_appimage 0.9.0

mgrojo:
Created via `alr publish` with `alr 2.0-rc1`

### DIFF
--- a/index/al/alr_appimage/alr_appimage-0.9.0.toml
+++ b/index/al/alr_appimage/alr_appimage-0.9.0.toml
@@ -1,0 +1,44 @@
+name = "alr_appimage"
+description = "Tool to create an AppImage executable from an Alire crate"
+version = "0.9.0"
+website = "https://github.com/mgrojo/alr-appimage"
+long-description = """
+There are two prerequisites for your project to work with this tool:
+- It has to be a crate with an `executables` field. Its first value
+  has to be the main application program.
+- It must be installable using Alire, including all the needed resources.
+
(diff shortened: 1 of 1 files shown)


mgrojo:
The errors are:
```
ERROR: /Users/runner/work/alire-index/alire-index/index/index.toml: index version (1.3.0) is newer than that expected by alr (1.2.1). You may have to update alr
```
I guess it is something with the imminent release and using the rc1, but I don't know what should I do to fix it. Maybe use the beta, as the last time, or doing it manually; so I'll wait for advice.

mosteo:
Submissions against index version `1.3.0` are not yet working; the branch exists but it is not the default. If you reopen against base branch `stable-1.2.1`, it should work.

mosteo:
Ah, but your crate requires `2.0`. Then I'd say just wait until it gets released, as we don't plan on having a devel index branch that would allow testing this in advance to release.

mgrojo:
In fact, my original idea for this tool was to use this command for the installation in an AppDir directory:
```
alr exec -P -- gprinstall --prefix=AppDir --create-missing-dirs --mode=usage -f
```
But after `alr install` I thought it would make more sense to simply use that feature. Now I'm wondering if supporting Alire 1.x with that command after a failed attempt to run `alr install` would make sense. I guess the result would be the same.

mosteo:
Yup, our gprinstall call looks pretty much the same, so using that should work too.

PR #1010, opened 2024-03-10
Create echada-1.0.0.toml

PalsFreniers:


mosteo (changes requested):
Hi @PalsFreniers, the file is not at the proper location.

Current location: `upload/stable-1.2.1/index/ec/echada/echada-1.0.0.toml`
Correct location: `index/ec/echada/echada-1.0.0.toml`

PR #1011, opened 2024-03-12
alr_appimage v0.9.1 compatible with Alire 2.x and Alire 1.x

mgrojo:


mgrojo:
For windows-latest check, it seems that you have to change:

[C:\Users\runneradmin\AppData\Local\alire\msys64\usr\bin\pacman](https://github.com/alire-project/alire-index/actions/runs/8255493504/job/22582072725?pr=1011#step:4:442)

by

[C:\Users\runneradmin\AppData\Local\alire\cache\msys64\usr\bin\pacman](https://github.com/alire-project/alire-index/actions/runs/8255493504/job/22582072725?pr=1011#step:5:8)

On the other hand, I didn't expect it to run after the commit changing the availability to be true only on Linux.

mosteo:
> For windows-latest check, it seems that you have to change:
> 
> [C:\Users\runneradmin\AppData\Local\alire\msys64\usr\bin\pacman](https://github.com/alire-project/alire-index/actions/runs/8255493504/job/22582072725?pr=1011#step:4:442)
> 
> by
> 
> [C:\Users\runneradmin\AppData\Local\alire\cache\msys64\usr\bin\pacman](https://github.com/alire-project/alire-index/actions/runs/8255493504/job/22582072725?pr=1011#step:5:8)

Right, thanks for pointing it out.

> On the other hand, I didn't expect it to run after the commit changing the availability to be true only on Linux.

Alire must be prepared in advance to be able to check whether the crate is available on the platform, but I see your point.

mosteo (changes requested):
I have a concern here which is that the name may imply that this comes from the alire-project itself. Also we haven't yet settled how `alr` extensions could be packaged and named.

IOW, i'd prefer if this were named `appimage_creator` or basically anything not starting with `alr_`. (The repo name can stay the same).

mgrojo:
> I have a concern here which is that the name may imply that this comes from the alire-project itself. Also we haven't yet settled how `alr` extensions could be packaged and named.

After our conversations on https://github.com/alire-project/alire/issues/1352 and my later analysis of what my tool should do, I no longer consider it an extension of `alr`. It's just a tool that reads your `alire.toml`, calls `alr` as a user could manually do, check its exit code, and at the end, gets a result unrelated to Alire. On that regard, it only automates steps using interfaces already provided to the user and which I think will be kept stable for the sake of users, so there's no real need for considering it an extension of the `alr` tool or Alire in general.

> 
> IOW, i'd prefer if this were named `appimage_creator` or basically anything not starting with `alr_`. (The repo name can stay the same).

Are we talking here only about the crate name, or also the project and executable name? In any case, `appimage_creator`, for me, is too unspecific. I'd like the name to convey what the tool is for. What do you think about calling it `alr2appimage`  or `alire2appimage`? 

Imagine also that I name it `appimage_creator` and the AppImage authors tell me that the name implies that it is an official tool of the AppImage project, and I should rename it. :wink:

I don't mind writing a disclaimer in the crate long description and in the repository readme to clarify the situation. Something like this:

> [!IMPORTANT]  
> This tool is not part of, is endorsed by or supported by the Alire nor the AppImage projects, in any way. It is a third-party project intended to help users of both projects.

mosteo:
> Are we talking here only about the crate name, or also the project and executable name? 

I'm only concerned in the context of the project/community index, so more or less what a user would find if doing `alr search alr` or `alr search alire`. So just the crate name.

>In any case, `appimage_creator`, for me, is too unspecific. I'd like the name to convey what the tool is for. What do you think about calling it `alr2appimage` or `alire2appimage`?

I think those are OK. Until (if ever) we have some sort of namespaces, the `alr_` is what kind of acts as one.

> Imagine also that I name it `appimage_creator` and the AppImage authors tell me that the name implies that it is an official tool of the AppImage project, and I should rename it. 😉

Yeah, that was just a low-effort idea :sweat_smile: 

> I don't mind writing a disclaimer in the crate long description and in the repository readme to clarify the situation.

It can't hurt, although if from context the situation is clear, I feel that's up to you.

Thanks for the understanding.

Fabien-Chouteau:
Looking at both git and cargo, it looks like neither of them has this kind of restriction. I understand your concerns @mosteo but I think we have to accept `alr.+` and `alire.+` crate names.

Now @mgrojo is you use that name, you have to know that the Alire project may introduce a `alr appimage` command in the future. I don't think it's likely, and we will probably put you in the loop if it happens, but that's a possibility.

mgrojo:
I'll rename it as `alr2appimage`.

PR #1012, opened 2024-03-14
Add gprbuild 24.0

reznikmm:
I've tested it with my `als-alire-index` on linux x86_86 and it seems it works.


Fabien-Chouteau:
I'd like to wait a little bit for the 2.0 dust to settle before we bring in new grpbuilds.

reznikmm:
I agree, no need to harry with this.

reznikmm:
In my opinion, we should rebuild gprbuild 24.0 release, because 24.0.1 uses `-static` option on linking stage for some reason. I don't know why, but if I rebuild it I don't see this behavior aby more :shrug:

PR #1013, opened 2024-03-14
Add gnat_native for MacOS X AArch64

reznikmm:


index/gn/gnat_native/gnat_native-13.2.1.toml
```diff
@@ -29,3 +29,8 @@ binary = true
 url = "https://github.com/alire-project/GNAT-FSF-builds/releases/download/gnat-13.2.0-1/gnat-x86_64-darwin-13.2.0-1.tar.gz"
 hashes = ["sha256:c21501ad511255d8e4a589f5e2e0aab883e9762617303010e0e82c4c6ec09275"]
 binary = true
+
+[origin."case(os)".macos."case(host-arch)".aarch64]
+url = "https://github.com/alire-project/GNAT-FSF-builds/releases/download/gnat-13.2.0-2/gnat-aarch64-darwin-13.2.0-2.tar.gz"
```
Fabien-Chouteau: This is release `13.2.2` not `13.2.1`, so we would need another manifest just for this one.

reznikmm: indeed! :100:

mosteo:
The failure in centos is expected until I fix the conundrum with the old GLIB version, so merging.

mosteo:
I mean, it can be merged, but better if @Fabien-Chouteau does it as he's in the loop in regard to this.

index/gn/gnat_native/gnat_native-13.2.1.toml
simonjwright: > This is release `13.2.2` not `13.2.1`, so we would need another manifest just for this one.

I’m not clear what makes it 13.2.2? Is it that without the release change `alr` won’t realise that there’s been an update?

I do realise that the GCC source for the aarch64 version isn’t _quite_ the same as the FSF source!

But the other compilers would remain the same.

Fabien-Chouteau: We use the patch version to differentiate different builds from the GNAT-FSF-builds repo.
This compiler was made with slightly different scripts.

Fabien-Chouteau (approved):

Fabien-Chouteau:
Ok for me @mosteo

mosteo (approved):

PR #1014, opened 2024-03-14
alr2appimage 0.9.2

mgrojo:
Created via `alr publish` with `alr 2.0`


mosteo (approved):

mosteo:
The trouble with old GLIB versions is going to be a bit involved, so here's not the place to fix it. Otherwise, this looks good, so merging.

PR #1015, opened 2024-03-19
WikiBook Ada Programming module wikibook_hello_world_1 release 1.0.0

krischik:
Updating the sample code for the WikiBook Ada Programming to Alire. I hope that will make it easier for anyone who reads the book.


mgrojo:
Why don't you include all the examples of the wikibook?

krischik:
> Why don't you include all the examples of the wikibook?

Because that is a lot of work which will take a few weeks or month. And one usually starts with a prove of concept. But if you have lot's of spare time I'll add you to the project ;-)

But more  importantly what is wrong with the build?

mgrojo:
> > Why don't you include all the examples of the wikibook?
> 
> Because that is a lot of work which will take a few weeks or month. And one usually starts with a prove of concept. But if you have lot's of spare time I'll add you to the project ;-)

OK, understood. I had curiosity about what you had in mind. Unfortunately, I'm spending more time in open source than what I should.

> 
> But more importantly what is wrong with the build?

It might be that the location is not correct. `index/ad/wikibook_hello_world_1` should be `index/wi/wikibook_hello_world_1`.

mgrojo:
I'm still wondering, if this is a placeholder, shouldn't it be wikibook_examples? Or do you plan to include a crate for each example? I think that would be overkill.

krischik:
> I'm still wondering, if this is a placeholder, shouldn't it be wikibook_examples? Or do you plan to include a crate for each example? I think that would be overkill.

Yes that is was plan. It's my understanding that crates are ment to be small and self contained. And putting all samples in one crate would be rather convoluted. Currently there are ≈20 gpr project file. So even back in the days I didn't put everything in one project. 

But since the pull request is not accepted yet there is still time to discuss this. I can always abandon the pull request and make a new one.

I see three options:

1. One crate per example
2. One crate to rule the all
3. One crate for each of the current ≈20 projects.

mgrojo:
Given the reduced size of example programs and for user convenience, I'd go for a single crate for downloading all the examples, that is, a single crate is included in the community index. As far as I understand, that doesn't mean that you cannot include other subcrates in the repository, one per project, if you like Alire to be used for building them, which also makes sense.

The ada_spark_workflow proposes a similar approach for the `tests/` directory, but I guess nothing impedes extending this to several subcrates, one for each example program, specially if they have different dependencies, although most will be vanilla Ada, I guess. See https://github.com/alire-project/ada_spark_workflow?tab=readme-ov-file#unit-testing

mosteo:
I'd go with @mgrojo's suggestion and have a single crate, with separate nested crates for each example. Note that such nested crates will be reported when getting or withing your top-level crate, e.g.:
```
$ alr get adacl
ⓘ Deploying adacl=5.15.1...                                              
adacl=5.15.1 successfully retrieved.
ⓘ Found 1 nested crate in /tmp/a/adacl_5.15.1_e7c1515b:
   test/adacl_test=5.15.1: Ada Class Library - AUnit test
```

krischik:
> for user convenience

Of course I always hope that more Ada developer join and supply sample code.

> I'd go with @mgrojo's suggestion and have a single crate, with separate nested crates for each example.

Interesting. Sound good. Three question:

1. How deep can you nest? Some of the samples could benefit from unit tests so I might need thee level `wikibook/sample/test` crates. I don't yet know but I don't want to close the option.
1. Will the GitHub test framework see and build them? I know that all crates which match “test*” will be build and run.
1. Under this setup the top level crate will be empty. Is that at all possible?

Of course I could start all the crates with “test” then they will all be build and run by the continues integration framework.

mosteo:
> 1. How deep can you nest? Some of the samples could benefit from unit tests so I might need thee level `wikibook/sample/test` crates. I don't yet know but I don't want to close the option.

There's no limit. But also there's no way to exclude a crate from being found.

> 2. Will the GitHub test framework see and build them? I know that all crates which match “test*” will be build and run.

No, but see below.

> 3. Under this setup the top level crate will be empty. Is that at all possible?

Not really; you'd need some root crate that might do nothing but refer people to the actual examples.

> Of course I could start all the crates with “test” then they will all be build and run by the continues integration framework.

That was a hack that might be removed at any time because there was an issue with `alr test`, that is fixed in alr 2.0. When `alr test` is run, the default test is to build the crate, but you can define any number of test actions (look for actions in the [spec](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md)); for example you could have one for each example that `alr build`s it, and one action parameter is in which folder the action must run.

krischik:
Ok try this and see how it works.

krischik:
```toml
executables                 = ["hello_world_1","hello_world_2","hello_world_3","hello_world_4"]
```

that doesn't actually work. Only the first executable is created.

mosteo:
> that doesn't actually work. Only the first executable is created.

What are you trying to do here? That field only informs Alire of which executables should exist after a successful build.

PR #1016, opened 2024-03-19
Add Cheddar 3.3

reznikmm:


index/oc/ocarina_lib/ocarina_lib-1.1.0-20070603.toml
```diff
@@ -0,0 +1,37 @@
+name = "ocarina_lib"
+description = "Ocarina is a stand-alone AADL model processor, written in Ada."
+version = "1.1.0-20070603"
```
mosteo, line 3: Is this really intended to be a pre-release?

reznikmm: I found `ocarina-1.1w-src-20070603.tar.gz` archive. They use it to build Cheddar 3.3. It seems the archive is just some snapshot of 1.1 work-in-progress version. That's why I named it `1.1.0-20070603`.

What is your advice?

mosteo: From what you say, I think that's the right way to name it, thanks for the details.

mosteo (approved):

PR #1019, opened 2024-03-22
wikibook 1.0.0

krischik:
Created via `alr publish` with `alr 2.0`


krischik:
ubuntu-lts is a little bit outdated:

```
```

mosteo:
> ubuntu-lts is a little bit outdated:

In that case you need a minimum version dependency on `gnat`.

krischik:
> In that case you need a minimum version dependency on `gnat`.

Since we have sub-crates that means a new tar file :-( .

krischik:
@mosteo done.

mosteo (approved):

mosteo:
Merged, thanks.

PR #1023, opened 2024-04-05
wolfss_ada 5.7.0-dev

Irvise:

### DIFF
--- a/index/wo/wolfss_ada/wolfss_ada-5.7.0-dev.toml
+++ b/index/wo/wolfss_ada/wolfss_ada-5.7.0-dev.toml
@@ -1,0 +1,21 @@
+name = "wolfss_ada"
+description = "WolfSSL encryption library and its Ada bindings"
+version = "5.7.0-dev"
+
+authors = ["Fernando Oleo Blanco"]
+maintainers = ["Fernando Oleo Blanco <irvise@irvise.xyz>"]
+maintainers-logins = ["Irvise"]
+licenses = "GPL-2.0-only"
+website = "https://www.wolfssl.com/"
+project-files = ["wolfssl.gpr"]
+tags = ["ssl", "tls", "embedded", "spark"]
+
+#[origin]
+#url="https://github.com/wolfSSL/wolfssl/releases/download/v5.6.6-stable/wolfssl-5.6.6.zip"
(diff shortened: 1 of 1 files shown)


Irvise:
Please, do not merge this PR! This is just a test to see if what I am doing is actually working!

Also, if you have any feedback, that would be appreciated. Did I do all right? Does the TOML look alright?

Also, it seems that the new "Fine-grained personal access tokens Beta" tokens do not work with Alire even when all the rights are granted. I had to use the classic one for it to work.

Cheers,
Fer

mgrojo:
Is it intentional the removal of the second "l" in the crate name, compared to WolfSSL?

Irvise:
> Is it intentional the removal of the second "l" in the crate name, compared to WolfSSL?

Oups! No, it was not intentional. Thank you for finding the issue!

index/wo/wolfss_ada/wolfss_ada-5.7.0-dev.toml
```diff
@@ -0,0 +1,21 @@
+name = "wolfss_ada"
+description = "WolfSSL encryption library and its Ada bindings"
+version = "5.7.0-dev"
```
mosteo, line 3: I know this is a test, but just in case, the `-dev` should not be use for a submitted release unless there are very good reasons that I can't come up with an example right now.

mosteo:
The manifest looks good to me. The problem on Windows seems an easy fix (some bad value for an external variable is used).

Irvise:
> The manifest looks good to me. The problem on Windows seems an easy fix (some bad value for an external variable is used).

Yes, the GPRBuild file checks for the "Windows" string variable for the OS. The runner says it is "Windows NT", which GPRBuild cannot match and fails. I will see if I can get it fixed. Thanks

mgrojo:
@Irvise are you still working on this? I'd like to see WolfSSL in the Alire index, and if you are no longer on it, I'd like to continue the work.

Irvise:
Hola Manuel,

I am still on this, I have not forgotten about it... I want to close WolfSSL's PR regarding Alire first, then maybe I will push this to Alire's index. The crate/project configuration needs to make a couple of other things work out-of-the-box, which they still don't, so that may take a bit longer...

mgrojo:
Hola Fernando,

Perfect. I just wanted to confirm that you still had the intention to get that done. Thanks and have a happy summer.

PR #1028, opened 2024-04-13
tiny_text 1.4.0

JeremyGrosser:


JeremyGrosser:
@Fabien-Chouteau I added a test that depends on qoi-spark, which is breaking the build

JeremyGrosser:
On second thought, I'm not quite ready for a stable release here. Will resubmit later.

Fabien-Chouteau:
> @Fabien-Chouteau I added a test that depends on qoi-spark, which is breaking the build

Here?:
```
qoi.ads:39:06: "Relaxed_Initialization" is not a valid aspect identifier
```

Looks like I have to set a `gnat` dependency constraint on QOI.

PR #1030, opened 2024-04-15
Update gwindows-1.4.2.toml

zertovitch:
Now that the GNAT setup is working, I am trying a version of the .toml two steps earlier in the previous PR.
- the gnatcom path is restored (not sure it is of any use, but it was put there by the first author of the .toml)
- Zip ball is used instead of Git (just for feeding the download statistics!)


Fabien-Chouteau:
Hi @zertovitc, we are not supposed to change the origin of an already published crate (https://alire.ada.dev/docs/#release-immutability).

zertovitch:
Ok, thanks for the explanation!

PR #1034, opened 2024-04-16
hirtos_separation_kernel 2.0.0

jgrivera67:

Updated HiRTOS separation kernel crate to version 2.0.0:
- Bug fixes
- Code refactoring to make future ports easier

### DIFF
--- a/index/hi/hirtos_separation_kernel/hirtos_separation_kernel-2.0.0.toml
+++ b/index/hi/hirtos_separation_kernel/hirtos_separation_kernel-2.0.0.toml
@@ -28,7 +28,7 @@ directory = "../sample_apps/hello_partitions"
 
 
 [origin]
-commit = "b091813c4123af30d6601ed70aea2b0dbbae55e5"
+commit = "ebcd8cb20115356aa0d99c97124e47a175e56e8e"
 subdir = "./hirtos_separation_kernel/"
 url = "git+https://github.com/jgrivera67/HiRTOS.git"
 
--- a/index/hi/hirtos_separation_kernel/hirtos_separation_kernel-2.0.0.toml
+++ b/index/hi/hirtos_separation_kernel/hirtos_separation_kernel-2.0.0.toml
@@ -28,7 +28,7 @@ directory = "../sample_apps/hello_partitions"
 
 
 [origin]
-commit = "0c01280da07399a8a7fa8cb7980c3b7c3713801d"
+commit = "b091813c4123af30d6601ed70aea2b0dbbae55e5"
 subdir = "./hirtos_separation_kernel/"
 url = "git+https://github.com/jgrivera67/HiRTOS.git"
 
--- a/index/hi/hirtos_separation_kernel/hirtos_separation_kernel-2.0.0.toml
+++ b/index/hi/hirtos_separation_kernel/hirtos_separation_kernel-2.0.0.toml
@@ -1,6 +1,11 @@
 #
-# NOTE: Before building for the first time with `alr build`, select the
-# corresponding cross compiler toolchain by running `alr toolchain --select`
+# Copyright (c) 2022-2024, German Rivera
+#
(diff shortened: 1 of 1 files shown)


index/hi/hirtos_separation_kernel/hirtos_separation_kernel-2.0.0.toml
```diff
@@ -0,0 +1,33 @@
...
+maintainers-logins = ["jgrivera67"]
+
+[configuration.variables]
+Separation_Kernel_Debug_Tracing_On = {type = "Boolean", default = false}
+Platform = {type = "Enum", values = ["arm_fvp"], default = "arm_fvp"}
+
+[[depends-on]]
+gnat_arm_elf = "^13.2.1"
+#gnat_riscv64_elf = "^13.2.1"
+gnatprove = "^13.1.1"
```
mosteo: Hi @jgrivera67 , dependencies on `gnatprove` should be in a different nested crate that does the proving. Otherwise clients bear the burden of downloading gnatprove simply to use your library.

Nested crates are described here: https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing

jgrivera67: Changed as suggested

jgrivera67: Hi @Fabien-Chouteau and @mosteo. Can you guys please re-review this PR?

jgrivera67: Hi @mosteo, I don't understand why `alr test` is failing in the CI environment, since it works fine when I run it locally. I'm getting this error in the pre-build step:
```
ERROR: Mismatched crates for pin linking to /home/runner/work/alire-index/alire-index/HiRTOS_ebcd8cb2: expected hirtos but found hirtos_separation_kernel
```
Can you please help me undserstand if I'm doing something wrong here

mosteo: ~I'm looking into this, but at first I'd say this is an spurious error, and the real problem is in [this line](https://github.com/alire-project/alire-index/actions/runs/9175873865/job/25229833553?pr=1034#step:7:494).~

 Not right, I'm now on the good track.

mosteo: Unfortunately this is caused by a [bug](https://github.com/alire-project/alire/pull/1684) in `alr 2.0.1`. The cause is rather obscure but ultimately related to this being a nested crate in a monorepo, with a different crate at the root of the repo. 

Options:
- Wait for the bug to be fixed and `2.0.2` released (may take some time)
- Move the crate at the root (`hirtos`) to its own subdir
- Merge as-is. The bug will affect only people using `alr get` on a nested crate in the repo (e.g. `alr get hirtos_separation_kernel`), instead of e.g. `alr get hirtos`. The release is otherwise usable in normal circumstances (as a dependency), so that's why I don't see much issue with merging.

Your choice, I'm fine with either of those.

There's a separate issue: your `sample_apps/hello_partitions/build_partitions.sh` is not checking for errors, so that's why we are seeing mixed errors about pins but the build progresses until it fails. You can add `set -o errexit` to that script so it stops as soon as a failure is detected.

jgrivera67: Hi Alejandro, Thanks for investigating this. Can you please merge it as-is, so that both the `hirtos` and the `hirtos_spearation_kernel` crates published on alire are at the same release tag?
I'll fix the issue you mentioned for `sample_apps/hello_partitions/build_partitions.sh` in the next release.

mosteo (approved):

PR #1037, opened 2024-04-19
libsimpleio release 2.22223.1

pmunts:
Rebuilt for alr 2.0.


pmunts:
Forgot the postfetch script.  Will resubmit.

PR #1039, opened 2024-04-20
remoteio release 2.22297.1

pmunts:
Reworked for alr 2.0


pmunts:
Need to rework for macos, too.  Will resubmit later.

PR #1050, opened 2024-05-02
Allow zeromq on Mac OS X and Arch Linux

reznikmm:


reznikmm:
Currently `macos-latest` is `macos-14` which is ARM64/M1, so alr x86_64 can't run on it.

Fabien-Chouteau (approved):

PR #1060, opened 2024-05-10
optional 0.2.0

mosteo:

### DIFF
--- a/.github/workflows/authorship.yml
+++ b/.github/workflows/authorship.yml
@@ -27,11 +27,8 @@ jobs:
         ref: ${{github.base_ref}}
         path: ${{env.CHECKS_REPO}}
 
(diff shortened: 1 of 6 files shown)


mosteo:
Hi @simonjwright , I tried downgrading the macos images for the time being, but I'm seeing linking errors still in some cases. Any ideas?

simonjwright:
You need to go back to macos 12; the linking errors in x86_64 are caused by Xcode 15.0, which macos-13 and -14 are both using. macos-12 uses Xcode 14 (.2, I think).

Both the macOS linking errors are caused while trying to link the development alr. Why does the MacPorts run try to do this?

Fabien-Chouteau (approved):

PR #1062, opened 2024-05-17
emdee 0.2.0

SMerrony:


mosteo:
There is some missing Gtk dependency in Arch and the compiler in Ubuntu LTS is too old, so a dependency on the minimum GNAT is needed.

mosteo:
The remaining errors don't seem to be the fault of the submitted crate. I'm on it.

SMerrony:
Closing due to bugfix release - will create new pull.

PR #1068, opened 2024-05-21
dashera 0.15.1

SMerrony:
Update to latest release.

### DIFF
--- a/index/da/dashera/dashera-0.15.1.toml
+++ b/index/da/dashera/dashera-0.15.1.toml
@@ -1,0 +1,24 @@
+name = "dashera"
+description = "DASHER Terminal Emulator"
+version = "0.15.1"
+licenses = "MIT"
+website = "https://github.com/SMerrony/dashera"
+
+authors = ["Stephen Merrony"]
+maintainers = ["Stephen Merrony <merrony@gmail.com>"]
+maintainers-logins = ["SMerrony"]
+
+tags = ["gtk", "terminal", "dasher", "emulator",]
+
+executables = ["dashera"]
+
+[[depends-on]]
+gtkada = "^23.0.0"
+
+[build-switches]
+development.style_checks = ["-gnaty-m"]
+release.style_checks = "No"
+[origin]
+commit = "3d8309f0aae36124bef9ceee487213a0e5e8bccf"
+url = "git+https://github.com/SMerrony/dashera.git"
+


mosteo:
Similar errors to https://github.com/alire-project/alire-index/pull/1062 ?

SMerrony:
> Similar errors to #1062 ?

Indeed - but from my perspective, the problem is within the Gtkada crate, not mine.

mosteo:
> Indeed - but from my perspective, the problem is within the Gtkada crate, not mine.

That's possible, a dependency missing in gtkada itself?

simonjwright:
~~I can speak to the macports issue: in libgtk3-external, I should have used the gtk3-devel package, not plain gtk3. Will get right to it.~~

I wonder what other similar errors we have? I’m not a macports user.

Should we update pkg_config-external to support Centos? (the package would be pkgconfig)

simonjwright:
There's a tricky problem with `libgtk3` on macOS MacPorts.

The trouble is that, for [reasons](https://trac.macports.org/ticket/60511) whose force I don't see, the MacPorts maintainers insist that the default backend for gtk3 and glib2 should remain x11, whereas the native macOS backend is [quartz](https://en.wikipedia.org/wiki/Quartz_(graphics_layer)). Homebrew has gone over to quartz completely.

You would think we could change `libgtk2-external` to say `macports = ["gtk3", "+quartz"]` or something similar, but `alr` doesn't support that. And, in any case, MacPorts would happily install the `+x11` version of `glib`, regardless of the compatibility with `gtk3`.

There seems to be one way of countering this problem, and anyone who uses MacPorts and gtkada will have to go this way (if they haven't already): edit the file `/opt/local/etc/macports/variants.conf`, and uncomment the line 
```
#   -x11 +no_x11 +quartz
```

As to what we should do about the current crate CI situation, we could
* not test MacPorts, or
* make sure to accept crates which fail for this reason, or
* edit `variants.conf` as above.

mosteo:
> * edit `variants.conf` as above.

Thanks for the assessment, @simonjwright. If users would need to do this anyway, or ensure they use the quartz variant somehow, I think we can adjust our workflow to do the same.

mosteo:
@SMerrony, when I labeled this as `fix dependencies`, I don't mean you must take care of fixing whatever problem is happening upstream; rather that you mark unavailable any platform that is not going to work in normal circumstances, or help identify ambiguous situations that we can work out together.

For example, the Arch failure might be a missing C library in base Arch installations that `gtkada` should make explicit. This could be fixed with a external dependency here, or adding the dependency to our Arch image, depending on what's most likely to be a regular user situation.

mosteo:
> Should we update pkg_config-external to support Centos? (the package would be pkgconfig)

CentOS was missing libgtk3, which I just added, but even so its version is too old (3.22).

mosteo:
And Arch I'm not sure what's going on, because `/usr/include/glib-2.0/glib.h` is installed in our Docker image, which is the same location as in Ubuntu.

mosteo:
@SMerrony If it's OK with you, we can go ahead and merge, as the trouble with the remaining platforms is in gtkada.

SMerrony:
> @SMerrony If it's OK with you, we can go ahead and merge, as the trouble with the remaining platforms is in gtkada.

Fine by me.  Sorry for any misunderstandings.

mosteo (approved):

PR #1072, opened 2024-05-23
Set Quartz backend for MacPorts

mosteo:
As discussed in https://github.com/alire-project/alire-index/pull/1068#issuecomment-2124945403


mosteo:
Tested at https://github.com/mosteo/alire-index/pull/2

PR #1077, opened 2024-05-25
asfml 2.6.1

mgrojo:

### DIFF
--- a/index/as/asfml/asfml-2.6.1.toml
+++ b/index/as/asfml/asfml-2.6.1.toml
@@ -1,0 +1,39 @@
+name = "asfml"
+description = "Ada binding to SFML, the Simple and Fast Multimedia Library"
+version = "2.6.1"
+long-description = """
+[![ASFML logo](https://raw.githubusercontent.com/mgrojo/ASFML/master/images/ASFML_Logo.svg)](https://www.sfml-dev.org)
+![Ada (GNAT)](https://github.com/mgrojo/ASFML/workflows/Ada%20(GNAT)/badge.svg)
+[![Gitter chat](https://badges.gitter.im/gitterHQ/gitter.png)](https://gitter.im/ada-lang/Lobby)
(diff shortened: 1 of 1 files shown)


mgrojo:
It seems CSFML v2.6.1 is too new for the distributions to be updated. In a case like this, should we wait for inclusion in the Alire index? I think it makes sense the inclusion, since people can always provide a locally installed version of the external library, correct?

mosteo:
So there's not a single platform that has a recent enough csfml? How are you testing locally?

mgrojo:
Oh, I'm cloning SFML and CSFML, building them, and I've modified my project files adding `for Linker_Options use  ("-L../../SFML/SFML/lib", "-L../CSFML/lib",` so my compilation takes the latest libraries. Thus, I guess users won't get any benefit of having this in the index before the installation of the external crate can be done by Alire. The only advantage from having this already in the index might be that it will start working when the version is included in the distributions. Maybe it's going to be confusing for users, and it's better if it's only released in my repository until CSFML v2.6.1 is more widespread?

mosteo:
I see. Yes, for the time being I think it makes more sense to not include it yet in the public index if it requires manual intervention. 

Still, I'm rethinking how we should deal here with missing dependencies, rather than failing the workflow outright. Alas, there's no middle way in github actions for now.

mgrojo:
> Still, I'm rethinking how we should deal here with missing dependencies, rather than failing the workflow outright. Alas, there's no middle way in github actions for now.

Maybe [Action Summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) can help a bit, at least in conveying easily the different results of a failing check?

See example of the Link checker in the awesome-ada repository: https://github.com/ohenley/awesome-ada/actions/runs/8863535503

mosteo:
> Maybe [Action Summaries](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) can help a bit, at least in conveying easily the different results of a failing check?

Looks promising! Thanks for the pointer.

mosteo (approved):

PR #1082, opened 2024-05-29
gnoga metadata: removed conflicting project variable settings

mosteo:


mosteo:
Windows checks cannot succeed until simple_components is split into finer-grained crates with separate dependencies.

PR #1083, opened 2024-05-31
embedded_components 0.3.0

Fabien-Chouteau:


mosteo:
My guess is that the Ubuntu LTS compiler is too old, so a minimum version of GNAT should be added.

mosteo (approved):

PR #1085, opened 2024-06-01
Update libmariadb Debian/Ubuntu package name

stcarrez:
The Debian package was renamed into libmariadb-dev.

With the old name, alr fails to find a solution and looks forever.


mosteo:
I guess the packages attempted to be removed are the old versions so the failure is to be expected.

mosteo:
Rather than removing the old package name, both should be kept so people using the pre-renaming OS versions don't loss access to this library.

mosteo:
I've done the change

PR #1086, opened 2024-06-01
Update globe_3d-2023.11.12.toml metadata (tags)

zertovitch:
Metadata update only in .toml: added a few tags.


zertovitch (commented):
Metadata update only in .toml: added a few more tags.

zertovitch (commented):
Metadata update only in .toml: shortened tags (max len=15).

mosteo (approved):

PR #1092, opened 2024-06-17
SDL2 broken on macOS at link time

mosteo:


simonjwright:
As far as I know, the problem here is that the library (e.g. `sdl.ads`, but others too, no doubt) has Ada 2022 constructs, so that a client crate has to use `-gnat2022` or equivalent. Are there other library crates which impose this on users?

mosteo:
> Are there other library crates which impose this on users?

One recent example that I found is `vss`. And `uxstrings` I think too.

If this is all there is to it, we need to add a proper test action for such crates.

I wonder if the minimum Ada version should become part of the manifest.

PR #1093, opened 2024-06-18
getada 1.0.1 

AJ-Ianozi:
Is this recursive? Perhaps, but if you want to build `getada` without having to go to the repo, this is helpful.

### DIFF
--- a/index/ge/getada/getada-1.0.1.toml
+++ b/index/ge/getada/getada-1.0.1.toml
@@ -1,0 +1,31 @@
+name = "getada"
+description = "The unofficial mac and linux installer for Alire."
+version = "1.0.1"
+licenses = "GPL-3.0-only"
+
+website = "https://www.getada.dev/"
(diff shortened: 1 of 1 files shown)


AJ-Ianozi:
It looks like `(alr dev) / windows-latest` failed for an unrelated reason (an msys2 package was too slow to download?).  Anyway just to re-run that workflow?

It's worth noting that had it gotten to the point where it tried building the package, it would have failed gracefully anyway (getada isn't available on windows as per the `[available.'case(os)]'` section in its `alire.toml`)

If there's a way someone could run that windows-latest dev again, it may "just work"

mosteo (approved):

mosteo:
Yup, Windows issue was some internal networking. With the amount of checks we run, I'm used to seeing some errors in the backend. E.g., macOS images are particularly prone to disk errors for some reason (or they detect them better).

Anyway, all it's OK now.

PR #1108, opened 2024-07-02
Adabots 1.9.0

TamaMcGlinn:


Fabien-Chouteau:
It looks like json_ada doesn't compile with GNAT FSF 14:
```
json-streams.ads:32:29: error: function cannot have parameter of mode "out" or "in out" in SPARK [E0015]
```

Fabien-Chouteau:
Hi @onox, GNAT FSF 14 is now checking more SPARK restrictions. Like this one:
```
json-streams.ads:32:29: error: function cannot have parameter of mode "out" or "in out" in SPARK [E0015]
```

We will have to specify a `gnat` dependency on the `json_ada` releases.

onox:
> We will have to specify a `gnat` dependency on the `json_ada` releases.

Thanks for mentioning this. Tracked in https://github.com/onox/json-ada/issues/12. I'll try to fix it this weekend to unblock this PR ASAP.

onox:
I have openend #1164. json v6.0.0 should compile with GNAT < 14.

mosteo:
```
json-streams.ads:32:29: error: function cannot have parameter of mode "out" or "in out" in SPARK [E0015]
json-streams.ads:32:29: error: launch "gnatprove --explain=E0015" for more information
```

TamaMcGlinn:
Ah, I think the issue may be that I was trying to release for alire 1.3.0 - but I am currently using 1.2.2 and actually, alire 1.3.0 isn't even in the [releases list](https://github.com/alire-project/alire/releases/). I don't understand why there is even a 1.3.0 branch at all.

mosteo:
> I don't understand why there is even a 1.3.0 branch at all.

Hi Tama. That's because the versioning of the index and of Alire aren't the same. The latest index format version is `1.3.0`, the latest Alire is `2.0.1`. Which index versions are understood by Alire is shown by `alr version`.

But I don't think that's related to the errors, which seem to come from more recent versions of the toolchain being more restrictive.

mosteo:
Hi @TamaMcGlinn, I've re-run the tests with updated Docker images and our latest alr release and there are no issues remaining, so I'm merging this PR. Thanks for the patience.

mosteo (approved):

PR #1110, opened 2024-07-04
PowerJoular 1.0.4

adelnoureddine:
Update Alire's index with the latest major version of PowerJoular.


adelnoureddine:
Here is an update. It compiles on Linux which is the only platform PowerJoular supports for now.

mosteo:
Replaced by #1123, as this PR's base branch is now obsolete (and it can't be changed without closing).

PR #1114, opened 2024-07-08
pragmarc 1.0.0

bracke:

### DIFF
--- a/index/pr/pragmarc/pragmarc-1.0.0.toml
+++ b/index/pr/pragmarc/pragmarc-1.0.0.toml
@@ -1,6 +1,6 @@
 name = "pragmarc"
 description = "The PragmAda Reusable Components"
-version = "1.0.0"
+version = "v20240323"
 
 authors = ["Jeff Carter"]
 maintainers = ["Bent Bracke <bent@bracke.dk>"]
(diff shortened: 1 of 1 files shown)


mosteo:
Hi @bracke , have you agreed with @jrcarter on the versioning numbers here? I think he's simply using the date for releases.

bracke:
No, we have not talked about versioning.
Would a date even be useable in Alire?

zertovitch:
It is possible to use a date. For instance here:
https://github.com/alire-project/alire-index/blob/stable-1.3.0/index/ma/mathpaqs/mathpaqs-20230121.0.0.toml

mosteo:
> No, we have not talked about versioning. Would a date even be useable in Alire?

Then, I'd very much prefer if the version matched, as @zertovitch exemplified, the tag from which this release is taken (assuming you're using one of @jrcarter's tagged releases).

bracke:
>Then, I'd very much prefer if the version matched, as @zertovitch exemplified,
> the tag from which this release is taken (assuming you're using one of @jrcarter's tagged releases).

I'm using the default branch, which seems to be the tag v20240323.
I have changed the version to v20240323.

index/pr/pragmarc/pragmarc-1.0.0.toml
```diff
@@ -0,0 +1,22 @@
+name = "pragmarc"
+description = "The PragmAda Reusable Components"
+version = "v20240323"
```
mosteo, line 3: The `v` should be removed, and the file renamed to match the version, to `pragmarc-20240323.0.0.toml`.

mosteo (changes requested):

PR #1119, opened 2024-07-12
Add gprbuild 24.0 with Mac OS M1

reznikmm:


mosteo (approved):

index/gp/gprbuild/gprbuild-24.0.1.toml
```diff
@@ -0,0 +1,33 @@
+name = "gprbuild"
+version = "24.0.1"
+description = "The GPRBuild Ada/multilanguage build tool"
+maintainers = ["chouteau@adacore.com"]
+maintainers-logins = ["Fabien-Chouteau"]
```
mosteo, line 5: @reznikmm @Fabien-Chouteau Should @reznikmm be added here?

reznikmm: I have no objection to append me to maintainers list. But I don't think I can make big contribution here.

reznikmm:
I've tested this gprbuild 24.0 with my ada_language_server GitHub CI. It's able to build shared and static libraries, executables for Linux x86_64, MacOS X x86_64/AArch64. I would like to use it from the community index.

mosteo:
@Fabien-Chouteau is it OK to add this build of `gprbuild` to our repertoire?

Fabien-Chouteau:
> @Fabien-Chouteau is it OK to add this build of `gprbuild` to our repertoire?

If Max tested it then yes.

mosteo:
Hi @reznikmm, this release introduced the regression described in https://github.com/alire-project/GNAT-FSF-builds/issues/71#issuecomment-2260932828

I missed that this PR was adding all the builds for v24, at some point I started thinking it only added the ARM64 build. So this will affect people generally.

We have a rule against pulling releases from the index, but in this case I'm not so sure... Unless a fix can be released quickly, I'd rather pull this down for the time being.

Fabien-Chouteau:
We should remove it, unfortunately, as we won't be able to fix it quickly. What will be the consequences for users if they selected this a default GPRbuild?

mosteo:
It seems to fail at linking time when using dynamic libraries, but only for executables. 

By default, with recent crates or libraries that do not bother to modify defaults in the gpr file, it seems to work normally.

Also, when building a library it also succeeds, even if the library depends on other ones. It's only when linking a final executable that it fails (in my limited testing).

reznikmm:
Can we left it for macos x arm64?

Чт, 1 серп. 2024, 12:00 користувач Alejandro R Mosteo <
***@***.***> пише:

PR #1145, opened 2024-07-21
image_random 20200720.0.0

bracke:

### DIFF
--- a/index/im/image_random/image_random-20200720.0.0.toml
+++ b/index/im/image_random/image_random-20200720.0.0.toml
@@ -1,0 +1,35 @@
+name = "image_random"
+description = "True random numbers from a digital camera "
+version = "20200720"
+
+authors = ["Jeff Carter"]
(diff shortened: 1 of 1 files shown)


mosteo (commented):
It's now a bit late since there are a bunch of releases already using this versioning, but I've realized that it would be more future-proof to use the pre-1 versions for these dates. The issue is that Jeff won't use (for now) semantic versioning, but if he wanted to start at some point, he would have to use inordinately large major numbers.

That is, it would be better to use `0.20200720.0` (as there is no expectancy of compatibility between `0.x` versions) and so, if Jeff adopted semantic versioning at some point, he could start at `1.0`.

Alas, I didn't think of this sooner.

bracke:
Are You saying I should resubmit it with the revised version number?

mosteo:
> Are You saying I should resubmit it with the revised version number?

Not really. Just to have in mind for future opportunities.

mosteo (approved):

PR #1150, opened 2024-07-21
qplt 20230720.0.0

bracke:

### DIFF
--- a/index/qp/qplt/qplt-20230720.0.0.toml
+++ b/index/qp/qplt/qplt-20230720.0.0.toml
@@ -55,6 +55,9 @@ pragmarc = "^20240323.0.0"
 [[depends-on]]
 ada_gui = "^20240224.0.0"
 
+[[depends-on]]
+ssl = "*"
+
 [origin]
 commit = "2f7f77c911a868ea77d215a81d4d23175f531af8"
 url = "git+https://github.com/bracke/Qplt.git"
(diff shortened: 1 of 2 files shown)


bracke:
It seems to complain about ssl. But only on some platforms?

mosteo:
> It seems to complain about ssl. But only on some platforms?

In those not failing, something is requesting the linking to those things in some undeclared way. The proper thing for the Alire index, though, is that the first dependency requiring `ssl` declares explicitly the dependency on the external library. I'm not sure which one is though, I see `libtls` in the index, but I also see `aws` depending on `openssl`.

bracke:
The project only depends on Ada_Gui and PragmARC.
AdaGui depends on SimpleComponents but also GNAT.Sockets (referenced & used in the code).

Could GNAT.Sockets be the culprit?

mosteo:
> The project only depends on Ada_Gui and PragmARC. AdaGui depends on SimpleComponents but also GNAT.Sockets (referenced & used in the code).
> 
> Could GNAT.Sockets be the culprit?

From that list, the likely culprit looks to me to be `simple_components`. But it's strange, since it depends on `libgnutls`, which should be pulling in any needed system dependency.

Let me do a test, I will push to your branch.

mosteo (approved):

mosteo:
Indeed adding a dependency on `ssl` fixes it.

PR #1158, opened 2024-07-21
controlled_io 20240402.0.0

bracke:


index/co/controlled_io/controlled_io-20240402.0.0.toml
```diff
@@ -0,0 +1,34 @@
+name = "controlled_io"
+description = "Files that are open while they exist and closed when they don't "
+version = "20240402"
```
mosteo, line 3: If there are no source code changes w.r.t. the `20240401` version, a simple patch version upgrade to `20240401.0.1` would suffice.

PR #1159, opened 2024-07-21
ada_gui 20240225.0.0

bracke:


index/ad/ada_gui/ada_gui-20240225.0.0.toml
```diff
@@ -0,0 +1,59 @@
+name = "ada_gui"
+description = "An Ada-oriented GUI"
+version = "20240225"
```
mosteo, line 3: Likewise, is there a real change wrt the version one day earlier?

PR #1164, opened 2024-07-28
Add json 6.0.0 and update json 5.0.x

onox:


Fabien-Chouteau:
I see compile failures because of a warning:
```
json-streams.ads:17:17: warning: unit "Ada.Finalization" is not referenced [-gnatwu]
```

In general you don't want to treat warnings as errors for release builds.

mosteo (approved):

PR #1170, opened 2024-08-02
lispcli 0.1.0

BrentSeidel:

### DIFF
--- a/index/li/lispcli/lispcli-0.1.0.toml
+++ b/index/li/lispcli/lispcli-0.1.0.toml
@@ -11,7 +11,7 @@ tags = ["lisp"]
 
 executables = ["lispcli"]
 
-[available."case(os)"]
+[available."case(distribution)"]
 "suse" = false
 "..." = true
 
--- a/index/li/lispcli/lispcli-0.1.0.toml
+++ b/index/li/lispcli/lispcli-0.1.0.toml
@@ -18,7 +18,7 @@ bbs = "~0.1.0"
 bbs_lisp = "~0.1.0"
 
 [origin]
-commit = "650fe13c1bad00c8242f54ad2b0916ed55505857"
+commit = "640af271f4d8b053d6d95952804fdd0045771e62"
 subdir = "./cli/"
 url = "git+https://github.com/BrentSeidel/Ada-Lisp.git"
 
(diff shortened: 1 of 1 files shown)


BrentSeidel:
This is meant to be a sub-crate of bbs_lisp.  Not sure that I did it correctly.

BrentSeidel:
I see why the checks failed (the executable produced was "lisp" rather than "lispcli") and have pushed an update to my repository.  How do I update the pull request (something like "alr publish update" would be nice)?

mosteo:
> How do I update the pull request (something like "alr publish update" would be nice)?

You clone your fork of the index, check out the branch for this PR, update the manifest for your crate there to fix any necessary metadata, commit and push. Basically the same procedure as to fix any regular PR.

You could also cancel this PR and open a new one, but that's overkill if you're used to updating PRs as I described.

Feel free to open an issue to request the `alr publish --update` feature, I agree that it could come in handy.

index/li/lispcli/lispcli-0.1.0.toml
```diff
@@ -0,0 +1,28 @@
...
+maintainers-logins = ["BrentSeidel"]
+licenses = "GPL-3.0-or-later"
+website = "https://github.com/BrentSeidel/Ada-Lisp"
+tags = ["lisp"]
+
+executables = ["lispcli"]
+
+[available."case(distribution)"]
+"suse" = false
+"..." = true
```
mosteo, line 16: Out of curiosity, what's the trouble with SUSE?

mosteo (approved):

index/li/lispcli/lispcli-0.1.0.toml
BrentSeidel: The build on Open SuSe failed with a circular dependency in bbs_lisp, which is odd because bbs_lisp built just fine by itself.  Not just if it's something to do specifically with SuSe itself or the version of gnat on it or just what.  I don't have a SuSe system myself to do debugging so I just made it not available.  If anyone has ideas or suggestions, I would welcome them.

mosteo: I suspected as much. It's a small difference, but for future releases it may be better to specify a minimal gnat version rather than disabling the platform (that might eventually update their gnat).

PR #1176, opened 2024-08-07
simcpucli 0.1.0

BrentSeidel:

### DIFF
--- a/index/si/simcpucli/simcpucli-0.1.0.toml
+++ b/index/si/simcpucli/simcpucli-0.1.0.toml
@@ -29,3 +29,8 @@ bbs_lisp = "~0.1.0"
 
 [[depends-on]]
 bbs_simcpu = "~0.1.0"
+
+[origin]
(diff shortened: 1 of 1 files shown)


BrentSeidel:
Do you know what gnat version openSuSe uses so that I can request a later version?  Otherwise I can just omit it like I did with my lispcli crate.

BrentSeidel:
Did something change with the MacOS build environment?  It looks like it is failing before it actually tries to fetch and build my crate.  I also develop on MacOS and it builds fine there.

mosteo:
> Do you know what gnat version openSuSe uses so that I can request a later version? Otherwise I can just omit it like I did with my lispcli crate.

Open SUSE is on GNAT 7.5

BrentSeidel:
With the gnat > 7.5 dependency, OpenSuSe tries to install gnat 14.1.3, but the installation fails.  Should I disable this for SuSe or just let it go?

mosteo:
This is for us to fix in the opensuse support. No need to change anything.

mosteo (approved):

PR #1177, opened 2024-08-07
loadcpm 0.1.0

BrentSeidel:


BrentSeidel:
Did something change with the MacOS build environment?  It looks like it is failing before it actually tries to fetch and build my crate.  I also develop on MacOS and it builds fine there.

mosteo:
Yes, I'm seeing the same failure in that check for all PRs. Since it only affects the dev version we'll see about that separately.

mosteo (approved):

PR #1182, opened 2024-08-10
Stable 1.2.1 Adare-Dilithium5 and Adare-HQC256 pull request

danieagle:
Adare-Dilithium5 and Adare-HQC256 pull request.

GrateFull,
Dani.


mosteo:
Hi @danieagle, two things here:

- The PR should be against base branch `stable-1.3.0`.
- Since these self-describe as bindings, I presume they need some extra library that should be declared as a dependency? Or is it included in sources and built too?

danieagle:
> Hi @danieagle, two things here:
> 
>     * The PR should be against base branch `stable-1.3.0`.

Ok! I Will do It.

> 
>     * Since these self-describe as bindings, I presume they need some extra library that should be declared as a dependency? Or is it included in sources and built too?

Hi!  @mosteo

These packages includes all files needed to build.

Very Very Very Thanks!

GrateFull,
Dani.

PR #1184, opened 2024-08-10
Create bar_codes-3.0.0.toml

zertovitch:
New version with Data Matrix 2D bar codes.


zertovitch (commented):
Shortened description.

zertovitch:
The failed build on opensuse-leap is about a pre-Ada-2012 GNAT:
```
bar_codes_media.adb:155:07: prefix of "Image" attribute must be a type
bar_codes_media.adb:156:07: prefix of "Image" attribute must be a type
```

mosteo:
Yes, opensuse ships with a rather old GNAT. You need to add a dependency on gnat>=10 or so.

mosteo (approved):

PR #1192, opened 2024-08-12
bbs_lisp 0.1.1

BrentSeidel:


BrentSeidel:
This is primarily a metadata and documentation update.

mosteo (approved):
This PR changes the origin so it requires a different version, but know that if you only want to update the metadata in the index for minor things like tags and descriptions, you could do so without a new release.

PR #1208, opened 2024-08-23
Add enet_stm32-1.0.0.toml, ENet driver for STM32

reznikmm:


reznikmm:
This is a generic library that depends on cortex_m. I can't define a value for `cortex_m.core` because only user of the library will know it. I don't know how to fix CI.

If you agree to merge this PR with broken CI, please, let me know and I'll ask Stefane to accept my changes and correct `origin` and `commit` in this PR after that.

mosteo:
The advice for such crates is to have a nested testing crate that uses some meaningful values, and add the test action in the main crate. See e.g. the `hirtos 2.0` manifest.

reznikmm:
HiRTOS doesn't help me. I've opened an issue with a reproducer. 

https://github.com/alire-project/alire/issues/1742

mosteo:
Thanks for the reproducer. The original problem is now fixed in the checks, but unfortunately an unreleased fix in `alr` itself is still needed for an unrelated issue that arises with `alr get` + pins + monorepos. That's why the few tests using the master branch do pass. 

I'm going to merge for now as-is given those successes, and `2.0.2` should be arriving soon.

mosteo (dismissed):

mosteo:
Is this OK to merge, @reznikmm ? Since it's still marked as draft.

reznikmm:
Thank you @mosteo for quick response and fixes!

Please don't merge this, before I update `origin.url` link to upstream repository (`stcarrez/ada-enet`).

reznikmm:
I've updated commit and origin.url, but perhaps we should wait for `2.0.2`. I'm not sure, that I did everything right. I've tried to run `alr get --build enet_stm32` and it fails (both `2.0.1` and nightly version).

reznikmm:
I've updated commit and `origin.url`. I've tested it with `alr init --bin test; <edit gpr&toml>; alr with enet_stm32; alr build` and it works. (But `alr get -build enet_stm32` doesn't work, BTW).

mosteo:
> But `alr get --build enet_stm32` doesn't work

I expect this to fail due to incomplete configuration, yes. If the error is different then we should look at it.

reznikmm:
yes, indeed.

index/en/enet_stm32/enet_stm32-1.0.0.toml
```diff
@@ -0,0 +1,32 @@
...
+generate_c = false
+
+[configuration.variables]
+TX_Ring_Size  = {type = "Integer", first = 0, last = 1024, default = 8}
+RX_Ring_Size  = {type = "Integer", first = 0, last = 1024, default = 8}
+Extra_Buffers = {type = "Integer", first = 0, last = 1024, default = 8}
+
+[[actions]]
+type = "test"
+command = ["alr", "-C", "../../demos/ping_text_io", "build"]
```
mosteo, line 22: No need to edit, but so you know: there's a `directory` action field you can use instead of `-C`.

mosteo (approved):

PR #1209, opened 2024-08-25
yass 3.1.0

AJ-Ianozi:


AJ-Ianozi:
The failing fedora crate should work after https://github.com/alire-project/alire-index/pull/1210 goes through (I'd be happy to have that test re-ran to confirm).

PR #1211, opened 2024-08-25
Create azip-2.7.0.toml

zertovitch:


zertovitch:
New in version 2.7:

*    Tabs: each archive subwindow has an own tab on a tab bar (new). This makes the handling of AZip easier to understand.
*    Temporary directory: you can optionally set a specific temporary directory - typically, in a RAM disk - for AZip's operations such as archive recompression. Your HD or SSD will thank you.
*    Fix: sorting column and direction are preserved on operations that modify a Zip archive (delete, add, update, recompress, etc.).
*    Single instance: when Zip archives are opened and AZip is already running, those archives are opened in that running instance - thanks to Nicolas Pinault (DrPi) for that feature!

mosteo (approved):

PR #1222, opened 2024-09-07
a0b_callbacks 0.1.0

godunko:


mosteo (approved):

Fabien-Chouteau (approved):
Nice one. It would be great to remove the unused dependency on `a0b_base`.

godunko:
Thanks! 

@Fabien-Chouteau, `a0b_base` provides root package `A0B`, don't know how to remove dependency without moving `A0B.Callbacks` package outside of the `A0B` hierarchy. I don't want to pollute global namespace.

PR #1231, opened 2024-09-22
Add OpenAPI 0.8 crates

stcarrez:


mosteo:
The bug you're hitting here is solved in the development version, which is why the `(alr dev)` tests are passing. The problem only happens when `alr get`ting crates, so they should work properly when used as a dependency. Hence, I can merge if you tell me this is ready.

mosteo (approved):

PR #1235, opened 2024-09-27
Reinstate macos test on master

mosteo:


mosteo:
Seems to have been related to the bad release of gprbuild.

PR #1238, opened 2024-09-29
Add AWA 2.5.0 crate (created in Aug 2023, updated for AWS 24)

stcarrez:


index/aw/awa/awa-2.5.0.toml
```diff
@@ -0,0 +1,90 @@
...
+type = "post-fetch"
+command = ["sh", "./alire-setup.sh"]
+
+[gpr-externals]
+AWA_LIBRARY_TYPE = ["relocatable", "static", "static-pic"]
+AWA_BUILD = ["distrib", "debug", "optimize", "profile", "coverage"]
+AWA_AWS_IMPL = ["none", "1", "2", "3", "4"]
+
+[gpr-set-externals]
+AWA_AWS_IMPL = "4"
```
mosteo, line 81: I presume this must always be set to this value? It is not for clients to decide? Otherwise this should be a default not forced in the manifest.

stcarrez: In fact, it is very useful for a client to decide due to Alire dependency issues. I can ask `AWA_AWS_IMPL=none` and the module will not use AWS API at all.

stcarrez: Forgot to say that I intended to have support for build for Windows with AWS 23, hence I would have defined `AWA_AWS_IMPL=3`. But, I saw the Ada Utility Library AWS package that failed for Windows so I backtracked on this Alire and removed the support for Windows in the hope it would work.

mosteo: I see. But then, if you set it here, users of AWA cannot change it, is that how it's intended to be?

stcarrez: The following works for me:
```
alr build -- -XAWA_AWS_IMPL=none
```

mosteo: Ah, I see. I was thinking in terms of overriding via manifest files which, at the moment, causes a conflict. That's why we recommend not setting parameters in the manifest if a default can be applied transparently e.g. in the project file.

mosteo (approved):

PR #1249, opened 2024-10-17
a0b_i2c 0.1.0

godunko:

### DIFF
--- a/index/a0/a0b_i2c/a0b_i2c-0.1.0.toml
+++ b/index/a0/a0b_i2c/a0b_i2c-0.1.0.toml
@@ -1,0 +1,28 @@
+name = "a0b_i2c"
+description = "A0B I2C API"
+version = "0.1.0"
+
+authors = ["Vadim Godunko"]
+maintainers = ["Vadim Godunko <vgodunko@gmail.com>"]
+maintainers-logins = ["godunko"]
+licenses = "Apache-2.0 WITH LLVM-exception"
+tags = ["a0b", "embedded", "i2c"]
+
+project-files = ["gnat/a0b_i2c.gpr"]
+
+[configuration]
+disabled = true
+
+[[depends-on]]
+a0b_base = "*"
+a0b_callbacks = "*"
+
+[[actions]]
+type = "test"
+directory = "selftest"
+command = ["alr", "build"]
+
+[origin]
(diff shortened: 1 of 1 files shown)


godunko:
Selftest requires `arm-eabi` target  and `light-cortex-m0` runtime, it seems they are not provided by Ubuntu/Debian distro toolchains, and Alire's GNAT is not used for some reason.

Is there some combination of target/runtime that is present on all platforms?

Or is there some configuration parameters to skip builds with distro toolchain in CI?

mosteo:
On cursory look, the failure seems to be that the cross-compiler is installed but the compatible gprbuild, so the distro-provided one is unable to find the compiler. That's for me to fix.

godunko:
Thanks! 

I've submitted one more MR (https://github.com/alire-project/alire-index/pull/1255) and CI works fine on it. The difference is that `a0b_stm32f401` crate has `gnat_arm_elf` in dependencies, while `a0b_i2c` crate doesn't have such dependency and it added by the `selftest` subcrate.

mosteo:
> The difference is that `a0b_stm32f401` crate has `gnat_arm_elf` in dependencies, while `a0b_i2c` crate doesn't have such dependency and it added by the `selftest` subcrate.

Right. The CI check isn't clever enough to know where tests are and doesn't proactively install `gprbuild`. A workaround for the time being would be to add a dependency on `gprbuild` in the testing crate itself, besides on gnat. Still, I want to have a proper solution to this (but you may try it if you want).

godunko:
Adding of

```
gprbuild = "*"
```

has undesirable side effect on my setup, it attempts to install `gprbuild` provided by the system (Ubuntu 22.04), and not provided by Alire :( Adding of

```
gprbuild = "^22.0.1"
```

works on my side, but it seems CI fails on MacOS.

mosteo:
I see. This is because only the latest toolchains work on the latest macOS ARM runners.

Let me see if I can come up with a general fix. If you want the crates merged already, given that we know the issue is unrelated, I can merge them so you don't have to wait.

godunko:
Up to you, I can wait for few days, I have more crates to add selftests and to prepare MRs.

mosteo:
This issue should be fixed for the future, I hope. Thanks for the patience.

mosteo (approved):

PR #1250, opened 2024-10-17
utils 0.1.0

mgharbour:

### DIFF
--- a/index/ut/utils/utils-0.1.0.toml
+++ b/index/ut/utils/utils-0.1.0.toml
@@ -1,0 +1,15 @@
+name = "utils"
+description = "mast utilities"
+version = "0.1.0"
+
+authors = ["Michael Gonzalez"]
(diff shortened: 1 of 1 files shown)


index/ut/utils/utils-0.1.0.toml
```diff
@@ -0,0 +1,15 @@
+name = "utils"
+description = "mast utilities"
```
mosteo, line 2: Hello @mgharbour. Given that "utils" is a very general name, and I guess these utilities are tailored to MAST, I think it would be preferable to name the crate "mast_utils". (Incidentally, I just answered your email; maybe you prefer to discuss a bit before continuing.)

krischik:
If you look at the [crate list](https://alire.ada.dev/crates.html) you want a name and description that sticks out to those who browse the list to see what is available.

mosteo:
After our conversation I think the proper way is to just have a "mast" crate. BTW, @mgharbour, let me know if you need any additional guidance.

PR #1254, opened 2024-10-21
a0b_exti 0.1.0

godunko:


mosteo:
On cursory look, the failure seems to be that the cross-compiler is installed but the gprbuild is not, so the distro-provided one is unable to find the compiler. That's for me to fix.

mosteo (approved):

PR #1259, opened 2024-10-22
a0b_armv7m_systick_clock 0.1.0

godunko:

### DIFF
--- a/index/a0/a0b_armv7m_systick_clock/a0b_armv7m_systick_clock-0.1.0.toml
+++ b/index/a0/a0b_armv7m_systick_clock/a0b_armv7m_systick_clock-0.1.0.toml
@@ -1,0 +1,33 @@
+name = "a0b_armv7m_systick_clock"
+description = "A0B Monotonic Clock by ARMv7M SysTick"
+version = "0.1.0"
+
+authors = ["Vadim Godunko"]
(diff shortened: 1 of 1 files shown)


godunko:
@mosteo failure looks strange, while legitimate. I suppose there was an issue with CI for https://github.com/alire-project/alire-index/pull/1253 where `a0b_time` was introduced. It should fails on Windows with the same error.

Do you have some recommendation to "fix" failure of the DOS line termination style check? I can just suppress it for `a-reatim.adb`, but same style check reported for many other files, so may be I need to configure Git repository instead?

mosteo:
Given that the failure pops up in specification files, maybe the problem is in the project here that's including the others. So I'd start by checking that the particular check is indeed not enabled. I've taken an initial look but I couldn't locate the `a0b_common.gpr` that's supplying the switches. Though, presuming this is shared by all projects, it's puzzling that the problem pops up only here. As you say, it might be related to the repo configuration, but I'm also presuming is the same for all your other repos...

godunko:
`a0b_common.gpr` is in the `a0b_base` crate, see https://github.com/godunko/a0b-base/blob/main/gnat/a0b_common.gpr

Sorry for the noise, it is side effect of the `-gnatg` switch for `a-reatim.adb` file, compiler checks all files `with`-ed by it.

mosteo (approved):

PR #1262, opened 2024-10-22
light_blackpill_stm32f401 0.1.0

godunko:

### DIFF
--- a/index/li/light_blackpill_stm32f401/light_blackpill_stm32f401-0.1.0.toml
+++ b/index/li/light_blackpill_stm32f401/light_blackpill_stm32f401-0.1.0.toml
@@ -1,0 +1,27 @@
+name = "light_blackpill_stm32f401"
+description = "BlackPill STM32F401 BSP for `light` GNAT Runtime"
+version = "0.1.0"
+
+authors = ["Vadim Godunko"]
+maintainers = ["Vadim Godunko <vgodunko@gmail.com>"]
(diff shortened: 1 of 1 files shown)


index/li/light_blackpill_stm32f401/light_blackpill_stm32f401-0.1.0.toml
```diff
@@ -0,0 +1,27 @@
+name = "light_blackpill_stm32f401"
+description = "BlackPill STM32F401 BSP for `light` GNAT Runtime"
```
mosteo, line 2: Would perhaps be better to use `blackpill_stm32f401_light` so the main information comes first? Also thinking of the alphabetical ordering when crates are listed.

godunko: I've few BSP packages for use with `light-cortex-*`, their names starts with `light` to highlight this:

 * light_arduino_due
 * light_blackpill_stm32f401
 * light_fk723m1_zgt6
 * light_stm32f4_disco

I don't have really strong opinion about naming, so can rename them.

mosteo: ah, then it doesn't make much sense to change just this one

mosteo (approved):

PR #1269, opened 2024-10-24
cortex_m 1.0.0

Fabien-Chouteau:


Fabien-Chouteau:
@mosteo This crate has a config variable without default. That's why the builds fail.

mosteo (approved):

mosteo:
> @mosteo This crate has a config variable without default. That's why the builds fail.

For these cases the advice is to have a nested crate that supplies the missing info and is configured as a test action. In case you want to add this in the future.

Fabien-Chouteau:
I do have a nested crate. How do I specify it should be used for testing?

mosteo:
Then you add this to your main crate:
```
[[actions]]
type = "test"
directory = "your/nested/crate/path"
command = ["alr", "build"]
```
The `alr build` could be an `alr run` if you want to run unit tests or something.

PR #1270, opened 2024-10-24
Update cortex_m-1.0.0.toml

Fabien-Chouteau:


Fabien-Chouteau:
Still not perfect because of libraries on macOS, but it's better.

mosteo (approved):

PR #1271, opened 2024-10-24
rp2040_hal 2.4.0

JeremyGrosser:


index/rp/rp2040_hal/rp2040_hal-2.4.0.toml
```diff
@@ -0,0 +1,31 @@
...
+maintainers-logins = ["JeremyGrosser"]
+tags = ["embedded", "nostd", "rp2040", "raspberrypi", "drivers"]
+website = "https://pico-doc.synack.me/"
+
+[[depends-on]]
+cortex_m = "^1"
+atomic = "^1"
+hal = "^1"
+usb_embedded = "^1"
+gnat_arm_elf = "^14"
```
mosteo, line 17: Hello @JeremyGrosser, I wonder if this dependency on GNAT should better be ">=14".

JeremyGrosser: I can’t guarantee it’ll work with future versions without testing.

mosteo: That's right, of course. I mention this because I've seen incompatibilities due to crates forcing different major versions.

But, if you prefer the conservative approach, that's fine with me.

mosteo (approved):

PR #1273, opened 2024-10-25
gnatcoll, libgpr, xmlada 25.0.0

Fabien-Chouteau:


Fabien-Chouteau:
@mosteo can you have a look at the `debian-stable` run? I specified a `>=13` dependency on GNAT but the run shows a GNAT bug box on a GNAT version `12.2.0`.

mosteo:
I'm on it. I'm able to reproduce the issue locally too.

mosteo:
I understand the problem now: since there is no dependency on `gprbuild`, the system one is being used, which doesn't see the intended gnat even if it is first in path. 

The temporary solution is to add a dependency on `gprbuild<2000` (to avoid the one in the system), but the root problem should be addressed in Alire itself to inject a dependency on a indexed gprbuild whenever an indexed gnat is in use.

Fabien-Chouteau:
It would be strange to add such constraint in the gnatcoll release. Are you OK to merge the release as is?

mosteo:
> It would be strange to add such constraint in the gnatcoll release. Are you OK to merge the release as is?

Yes, this won't affect people with a selected toolchain anyway.

mosteo (approved):

PR #1279, opened 2024-10-28
gnatcoll bindings and DB 25.0.0

Fabien-Chouteau:


Fabien-Chouteau:
@mosteo we will have the same problem for every crate that depends on gnatcoll. :(

mosteo:
Ah, right... I can hardcode a workaround in the checks for the time being.

mosteo (approved):

PR #1290, opened 2024-11-13
AWS 25.0.0

Fabien-Chouteau:


bracke:
Is this going to be fixed? It prevents using the other AdaCore libraries  @2025, since aws is only available as a 2024 crate.

PR #1296, opened 2024-11-18
hp41cx_tools 1.6.0

krischik:

Tools for manipulating memory dumps from HP-41CX emulators.

The following HP-41CX emulators are supported:

* [PX-41CX](https://paxer.net/PX-41CX/) from Paxer.
* [DM41X](https://www.swissmicros.com/product/dm41x) from SwissMicros.

Currently hex dump files can be decoded to user readable files.


krischik:
Windows and it's 🤬 CR/LF. No one needs that any more. Even NotePad can do without.

PR #1310, opened 2024-12-01
minimal_containers 1.2.1

simonjwright:


simonjwright:
The reason for this is that the commit reference in 1.2.0 was wrong. No idea how that happened.

I suppose I could have deleted a branch in my own repo? but surely I’d only publish from 'main' (or 'master').

Fabien-Chouteau (approved):

PR #1312, opened 2024-12-04
adacl 6.0.0

krischik:

* Wide_GetOpt: correct encoding for windows which also uses UTF-8 with Ada.
* Wide_GetOpt: resolve ambiguous naming for Raise_Exeption
* Wide_GetOpt: add missing string utilities.
* Wide_GetOpt: Both trace and getopt are now unicode compatible.
* Wide_GetOpt: Next try using a mix of UTF_String and Wide_Wide_String. Won't need additional Wide and Wide_Wide classes any more.


krischik:
The price of having over 300 unit tests: They might fail.

PR #1318, opened 2024-12-07
nrf52-runtimes 14.1.0

damaki:
This adds v14.1.0 of the light, light-tasking, and embedded runtimes for some nRF52 SoCs.

### DIFF
--- a/index/em/embedded_nrf52832/embedded_nrf52832-14.1.0.toml
+++ b/index/em/embedded_nrf52832/embedded_nrf52832-14.1.0.toml
@@ -1,0 +1,56 @@
+name = "embedded_nrf52832"
+description = "embedded runtime for the nRF52832 SoC"
+version = "14.1.0"
+
+long-description = """
+## Usage
+
+First edit your `alire.toml` file and add the following elements:
+ - Add `embedded_nrf52832` in the dependency list:
+   ```toml
+   [[depends-on]]
+   embedded_nrf52832 = "*"
+   ```
+
+Then edit your project file to add the following elements:
+ - "with" the run-time project file. With this, gprbuild will compile the run-time before your application
+   ```ada
+   with "runtime_build.gpr";
+   ```
+ - Specify the `Target` and `Runtime` attributes:
+   ```ada
+      for Target use runtime_build'Target;
+      for Runtime ("Ada") use runtime_build'Runtime ("Ada");
+   ```
+"""
+
+authors = ["AdaCore", "Daniel King"]
+maintainers = ["Daniel King <damaki.gh@gmail.com>"]
+maintainers-logins = ["damaki"]
+licenses = "GPL-3.0-or-later WITH GCC-exception-3.1"
+tags = ["embedded", "runtime"]
+website = "https://github.com/damaki/nrf52-runtimes"
+
+project-files = ['runtime_build.gpr', 'ravenscar_build.gpr']
+
+[configuration]
+generate_gpr = false
+generate_c = false
+output_dir = "gnat_user"
+
+[configuration.variables]
+Time_Base = { type = "Enum", values = ["RTC0", "RTC1", "RTC2"], default = "RTC2" }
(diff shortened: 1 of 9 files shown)


index/em/embedded_nrf52832/embedded_nrf52832-14.1.0.toml
```diff
@@ -0,0 +1,56 @@
...
+output_dir = "gnat_user"
+
+[configuration.variables]
+Time_Base = { type = "Enum", values = ["RTC0", "RTC1", "RTC2"], default = "RTC2" }
+LFCLK_Src = { type = "Enum", values = ["Xtal", "RC", "Synth"], default = "Xtal" }
+Use_SWO_Trace = { type = "Boolean", default = true }
+Use_Reset_Pin = { type = "Boolean", default = true }
+
+[[depends-on]]
+gnat_arm_elf = "^14"
```
mosteo, line 48: I wonder if you might prefer ">=14". Although later GNAT versions might break something, this way you're anchoring all your clients to GNAT 14.

damaki: Thanks. My understanding is that the runtimes are generally tied to a particular GNAT version, and later GNAT versions might break something as you say, since the compiler/runtime interface could change with diffferent GNAT versions. 

My plan is to do one major release of these runtimes for each GNAT version. This version (14.1.0) is compatible with gnat-fsf-14. When gnat-fsf-15 is released then I'll do a 15.0.0, and I'm going to try doing releases for earlier GNAT versions too; this version is just the first attempt. This way users can either let the GNAT version be chosen automatically:
```toml
[[depends-on]]
light_nrf52840 = ">=14"
```

or let the runtime version be chosen automatically:
```toml
[[depends-on]]
gnat_arm_elf = ">=14"
light_nrf52840 = "*"
```

Fabien-Chouteau: Indeed, a given run-time can be strongly tied to a compiler version.

mosteo: OK, I didn't know this info about runtimes, I was only thinking of checks that find new errors. Then it must be like you have done it.

index/em/embedded_nrf52832/embedded_nrf52832-14.1.0.toml
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+authors = ["AdaCore", "Daniel King"]
+maintainers = ["Daniel King <damaki.gh@gmail.com>"]
+maintainers-logins = ["damaki"]
+licenses = "GPL-3.0-or-later WITH GCC-exception-3.1"
+tags = ["embedded", "runtime"]
+website = "https://github.com/damaki/nrf52-runtimes"
+
+project-files = ['runtime_build.gpr', 'ravenscar_build.gpr']
```
mosteo, line 34: Another clarification for my peace of mind: if you specify two project files, both will always be built by dependents. If those represent alternatives, they should go in separate crates.

damaki: Thanks for clarifying that. This is the intended behaviour (that both will be built). 

The two project files are really just a consequence of how the runtimes are organised where it's split into 2 libraries: libgnat (built by runtime_build.gpr) for the sequential stuff, and libgnarl (built by ravenscar_build.gpr) for the tasking stuff. The light runtimes don't support tasking, so only provide libgnat (so the light_nrfxxx runtimes only build runtime_build.gpr), whereas light-tasking and embedded provide both libgnat and libgnarl (so both projects are built for those runtimes).

mosteo: OK, thanks. Then I think this is ready to be merged.

mosteo (approved):

PR #1322, opened 2024-12-10
Removed support for obsolete 32-bit toolchains.

pmunts:
MuntsOS Embedded Linux no longer supports 32-bit targets (BeagleBone, Raspberry Pi 1, and Raspberry Pi 2).


Fabien-Chouteau:
Hi @pmunts , it is not allowed to remove releases from the index as some users may depend on it.

PR #1325, opened 2024-12-11
deepend 3.12.0

beemirang:

### DIFF
--- a/index/de/deepend/deepend-3.12.0.toml
+++ b/index/de/deepend/deepend-3.12.0.toml
@@ -1,0 +1,41 @@
+name = "deepend"
+description = "Bounded & Unbounded memory storage pools with subpool capabilities"
+version = "3.12"
+
+authors = ["Brad Moore"]
+maintainers = ["Brad Moore <bmoore.ada@gmail.com>"]
+maintainers-logins = ["beemirang"]
+licenses = "GPL-3.0-or-later WITH GCC-exception-3.1"
+website = "https://github.com/beemirang/deepend"
+tags = ["ada", "memory", "storage-pools", "subpools"]
+project-files = ["deepend.gpr"]
+
+executables = ["test_dynamic_pools_ada95",
+               "test_dynamic_pools_ada2005",
+               "test_dynamic_pools_ada2012",
+               "test_dynamic_pools_ada2022",
+               "test_bounded_dynamic_pools_ada95",
+               "test_bounded_dynamic_pools_ada2005",
+               "test_bounded_dynamic_pools_ada2012",
+               "test_bounded_dynamic_pools_ada2022",
+               "binary_trees_basic_pool_ada95",
+               "binary_trees_basic_pool_ada2005",
+               "binary_trees_basic_pool_ada2012",
+               "binary_trees_basic_pool_ada2022",
+               "binary_trees_basic_bounded_pool_ada95",
+               "binary_trees_basic_bounded_pool_ada2005",
+               "binary_trees_basic_bounded_pool_ada2012",
+               "binary_trees_basic_bounded_pool_ada2022",
+               "binary_trees_with_subpools_ada95",
+               "binary_trees_with_subpools_ada2005",
+               "binary_trees_with_subpools_ada2012",
+               "binary_trees_with_subpools_ada2022",
+               "binary_trees_with_bounded_subpools_ada95",
+               "binary_trees_with_bounded_subpools_ada2005",
+               "binary_trees_with_bounded_subpools_ada2012",
+               "binary_trees_with_bounded_subpools_ada2022"]
+
+[origin]
+commit = "df89d2652bc10e1ce90e607a71ab97e9c87a87db"
+url = "git+https://github.com/beemirang/deepend.git"
+


mosteo:
According to logs:

> FAIL: Declared executable not found after compilation: test_dynamic_pools_ada95

beemirang:
Hello, I believe I have fixed the issue with the build. The build can be set to build executables and libraries for either Ada 95, Ada 2005, Ada 2012, or Ada 2022. By default it only builds the Ada 2012 variant, (controllable by environment variables). The intention for Alire is that the Ada 2012 variant is what you get when you pull in the crate. I updated the .toml file to list only those executables that get built for the Ada 2012 variant. So, if you run the build now, you should see all executables produced that are mentioned in the .toml file.

beemirang:
The exe list still wasn't correct, because you can also select whether you want bounded or unbounded memory pools, and for each of these, whether you want subpool capable or not.  The default is that you get bounded with subpool capability. The intent with alire is that that is what you get with the crate. I have now updated the .toml file to list only the test executables produced for the Ada 2012, bounded pools with subpools capabilities. It should work now.

index/de/deepend/deepend-3.12.0.toml
```diff
@@ -0,0 +1,41 @@
...
+               "binary_trees_basic_bounded_pool_ada2012",
+               "binary_trees_basic_bounded_pool_ada2022",
+               "binary_trees_with_subpools_ada95",
+               "binary_trees_with_subpools_ada2005",
+               "binary_trees_with_subpools_ada2012",
+               "binary_trees_with_subpools_ada2022",
+               "binary_trees_with_bounded_subpools_ada95",
+               "binary_trees_with_bounded_subpools_ada2005",
+               "binary_trees_with_bounded_subpools_ada2012",
+               "binary_trees_with_bounded_subpools_ada2022"]
```
mosteo, line 36: There must have been some confusion with the latest changes as I still see all the executables listed here (and, consequently, the tests fail).

Note that you can override the default test with an `[[actions]]` (or several)  [entry](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#release-information). That way you could declare all executables, or verify with each set of environment configurations with a bit of scripting.

But if you want to just list the default builds as you described that is also OK.

beemirang: I would have preferred to build more of the executables, but when I try to do that, alr build runs into parallel compilation failures. It seems to want a -j1 to specify that only one processor is used, but when I haven't had any luck attempting to change that, alire seems to override whatever I put in the project file. Maybe I can get things to work if my test programs try to link to the deepend as a library, rather than each one including the sources. I will look into that to see if that might work. I did check in a change to the alire.toml file, to only list the exes that are to be built, but perhaps your merge request doesn't have the latest sources?

beemirang:
I found I was able to fix the parallel build issues, by importing the main deepend_lib.gpr project into each of the test programs. Now all executables can be built at once in Alire, and so I have changed the defaults to build everything. Also, I updated the alire.toml file to list all these executables, like it was originally. Hopefully everything builds for you now.

beemirang:
Hi Alejandro,

I think I have fixed the build issues on my end. I found that importing the main library in all the
test programs changed the compilation so that there were no parallel build issues.
This is all checked into git hub, so I think if you tried the build again it should work.

Now the alire build builds everything, and I updated the alire.toml file to list all the executables.
THis is also checked in on git hub

Is there something more I need to do on my end?

Thanks,

Brad

----- Original Message -----

mosteo:
Hi Brad, I think there's some misconception on your side at play. If all executables are not built simultaneously, our default check will never succeed, because all listed executables are checked immediately after `alr build`.

Understanding that, you have to choices: 

1) I can merge as-is, as the library is actually correct, so there's no point in delaying it more. The drawback is that this will happen in future versions too, and one has to check manually every configuration to ascertain that there's actually no error.

2) Or, you can provide a custom test that does whatever you want to "approve" the build of the crate. This is done by defining a test action that invokes whatever you want. (Look for actions in the [catalog spec](https://github.com/alire-project/alire/blob/release/2.1/doc/catalog-format-spec.md#release-information).)

Of course I can merge now and you can prepare the test for any future submissions.

Just let me know how you want to proceed.

beemirang:
Hi Alejandro,
I believe I have fixed the parallel build issues with the most recent check-ins that I have made.
That is, you should now be able to build all executables simultaneously, so 
that sounds to me like the default check will succeed, once you pull in these changes.

Or are you saying that having that many executables produced by the build requires
too much overhead on your end, to verify that everything built correctly?

I am happy to release what I have now, but am also happy to work towards reducing the build to a single
executable, if need be, for future builds if that is what you would like to see for the alire package framework.

I will check out the link you sent below.

I am already planning to make more enhancements in the future. In particular;
  - Use a hash table lookup to find the subpools to be deallocated
  - Make another variant of bounded subpools that is more bounded than the current ones, which are really, semi-bounded.

Cheers,

Brad 
----- Original Message -----

mosteo:
> you should now be able to build all executables simultaneously, so
that sounds to me like the default check will succeed, once you pull in these changes.

You have to update this PR to point to your new commit (or create a new PR), we can't pull your changes.

> Or are you saying that having that many executables produced by the build requires
too much overhead on your end, to verify that everything built correctly?

No, it is not a problem of overhead. By default, we check that all listed executables exist after a single `alr build`. If you want a different behavior, you can override via test actions in your crate. This is only what we do by default here. Any kind of testing you want for your crate is valid.

> I am happy to release what I have now, but am also happy to work towards reducing the build to a single
executable, if need be, for future builds if that is what you would like to see for the alire package framework.

You release what and when you see fit, we do not want to impose anything on you :)

beemirang:
Hi Alejandro,

I closed the PR I had, and tried to generate a new one with
   alr publish

and it gets to the last step but then fails. 

I get the following error message:

Publishing assistant: step 10 of 11: Upload manifest
[release/deepend-3.12.0 a99b1ad] deepend 3.12.0 (via `alr publish --submit`)
 1 file changed, 43 insertions(+)
 create mode 100644 index/de/deepend/deepend-3.12.0.toml
remote: Permission to beemirang/alire-index.git denied to beemirang.
fatal: unable to access 'https://github.com/beemirang/alire-index/': The requested URL returned error: 403
error:    Command ["git", "push", "-f", "-u", "writable", "release/deepend-3.12.0"] exited with code 128


It appears I am being denied access to the alire-index.
I am not sure what to do at this point. Any suggestions?

Thanks,
Brad

----- Original Message -----

mosteo:
That's your fork of the index, and indeed I see the manifest at the expected location in the expected branch: https://github.com/beemirang/alire-index/blob/release/deepend-3.12.0/index/de/deepend/deepend-3.12.0.toml

If that manifest is correct, you could just open the PR manually now, from that fork. If not, you could try deleting that branch and try publishing again.

PR #1330, opened 2024-12-16
adacl_eastrings 6.1.1

krischik:

Found a workaround for that pesky GNAT bug I opened years ago. Instead of using constant I now use an inline function which always returns the same value and set the post condition to check that value. Annoying but no one worked on that bug and now I can finally release EAStrings.


Fabien-Chouteau (approved):

Fabien-Chouteau (changes requested):
There's an issue on Windows:
```
D:\a\alire-index\alire-index\adacl_eastrings_6.1.1_5884b939\src\windows\adacl-eastrings_unix_interface.adb:4:06: error: file "adacl-interfaces.ads" not found
```

krischik:
Have to try again.

PR #1334, opened 2024-12-16
adacl_eastrings 6.1.1

krischik:

Found a workaround for that pesky GNAT bug I opened years ago. Instead of using constant I now use an inline function which always returns the same value and set the post condition to check that value. Annoying but no one worked on that bug and now I can finally release EAStrings.

Fixed Windows.


krischik:
Sorry that I missed the Windows failure. Fixed now.

mosteo (approved):

PR #1340, opened 2024-12-25
adagl_gtk3 0.0.1

godunko:


godunko:
CI fails on Debian due to use of too old compiler (12.2). VSS requires GCC >= 14.

mosteo (approved):

PR #1342, opened 2024-12-26
adacl 6.2.0

krischik:

Add a queue specifically designed for multi process use. It addition to Ada.Containers.Synchronized_Queue_Interfaces it offers and end of queue flag, wait for all elements processed and premature abort command.

### DIFF
--- a/index/ad/adacl/adacl-6.2.0.toml
+++ b/index/ad/adacl/adacl-6.2.0.toml
@@ -1,0 +1,75 @@
+name                        = "adacl"
+description                 = "Ada Class Library (String, Trace, AUnit, Smart Pointer. GetOpt)"
+long-description            = """A class library for Ada for those who like OO programming.
+
+Currently the following functionality is migrated to Ada 2022:
+
+* Getopt commandline argument parser - with wide character support.
+* String utilities - with wide character support.
+* Calendar utilities - with wide character support.
+* Trace utility - with wide character support.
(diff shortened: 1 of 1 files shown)


index/ad/adacl/adacl-6.2.0.toml
```diff
@@ -0,0 +1,75 @@
...
+development.runtime_checks  = "Overflow"
+release.compile_checks      = "Warnings"
+release.contracts           = "No"
+release.runtime_checks      = "Default"
+validation.compile_checks   = "Warnings"
+validation.contracts        = "Yes"
+validation.runtime_checks   = "Everything"
+
+[[depends-on]]
+gnat_native                 = "^14.2"
```
mosteo, line 57: Is this really intended to be `gnat_native` instead of just `gnat`? With the latter, the library could be used with cross-compilers.

krischik: The crate will only work for Windows, Linux and macOS and I'm indeed unsure about about cross compiler. And then there was a problem with macOS picking a cross compiler even when native compilation was requested. See:

https://matrix.to/#/!DNeiNadAgLpbmdDKov:gitter.im/$MN8KphfRswrJ8Bw66FHWL2mVM_Xyd1eOgtZ9jFVrbSI?via=gitter.im (of you can make the link work)

So, yes, that was intentional.

mosteo (approved):

PR #1364, opened 2025-01-15
byteflippers 1.1.0

denkisikinoneko:


denkisikinoneko:
... interesting, it appears the macOS versions don't have support for explicitly 128-bit (18 digit) floats; pushing an update ....

mosteo (approved):

mosteo:
> ... interesting, it appears the macOS versions don't have support for explicitly 128-bit (18 digit) floats; pushing an update

Ada's portability + checks in several OSes/archs = peace of mind

PR #1380, opened 2025-01-30
bigints 0.1.0

AldanTanneo:


AldanTanneo:
@Fabien-Chouteau

index/bi/bigints/bigints-0.1.0-dev.toml
```diff
@@ -0,0 +1,23 @@
+name = "bigints"
+description = ""
+version = "0.1.0-dev"
```
mosteo: We usually don't want "interim" developer versions in the index, unless there is strong justification. This doesn't mean that partial or experimental versions aren't welcome. In other words: could this simply be `0.1.0` without the `-dev`?

index/bi/bigints/bigints-0.1.0-dev.toml
```diff
@@ -0,0 +1,23 @@
...
+website = "https://github.com/AldanTanneo/bigints"
+tags = []
+
+[build-switches]
+"*".ada_version = "Ada2022"
+development.optimization = ["-O3"]
+release.runtime_checks = "none"
+
+[[depends-on]]
+gnatprove = "^14.1"
```
mosteo: Dependencies used for testing only should be moved to a nested crate that performs the testing of the main crate, to avoid clients depending on `gnatprove` too.

index/bi/bigints/bigints-0.1.0-dev.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[origin]
+commit = "3eb781f188b77782be5be50a489a88ef0e236b2c"
+url = "https://github.com/AldanTanneo/bigints.git"
```
mosteo: This origin has been likely generated with a developer version of Alire and is not backward compatible (the reason for all test failures). You need to use `git+https` instead. (Incidentally, this has been recently fixed in `master`.)

AldanTanneo:
Addressed all comments :)

index/bi/bigints/bigints-0.1.0.toml
```diff
@@ -0,0 +1,20 @@
+name = "bigints"
+description = "Experimental SPARK Constant Time Big Integer library"
+version = "0.1.0"
+
+authors = ["Cesar SAGAERT"]
+maintainers = ["Cesar SAGAERT <sagaert@adacore.com>"]
+maintainers-logins = ["AldanTanneo"]
+licenses = "MIT"
+website = "https://github.com/AldanTanneo/bigints"
+tags = []
```
mosteo: Sorry for missing this the first time around. A couple of tags at least are strongly recommended, as that will increase visibility of the crate in searches and in `alire.ada.dev`.

AldanTanneo: added some tags :)

mosteo (approved):

PR #1386, opened 2025-02-13
bbs_lisp 0.2.0

BrentSeidel:


BrentSeidel:
This will allow a client program to provide an input file to be read by the Lisp parser instead of having it all entered on the command line.

mosteo (approved):

PR #1387, opened 2025-02-13
bbs_simcpu 0.3.1

BrentSeidel:


BrentSeidel:
bbs_simcpu.gpr didn't include the 6502 source files, so alr build probably wouldn't include them.

mosteo (approved):

PR #1388, opened 2025-02-14
simcpucli 0.3.0

BrentSeidel:


BrentSeidel:
Update to include 6502 simulator and allow Lisp code to be run from a file.

mosteo (approved):

PR #1398, opened 2025-03-05
prolog 1.0.0

jquorning:

### DIFF
--- a/index/pr/prolog/prolog-1.0.0.toml
+++ b/index/pr/prolog/prolog-1.0.0.toml
@@ -12,8 +12,6 @@ licenses = "custom-anna"
 website = "https://github.com/jquorning/Prolog"
 tags = ["prolog", "logic"]
 
-project-files = ["prolog_demo.gpr", "prolog_test.gpr"]
-
 [origin]
 commit = "6ac52773fdb45c0100152081780b2e31b5802986"
 url = "git+https://github.com/jquorning/Prolog.git"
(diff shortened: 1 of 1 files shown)


index/pr/prolog/prolog-1.0.0.toml
```diff
@@ -0,0 +1,20 @@
...
+           "Dale Stanbrough <dale@cs.rmit.edu.au>"]
+	   
+maintainers = ["Jesper Quorning <jesper.quorning@gmail.com>"]
+maintainers-logins = ["jquorning"]
+
+licenses = "custom-anna"
+website = "https://github.com/jquorning/Prolog"
+tags = ["prolog", "logic"]
+
+project-files = ["prolog_demo.gpr", "prolog_test.gpr"]
```
mosteo: I see in the repo that the main project file is just `prolog.gpr`, so these shouldn't really be listed here as they're not directly useful or required by clients of the crate. The `prolog.gpr` one can be given explicitly or, if `project-files` is removed, it will be the one used by default.

Ideally, these could be referenced in nested crates, but that's not mandatory for publishing.

mosteo (changes requested):
This is great, I loved Prolog back in the day.

Just a minor issue to fix with the project files.

index/pr/prolog/prolog-1.0.0.toml
jquorning: project-files removed.

mosteo (approved):

PR #1400, opened 2025-03-06
loadcpm 0.2.1

BrentSeidel:


BrentSeidel:
Generated boot file now supports 16 bit track and sector numbers.

mosteo (approved):

PR #1402, opened 2025-03-06
bbs_simcpu 0.3.2

BrentSeidel:


BrentSeidel:
Update disk controller to use 16 bit track and sector numbers.

mosteo (approved):

PR #1417, opened 2025-03-08
fix: improper syntax that will be rejected post-2.1

mosteo:
We had a missing check in index loading logic that should have rejected  an unexpected array.


Seb-MCaw:
Do you intend to backport this to the `1.1`, `1.2`, and `1.2.1` branches, @mosteo, or should we update `Alire.Index.Min_Compatible_Version_Str`?

mosteo:
Good observation. The thing is that website array should never have been accepted even in those versions, so I'll rather backport the simple fix.

PR #1433, opened 2025-03-14
blink_led_weact_stm32g474 0.1.0

godunko:


krischik:
It's probably be to late but it might have been better to have the common part of the crate name at the beginning. Remember the crates will be added to the [Crates](https://alire.ada.dev/crates.html) list where they will be sorted alphabetically. If you put the board name at the beginning all creates for that board will be placed together like this:

image

This could improve discoverability.

mosteo:
@godunko you ok with the current name?

godunko:
I think it is a bit late to rename crates :( There are few `blink_led` and `button_led` crates already in the index. And even worse, they are mentioned in Ada learning materials out of my control.

mosteo (approved):

PR #1445, opened 2025-03-20
fix: improper syntax that will be rejected post-2.1 (#1417)

mosteo:
We had a missing check in index loading logic that should have rejected an unexpected array.


mosteo:
Expected failures on older runners, index is OK.

PR #1450, opened 2025-03-21
ada_sqlite3 0.1.0

gtnoble:

### DIFF
--- a/index/ad/ada_sqlite3/ada_sqlite3-0.1.0.toml
+++ b/index/ad/ada_sqlite3/ada_sqlite3-0.1.0.toml
@@ -1,0 +1,27 @@
+name = "ada_sqlite3"
+description = "Sqlite3 bindings for Ada with embedded SQLite engine"
+version = "0.1.0"
+
+authors = ["Garret Noble"]
+maintainers = ["Garret Noble <garretnoble@gmail.com>"]
+maintainers-logins = ["gtnoble"]
+licenses = "MIT OR Apache-2.0 WITH LLVM-exception"
+website = ""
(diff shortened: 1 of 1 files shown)


index/ad/ada_sqlite3/ada_sqlite3-0.1.0.toml
```diff
@@ -0,0 +1,27 @@
...
+
+authors = ["Garret Noble"]
+maintainers = ["Garret Noble <garretnoble@gmail.com>"]
+maintainers-logins = ["gtnoble"]
+licenses = "MIT OR Apache-2.0 WITH LLVM-exception"
+website = ""
+tags = ["sqlite", "database", "sql", "embedded"]
+
+[[depends-on]]
+aunit = "25.0.0"
```
mosteo, line 13: I presume this is used for some kind of testing. If so, since clients of your library do not need such dependency in their projects, `aunit` and related testing should be moved to a [nested crate](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing).

gtnoble:
The tests were already in a nested crate, I just forgot to remove that dependency. After I make the fix, do I need to close this and make a new release, or is there a way to amend the release?

mosteo:
> do I need to close this and make a new release, or is there a way to amend the release?

You can edit on the branch and push commits with the new information.

mosteo:
@gtnoble is it then just a matter of removing the dependency on aunit? I can do that if that's the only thing needed.

gtnoble:
Created a new pull request with the fix. #1599

PR #1457, opened 2025-03-27
cortex_m 1.1.0

Fabien-Chouteau:


Fabien-Chouteau:
Builds will fail because of configuration variable without a default.

PR #1458, opened 2025-03-31
bbs_ansiterm 0.1.0

BrentSeidel:

### DIFF
--- a/index/bb/bbs_ansiterm/bbs_ansiterm-0.1.0.toml
+++ b/index/bb/bbs_ansiterm/bbs_ansiterm-0.1.0.toml
@@ -1,0 +1,23 @@
+name = "bbs_ansiterm"
+description = "Generate Escape Sequences for ANSI terminals"
+version = "0.1.0"
+
+long-description="""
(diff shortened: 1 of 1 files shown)


BrentSeidel:
Are these build failures expected or not?

mosteo:
Looks suspicious but I can't pinpoint an obvious reason. Please open the same PR against base branch `stable-1.4.0`.

BrentSeidel:
I think I managed to open it agains stable-1.4.0 and it looks like similar failures.  But I see some problems in the *.gpr files - maybe not something that's causing the issues here, but shows that I copied the files from bbs_simcpu.  Let me close these PRs, fix the *.gpr files, and try again.

BrentSeidel:
After playing around with this, updating to the latest ALR, and selecting an updated toolchain, it seems that the problem occurs between gprbuild v22.0.1 which works and v24.0.1 which gives the, "bbs_ansiterm.gpr:3:17: warning: there are no sources of language "Ada" in this project" error.

PR #1460, opened 2025-04-01
bbs_ansiterm 0.1.0

BrentSeidel:


BrentSeidel:
It looks like the problem last time I tried to submit this was that the source file names were upper case.  I think that the latest gprbuild is a little more picky about that.

mosteo:
> It looks like the problem last time I tried to submit this was that the source file names were upper case. I think that the latest gprbuild is a little more picky about that.

OK, good to know

mosteo (approved):

PR #1461, opened 2025-04-02
Update lea-0.92.0.toml

zertovitch:
Updated version number of a dependency, Zip-Ada (removes warnings & style notes).

### DIFF
--- a/index/le/lea/lea-0.92.0.toml
+++ b/index/le/lea/lea-0.92.0.toml
@@ -56,7 +56,7 @@ ini_files = "^10.0.0"
 [[depends-on]]
 gwindows = "^1.5.0"
 [[depends-on]]
-zipada = "^58.0.0"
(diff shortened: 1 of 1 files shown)


index/le/lea/lea-0.92.0.toml
```diff
@@ -56,7 +56,7 @@ ini_files = "^10.0.0"
 [[depends-on]]
 gwindows = "^1.5.0"
 [[depends-on]]
-zipada = "^58.0.0"
+zipada = "^61.0.0"
```
mosteo, line 59: Changing dependencies of an existing release is discouraged as this is essentially changing the `lea 0.92` build. The recommendation would be to make for example a patch release (`lea 0.92.1`) if there are no other changes.

zertovitch: Fine, I do it.

PR #1465, opened 2025-04-05
tiny_rp2040_bsp 1.0.0

hgrodriguez:


hgrodriguez:
this is just to use the latest rp2040_hal and bumping up the version number

mosteo (approved):

PR #1472, opened 2025-04-08
tecs 0.1.0-dev

tsobako:


index/te/tecs/tecs-0.1.0-dev.toml
```diff
@@ -0,0 +1,15 @@
+name = "tecs"
+description = "A library with another ECS implmentation"
+version = "0.1.0-dev"
```
mosteo, line 3: The `-dev` should be removed unless it is deliberate.

mosteo (changes requested):

PR #1473, opened 2025-04-11
pager2 2.0.0

poldy:

### DIFF
--- a/index/pa/pager2/pager2-2.0.0.toml
+++ b/index/pa/pager2/pager2-2.0.0.toml
@@ -1,0 +1,18 @@
+name = "pager2"
+description = "Manipulate \"paged files\" found on the Ada Software Repository"
+version = "2.0"
+
+authors = ["Darren Bane"]
+maintainers = ["Darren Bane <darren.bane.ac@posteo.net>"]
+maintainers-logins = ["poldy"]
+licenses = "custom-CC-PDM-1.0"
+website = ""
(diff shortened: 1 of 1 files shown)


index/pa/pager2/pager2-2.0.0.toml
```diff
@@ -0,0 +1,18 @@
+name = "pager2"
+description = "Manipulate \"paged files\" found on the Ada Software Repository"
+version = "2.0"
+
+authors = ["Darren Bane"]
+maintainers = ["Darren Bane <darren.bane.ac@posteo.net>"]
+maintainers-logins = ["poldy"]
+licenses = "custom-CC-PDM-1.0"
+website = ""
+tags = ["application", "archive"]
```
mosteo, line 10: Those tags don't seem representative of the crate, are they the intended ones?

poldy: It's my first time contributing to Alire so I'm still trying to get the hang of it. These were the best tags I could think of. It's an application (not a library); and its task is similar to zip (actually, more similar to the old shar utility), i.e. it creates or extracts archives of text files. Feel free to change if you think of something better.

mosteo: alright, it's no big deal, if you're happy with those.

mosteo (approved):

PR #1478, opened 2025-04-18
hostarm 25.0.1

jquorning:


mosteo:
I see two different errors: on ARM machines, an x86_64 dir seems to be wanted for some operation.

`sed` on some platforms is complaining about unknown switches.

jquorning:
`sed -s` is a GNU extension not working on Mac.
This was fixed in AWS 25 which stalled and not merged.

PR #1486, opened 2025-05-02
coap_spark 0.9.0

mgrojo:

### DIFF
--- a/index/co/coap_spark/coap_spark-0.9.0.toml
+++ b/index/co/coap_spark/coap_spark-0.9.0.toml
@@ -1,0 +1,69 @@
+name = "coap_spark"
+description = "CoAP implementation formally verified with SPARK/Ada"
+long-description = """
+CoAP-SPARK is a library implementing the Constrained Application Protocol (CoAP)
+as defined in
+[RFC 7252](https://www.rfc-editor.org/rfc/rfc7252), developed in the SPARK 
+language, the formally verified subset of the Ada programming language.
+
+This version implements the client side of the protocol with some limitations:
+- It does not support block-wise transfers.
+- It does not support retransmission of messages.
+- It only supports NoSec and PreSharedKey security modes.
+
+See LICENSING for licensing information.
+"""
+version = "0.9.0"
+
+authors = ["Manuel Gomez"]
+maintainers = ["Manuel Gomez <mgrojo@gmail.com>"]
+maintainers-logins = ["mgrojo"]
+licenses = "Apache-2.0 OR GPL-2.0-or-later"
+website = "https://github.com/mgrojo/coap_spark"
+tags = ["spark", "coap", "iot", "protocol"]
+
+[[depends-on]]
+gnatprove = "^14.1.1"
+
+[build-switches]
+"*".Style_Checks = [
+    "-gnaty3" # Specify indentation level of 3
+    ,"-gnatya" # Check attribute casing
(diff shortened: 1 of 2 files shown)


index/co/coap_spark/coap_spark-0.9.0.toml
```diff
@@ -0,0 +1,69 @@
...
+
+authors = ["Manuel Gomez"]
+maintainers = ["Manuel Gomez <mgrojo@gmail.com>"]
+maintainers-logins = ["mgrojo"]
+licenses = "Apache-2.0 OR GPL-2.0-or-later"
+website = "https://github.com/mgrojo/coap_spark"
+tags = ["spark", "coap", "iot", "protocol"]
+
+[[depends-on]]
+gnatprove = "^14.1.1"
```
mosteo, line 26: We recommend having the dependency on the prover on a subcrate that runs the proof, so clients can avoid this heavy dependency.

mosteo (changes requested):

index/co/coap_spark/coap_spark-0.9.0.toml
mgrojo: `coap_client` is a subcrate and I actually run the proof there. But I'm also including that binary crate in the index. Would it be enough to move the dependency to the `coap_client` subcrate (`client` subdirectory) or would I need to create a new subcrate or reuse the `tests/` directory for that?

mosteo: Sorry, I'm not sure I follow. The key is: is people expected to use `coap_client` or is it just for proving/exemplifying? If someone can be genuinely interested in using `coap_client`, but needs not to perform the proof, then the dependency must be moved elsewhere (so not to burden them with the heavy dependency). Otherwise it is OK to leave it as-is.

Just to be clear, to me a subcrate in this context is a crate in the same repo as the crate being published, but not intended for general use. In other words, it is a "dev/private" crate used for testing, examples, proving... that may have different/more dependencies than the public-intended crate.

In that sense, `coap_client` is not a subcrate as it is a crate on its own right, appearing in the index, even if it depends on other crates.

As for where to place the proof, if you indeed want to remove it from `coap_client`, that's up to you :-) It could be part of the tests, or its own subcrate...

Let me know if something is still fuzzy...

mgrojo: Crystal clear.

`coap_client` is both an example of use of the library and a useful tool.

What if I remove the dependency here in the toml file for the index, but not in my repository? For future versions, I will think how to restructure the crates and the dependencies.

mgrojo: Argh, I forgot that SPARKlib is obtained by the gnatprove dependency. Consequently, clients require this dependency until SPARKlib is packaged separately as intended by AdaCore according to [this post](https://forum.ada-lang.io/t/where-is-sparklib/218/2?u=mgrojo).

mgrojo: I've reverted the commit removing the dependency to gnatprove. It's ready again for review.

mosteo: > What if I remove the dependency here in the toml file for the index, but not in my repository? For future versions, I will think how to restructure the crates and the dependencies.

That would work too.

mosteo (approved):

PR #1502, opened 2025-06-07
lsystem_editor 1.0.0

Heziode:

### DIFF
--- a/index/ls/lsystem_editor/lsystem_editor-1.0.0.toml
+++ b/index/ls/lsystem_editor/lsystem_editor-1.0.0.toml
@@ -1,0 +1,155 @@
+name = "lsystem_editor"
+description = "L-Systems editor in Ada"
+version = "1.0.0"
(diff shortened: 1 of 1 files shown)


Heziode:
The arch fails seems due to missing libraries in the base system.

mosteo:
I've added `glib2` to the Arch image, but I wonder if this should be an explicit dependency of `gtkada`...

Heziode:
Hum… right, since `glib2` is related to GTK ecosystem, adding it to `gtkada` as an explicit dependency seems relevant.

mosteo (approved):

Heziode:
@Fabien-Chouteau or @reznikmm can you add `glib2` to `gtkada` as an explicit dependency?

mosteo:
Well, it already depends on `gtk3` which at least on Debian depends on `glib2`. So this is somehow specific to Arch.

PR #1503, opened 2025-06-07
nrf52-runtimes 15.0.0

damaki:


damaki:
Looks like the build is failing on centos due to a missing glibc 2.35, required by gprbuild 25:
```
/root/.local/share/alire/toolchains/gprbuild_25.0.1_9a2e6cfb/bin/gprbuild: /lib64/libc.so.6: version `GLIBC_2.35' not found (required by /root/.local/share/alire/toolchains/gprbuild_25.0.1_9a2e6cfb/bin/gprbuild)
```

mosteo (approved):

PR #1506, opened 2025-06-18
aho_corasick 0.1.0

docandrew:

### DIFF
--- a/index/ah/aho_corasick/aho_corasick-0.1.0.toml
+++ b/index/ah/aho_corasick/aho_corasick-0.1.0.toml
@@ -1,0 +1,20 @@
+name = "aho_corasick"
+description = "Aho-Corasick Implementation"
+version = "0.1.0"
+
+authors = ["docandrew"]
+maintainers = ["docandrew <19672868+docandrew@users.noreply.github.com>"]
+maintainers-logins = ["docandrew"]
+licenses = "MIT OR Apache-2.0 WITH LLVM-exception"
+website = ""
+tags = []
+
+# project-files = ["aho_corasick.gpr", "sparklib.gpr"]
+
+[[depends-on]]
+gnatprove = "^15.1.0"
+
+[origin]
+commit = "54861150d7122f2ffc422b3acef51623ea6cfb81"
+url = "git+https://github.com/docandrew/aho_corasick.git"
+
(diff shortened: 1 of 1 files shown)


index/ah/aho_corasick/aho_corasick-0.1.0.toml
```diff
@@ -0,0 +1,20 @@
+name = "aho_corasick"
+description = "Aho-Corasick Implementation"
+version = "0.1.0"
+
+authors = ["docandrew"]
+maintainers = ["docandrew <19672868+docandrew@users.noreply.github.com>"]
+maintainers-logins = ["docandrew"]
+licenses = "MIT OR Apache-2.0 WITH LLVM-exception"
+website = ""
+tags = []
```
mosteo: A couple of tags would improve crate discoverability. You can add them editing this PR, no need to close and reopen.

index/ah/aho_corasick/aho_corasick-0.1.0.toml
```diff
@@ -0,0 +1,20 @@
...
+
+# project-files = ["aho_corasick.gpr", "sparklib.gpr"]
+
+[[depends-on]]
+gnatprove = "^15.1.0"
```
mosteo, line 15: If `gnatprove` is used just for proving, it should be moved to a nested crate to spare the heavy dependency to users of your library. (See https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing .)

However, if you're using formal containers provided by SPARK, this is OK as there is no current way around it that I'm aware of.

mosteo (changes requested):

index/ah/aho_corasick/aho_corasick-0.1.0.toml
docandrew: In this case I'm using a package from SPARKLib, I think gnatprove is the only way to get that today - I thought I heard that SPARKLib was being split into its own crate soon, I will update the dependencies list if/when that happens!

index/ah/aho_corasick/aho_corasick-0.1.0.toml
docandrew: Done!

index/ah/aho_corasick/aho_corasick-0.1.0.toml
```diff
@@ -0,0 +1,20 @@
+name = "aho_corasick"
+description = "Aho-Corasick Implementation"
+version = "0.1.0"
+
+authors = ["docandrew"]
+maintainers = ["docandrew <19672868+docandrew@users.noreply.github.com>"]
```
Fabien-Chouteau, line 6: Hi @docandrew , we need a valid email address here so we can contact you in case we move away from GitHub at some point.

docandrew: I did some major refactoring so will re-submit w/ the new version and updated email - thanks!

PR #1511, opened 2025-06-23
hostarm 25.0.4

jquorning:

### DIFF
--- a/index/ho/hostarm/hostarm-25.0.4.toml
+++ b/index/ho/hostarm/hostarm-25.0.4.toml
@@ -1,0 +1,50 @@
+name = "hostarm"
+description = "Local hosting of Ada Reference Manual (ARM)"
+version = "25.0.4"
+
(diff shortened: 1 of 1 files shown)


mosteo:
The error about multiple compilers seems spurious, I'd say this is the real failure:

https://github.com/alire-project/alire-index/actions/runs/15819697195/job/44585886226?pr=1511#step:6:1145

Not sure which difference between `ubuntu-lts` and `ubuntu-latest` may be causing that.

jquorning:
I suspect it is caused by GNAT 15 even though it is excluded in the manifest

https://github.com/alire-project/alire-index/actions/runs/15819697195/job/44585886226?pr=1511#step:6:1086

mosteo:
OK to merge then?

jquorning:
Ok with me.

mosteo (approved):

PR #1525, opened 2025-07-07
a0b_tools 0.1.0

godunko:

### DIFF
--- a/index/a0/a0b_tools/a0b_tools-0.1.0.toml
+++ b/index/a0/a0b_tools/a0b_tools-0.1.0.toml
@@ -1,0 +1,35 @@
+name = "a0b_tools"
+description = "A0B Tools: Runtime Generator"
+website = "https://github.com/godunko/a0b-tools"
+version = "0.1.0"
+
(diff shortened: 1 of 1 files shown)


index/a0/a0b_tools/a0b_tools-0.1.0.toml
```diff
@@ -0,0 +1,35 @@
...
+generate_ada = false
+generate_gpr = true
+generate_c = false
+
+[environment]
+PATH.append = "${CRATE_ROOT}/bin"
+
+[[depends-on]]
+a0b_base = "*"
+gnat = "*"
```
mosteo, line 27: Does this dependency accomplish something?

godunko: `a0b-tools` is expected to be `alr with` by application's crate. Application crate has dependency from some bare board GNAT. It results in two different compilers - native and cross - as dependency for application and Alire complains. Changing of `gnat_native` to `gnat` resolves this.

mosteo (approved):

PR #1536, opened 2025-07-13
adacl 6.2.1

krischik:

### DIFF
--- a/index/ad/adacl/adacl-6.2.1.toml
+++ b/index/ad/adacl/adacl-6.2.1.toml
@@ -1,0 +1,83 @@
+name                        = "adacl"
+description                 = "Ada Class Library (String, Trace, AUnit, Smart Pointer. GetOpt)"
+long-description            = """A class library for Ada for those who like OO programming.
+
+Currently the following functionality is migrated to Ada 2022:
+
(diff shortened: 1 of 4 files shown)


krischik:
I don't understand. I tested under Windows, macOS, OpenMandiva and no problems. Sadly both my Debian and OpenSuSE can't be upgraded any more and I don't really want to make a fresh installation any more and strange that AUnit has no timeouts.

krischik:
@mosteo Please merge it for now. I might have to disable the unit tests that tests the protected object in the next version. The unit test normally run through in less then 10 seconds and there is not reason to hang at any point.

mosteo:
Sorry I missed this the first time around: there are two seemingly unrelated crates here (`press_utils` and `sparkfun...`). I guess that's because those were merged against index `1.3.0` and thus are missing in `1.4.0`. You happy with merging those in this PR too?

krischik:
@mosteo Yes, it should be 1.4.0 - I updated to gnat "^15.1" yesterday. And adac_regex was in the wrong branch as well. So that is fixed as well. So it's all fine by me.

mosteo (approved):

mosteo:
OK thanks.

PR #1558, opened 2025-08-01
aho_corasick 0.3.0

docandrew:

### DIFF
--- a/index/ah/aho_corasick/aho_corasick-0.3.0.toml
+++ b/index/ah/aho_corasick/aho_corasick-0.3.0.toml
@@ -1,0 +1,20 @@
+name = "aho_corasick"
+description = "Aho-Corasick Implementation"
+version = "0.3.0"
+
+authors = ["docandrew"]
+maintainers = ["docandrew <jonfandrew@fastmail.com>"]
+maintainers-logins = ["docandrew"]
+licenses = "MIT OR Apache-2.0 WITH LLVM-exception"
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau (approved):

mosteo (approved):

index/ah/aho_corasick/aho_corasick-0.3.0.toml
```diff
@@ -0,0 +1,20 @@
...
+maintainers = ["docandrew <jonfandrew@fastmail.com>"]
+maintainers-logins = ["docandrew"]
+licenses = "MIT OR Apache-2.0 WITH LLVM-exception"
+website = ""
+tags = ["strings", "search", "matching", "intrusion", "aho-corasick"]
+
+# project-files = ["aho_corasick.gpr", "sparklib.gpr"]
+
+[[depends-on]]
+gnatprove = "^15.1.0"
```
Fabien-Chouteau, line 15: Hello @docandrew ,

We advise against making the `gnatprove` verification tool a library dependency because doing so means that any users of this library will have a dependency on this tool, too, even if they don't use it.

We recommend having a "test" sub-crate in the repo that depends on `gnatprove` and perform the verification from this sub-create.

Fabien-Chouteau (changes requested):

Fabien-Chouteau:
Oups ^^

PR #1568, opened 2025-08-09
adacl 6.2.2

krischik:


krischik:
I try again. This time without `gnat_native` as I don't think it's needed (any more).

PR #1569, opened 2025-08-09
adacl 6.2.2

krischik:

### DIFF
--- a/index/ad/adacl/adacl-6.2.2.toml
+++ b/index/ad/adacl/adacl-6.2.2.toml
@@ -1,0 +1,74 @@
+name                        = "adacl"
+description                 = "Ada Class Library: Strings, Tracing, AUnit, Smart Pointers, Getopt"
+long-description            = """A versatile Ada 2022 class library for robust application development.
+
+Features:
+- Getopt: Command-line argument parser with wide character support
+- Strings: Utilities for String, Wide_String, and Wide_Wide_String
+- Calendar: Time and date utilities with wide character support
+- Tracing: Diagnostic trace utility with wide character support
+- Protected Queue: Supports finish, wait, and abort operations
+- Smart Pointers:
+  - Reference-counted pointers
+  - Unique pointers (C++-style)
+  - Shared pointers (C++-style)
+- AUnit-Compatible Assertions:
+  - Generic for access, array, discrete, floating-point, fixed-point, decimal, and vector types
+- AUnit Parameter: Run tests with multiple inputs and expected values
+
+Related crates:
+- EAStrings: Encoding-aware string utilities
+- RegEx: SPITBOL-inspired pattern matching and regular expressions for String, Wide_String, and Wide_Wide_String
+- SAR: Text search and replace for String, Wide_String, and Wide_Wide_String
+
+Source: [SourceForge](https://sourceforge.net/p/adacl/git/ci/master/tree/adacl/src/)
+Documentation: [GNATdoc](https://adacl.sourceforge.net/gnatdoc/adacl/index.html)
+"""
+version                     = "6.2.2"
+licenses                    = "GPL-3.0-or-later"
+authors                     = ["Martin Krischik <krischik@users.sourceforge.net>"]
+maintainers                 = ["Martin Krischik <krischik@users.sourceforge.net>"]
+maintainers-logins          = ["krischik"]
+website                     = "https://sourceforge.net/projects/adacl/"
+tags                        = ["library", "command-line", "trace", "logging", "string", "aunit", "assert", "container", "smart-pointer", "ada2022"]
+
+[build-switches]
+"*".ada_version             = "Ada2022"
+"*".source_encoding         = "UTF_8"
+development.compile_checks  = "Warnings"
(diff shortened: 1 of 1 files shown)


krischik:
@mosteo All builds end at the test phase after determining the which compiler to use:

```
TESTING CRATE
alire-index-checks/scripts/gh-build-crate.sh: line 321: gprbuild: command not found
No system toolchain found, skipping system toolchain
TOOLCHAIN SOURCE: indexed
Note: Requested crate resolved as gnat_native=15.1.2
Note: gnat_native=15.1.2 set as default in global configuration.
Note: Requested crate resolved as gprbuild=25.0.1
Note: gprbuild=25.0.1 set as default in global configuration.
CRATE       VERSION STATUS  NOTES
gprbuild    25.0.1  Default
gnat_native 15.1.2  Default
Error: The operation was canceled.
```

Needless to say that the tests work just fine here. So I wonder what it could be.

krischik:
Let's see if this helps:

```toml
# [[actions]]
# type                        = "test"
# command                     = ["alr", "run"]
# directory                   = "test"
```

krischik:
@mosteo Removing the test action worked but also results in the unit tests not being executed. That is very disappointing.

The previous simple unit test framework worked just fine while new new one doesn't. Can you point me to a crate with working unit tests so I see how it's done correctly? The prove of concept should do.

Because I can't see any on the [AUnit crate overview](https://alire.ada.dev/crates/aunit):

image

Or alternatively show me how it's supposed to work.

mosteo:
> Needless to say that the tests work just fine here. So I wonder what it could be.

I'm seeing timeouts after 6h on all tests. I seem to remember we started seeing this already in some previous version.

The old way of testing should still work, the new testing framework isn't still enabled (except for the `dev` checks that use the master branch, that still should fall back to the old behavior). So we need to find a way to reproduce this locally so it can be properly fixed.

What is your local devel platform?

Orthogonally: I'm guessing you want this merged without waiting for a fix for the root cause of the problem, yes?

krischik:
> > Needless to say that the tests work just fine here. So I wonder what it could be.
 
> I'm seeing timeouts after 6h on all tests. I seem to remember we started seeing this already in some previous version.

Last time it was 3 or 5 hours and then timeout and it was only a few Linux agents who failed. This time it was all Agents on all operating systems that failed.

> The old way of testing should still work, the new testing framework isn't still enabled (except for the `dev` checks that use the master branch, that still should fall back to the old behavior). So we need to find a way to reproduce this locally so it can be properly fixed.

That is strange. I believe for a while all "test*" subdirectories where searched and that seems to be not the case any more because I see not AUnit output.

> What is your local devel platform?

macOS 15.6 (24G84). I have Windows 10 and OpenMadiva setup as well OS dependent bugs. I can test those platforms.

For a while I was using gnat_native but I was able to return to just gnat. I made sure the to specify the newest versions.

> Orthogonally: I'm guessing you want this merged without waiting for a fix for the root cause of the problem, yes?

Yes, please merge. There will be another release soon anyway and there are three crates in the pipeline who depend on the new version.

Of course I love to help you out getting this to work. If necessary I make an no change release just to test the unit tests.

Regards

Martin

krischik:
@mosteo I'm a step further with my analysis. `alr test` doesn't like when there is a test action in the alire.toml and will just hang. Calling `alr test` without the test action will work fine, run the tests and write the result into a log file. That's perfect so I removed all test actions — they where are workaround after all.

Now that I noticed that `alr test` writes the results into a log file I had another close look at the github output and knowing what to look for I saw that the test where indeed executed:

```log
[alr test] Test completed SUCCESSFULLY
--->8--- LOG FILE END: alire/alr_test_local.log
TEST RESULT with SYSTEM toolchain: failed=false
Test succeeded with system toolchain, skipping indexed toolchain
```

So that is perfect as well. It was just the old workaround hanging the system and after removing the workaround everything works perfect. 

We can do the pull request in full confidence that the unit tests where indeed executed.

mosteo:
I'm afraid that's not the case, @krischik . The log you're seeing is `alr test` using `alr build` as a fallback when no test is defined. Your test actions should work and I will look into why they're hanging everything up. Too bad that this doesn't happen locally. (I haven't had the opportunity to try and reproduce it myself yet.)

> I believe for a while all "test*" subdirectories where searched and that seems to be not the case any more because I see not AUnit output.

Yes, that was a workaround for a while because there was some limitation with `alr test` at the time, so there was no other way to have a test other than just building the crate. Once `alr test` was fixed, that is the intended way to test a crate. This means that your test actions are still needed, and we should find the root cause.

krischik:
@mosteo 

> Too bad that this doesn't happen locally.

It does happen locally. I saw it yesterday. I just haven't noticed before as I usually start tests with ´make test´ which interlay used `$(MAKE) -C "test" test`.

Yesterday I used `alr test` for the first time in a year and it worked. It also created test output in the log file mentioned. It also hangs when the when the test action is added to the toml file.

I can do more tests in the evening for you.

mosteo:
Ah, if you can reproduce locally so I should can. In any case that will simplify getting to the bottom of the issue.

PR #1570, opened 2025-08-10
abohlib 1.0.1

abitofhelp:

### DIFF
--- a/index/ab/abohlib/abohlib-1.0.1.toml
+++ b/index/ab/abohlib/abohlib-1.0.1.toml
@@ -1,0 +1,89 @@
+name = "abohlib"
+description = "A Bit of Help Library - Reusable Ada 2022 Components"
+version = "1.0.1"
+authors = ["A Bit of Help, Inc. - Michael Gardner"]
+maintainers = ["A Bit of Help, Inc. - Michael Gardner <mjgardner@abitofhelp.com>"]
+maintainers-logins = ["abitofhelp"]
+licenses = "MIT"
+website = "https://github.com/abitofhelp/abohlib_ada.git"
+tags = [
+    "ada",
+    "ada2022",
+    "gnat",
+    "alire",
+    "domain-driven",
+    "clean",
+    "hexagonal",
+    "repository",
+    "value-objects",
+    "result",
+    "error-handling",
+    "state-machine",
+    "type-safety",
+    "strong-typing",
+    "retry-pattern",
+    "resilience",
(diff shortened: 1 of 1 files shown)


index/ab/abohlib/abohlib-1.0.1.toml
```diff
@@ -0,0 +1,89 @@
...
+    "aggregate-root",
+    "phantom-types",
+    "functional",
+    "saga-pattern",
+    "event-sourcing",
+    "cross-platform",
+]
+
+[[depends-on]]
+trendy_test = "~0.0.3"
```
mosteo, line 39: This dependency seems suspicious. If it's only used to test your library, it's better to move it to a [nested crate](https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing).

Then, in your test action you can simply `alr -C tests run` rather than manually `gprbuild`ing a project file.

index/ab/abohlib/abohlib-1.0.1.toml
```diff
@@ -0,0 +1,89 @@
...
+    "-j0",
+]
+directory = "."
+
+[[actions]]
+type = "test"
+command = [
+    "./bin/test_all",
+    "--all",
+]
```
mosteo, line 64: As I mentioned above, if you move the testing to its own nested crate, depending on your main library, you can simplify this part to just a call to `alr run`.

index/ab/abohlib/abohlib-1.0.1.toml
```diff
@@ -0,0 +1,89 @@
...
+directory = "."
+
+[configuration]
+auto_gpr_with = true
+
+[build-profiles]
+"*" = "development"
```
mosteo, line 71: As clients of your library might prefer other build profiles, you should leave this out of the index manifest (unless there's an actual reason behind it).

mosteo (changes requested):
Thank you for your submission, Michael.

It seems some library is missing for linking in some platforms: https://github.com/alire-project/alire-index/actions/runs/16857914546/job/47826098839?pr=1570#step:7:829

The other comments are strong hints for improvements rather than actual errors, but those would make your crate more appealing to potential users.

abitofhelp:
I have added platform-specific code for Linux, macOS, and Windows. I have removed the trendy_test package, which was not being used. I am closing this pull request and will generate a new one for v1.1.0.

PR #1579, opened 2025-08-19
base32 1.0.0

JeremyGrosser:


JeremyGrosser:
The macos-latest failure looks like a macports issue?

krischik:
> The macos-latest failure looks like a macports issue?

macOS usually works just fine. Maybe a one off fluke when downloading the package.

BTW there is also a `long-description` which supports markdown including links which allows you to tell everybody why your crate is the best 😉. See https://alire.ada.dev/crates/adacl for an example how that can look.

JeremyGrosser:
Can someone retrigger the macos-latest check?

mosteo:
The macOS issue is caused by macos-latest now being macOS 15, which breaks the action that installs Ports. I'm downgrading the Ports checks to macos-14 for the time being in a separate PR.

mosteo (approved):

PR #1583, opened 2025-08-27
adacl_eastrings 6.3.0

krischik:

### DIFF
--- a/index/ad/adacl_eastrings/adacl_eastrings-6.3.0.toml
+++ b/index/ad/adacl_eastrings/adacl_eastrings-6.3.0.toml
@@ -1,6 +1,6 @@
 name                        = "adacl_eastrings"
 description                 = "AdaCL: Encoding-Aware String Utilities"
-long-description            = """Encoding-aware string utilities for Ada 2022, developed by BjÃ¶rn Persson for robust internationalization (i18n).
+long-description            = """Encoding-aware string utilities for Ada 2022, developed by Björn Persson for robust internationalization (i18n).
 
 Supports text processing with the following encodings:
 - Universal Character Set 4, Big-Endian (UCS-4 BE)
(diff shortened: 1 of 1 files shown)


index/ad/adacl_eastrings/adacl_eastrings-6.3.0.toml
```diff
@@ -0,0 +1,64 @@
+name                        = "adacl_eastrings"
+description                 = "AdaCL: Encoding-Aware String Utilities"
+long-description            = """Encoding-aware string utilities for Ada 2022, developed by BjÃ¶rn Persson for robust internationalization (i18n).
```
mosteo: I had to fix the encoding of Björn last time too. @krischik, if you could try to identify at what point this is happening? Are you using something not UTF-8 at some point?

(Ironic that I've only seen this in this i8n library ;-))

mosteo (approved):

index/ad/adacl_eastrings/adacl_eastrings-6.3.0.toml
krischik: > I had to fix the encoding of Björn last time too. @krischik, if you could try to identify at what point this is happening? Are you using something not UTF-8 at some point?

The original file has a vim block that should force the file into UTF-8 even if the automatic detection fails:

```toml
# vim: set textwidth=0 nowrap tabstop=8 shiftwidth=4 softtabstop=4 expandtab :
# vim: set filetype=toml fileencoding=utf-8 fileformat=unix foldmethod=diff :
# vim: set spell spelllang=en_gb :
```

The checked in version looks OK as well: https://sourceforge.net/p/adacl/git/ci/master/tree/adacl_eastrings/alire.toml

It must be somewhere in the pull request then.

mosteo: I see... most curious.

PR #1598, opened 2025-09-13
a0b_tools 0.2.0

godunko:


index/a0/a0b_tools/a0b_tools-0.2.0.toml
```diff
@@ -0,0 +1,35 @@
...
+generate_gpr = true
+generate_c = false
+
+[environment]
+PATH.append = "${CRATE_ROOT}/bin"
+
+[[depends-on]]
+a0b_base = "*"
+gnat = "*"
+gnatcoll = "*"
```
Fabien-Chouteau, line 28: Blind dependency on any gnatcoll and vss dependency is bad for your users. 
The build will break sooner or later.

mosteo:
@godunko did you see the comment from Fabien? Do you plan to update the dependencies?

PR #1600, opened 2025-09-14
adacl 7.0.0

krischik:


krischik:
@mosteo You where right. The unit test are not executed. They would have failed because of faulty pins. I'll fix that in the next release.

fyi @Fabien-Chouteau

Fabien-Chouteau (approved):

PR #1611, opened 2025-10-03
scatcdsk_lib 1.0.0

poldy:


poldy:
Have to disable macOS 26 toolchain workaround

PR #1612, opened 2025-10-03
scatcdsk_lib 1.0.0

poldy:

### DIFF
--- a/index/sc/scatcdsk_lib/scatcdsk_lib-1.0.0.toml
+++ b/index/sc/scatcdsk_lib/scatcdsk_lib-1.0.0.toml
@@ -1,0 +1,23 @@
+name = "scatcdsk_lib"
+description = "Reusable software components extracted from the SCATC DSK"
+version = "1.0.0"
+
+authors = ["Richard Conn"]
+maintainers = ["Darren Bane <darren.bane.ac@posteo.net>"]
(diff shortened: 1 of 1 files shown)


poldy:
Windows builds fail because I run the tests from a UNIX shell script. I don't use Windows personally, but if this is a blocker please advise.

poldy:
> Windows builds fail because I run the tests from a UNIX shell script. I don't use Windows personally, but if this is a blocker please advise.

... and CentOS doesn't even have the POSIX-standard diff command???

mosteo:
> ... and CentOS doesn't even have the POSIX-standard diff command???

This we can add to the CentOS docker image.

mosteo:
> Windows builds fail because I run the tests from a UNIX shell script. I don't use Windows personally, but if this is a blocker please advise.

Run the tests only on Linux with `[[actions.'case(os)'.linux]]`. You can push that change directly to this PR (or even edit through the GH WebUI).

mosteo:
I did it for you and updated CentOS image, so it should pass all checks now.

poldy:
Thanks, I was at $DAYJOB but those changes look fine to me. I didn't know about actions.'case(os)' , it's just what I needed.

mosteo (approved):

PR #1617, opened 2025-10-06
VSS: fix packages hashes

AldanTanneo:
The recent renaming of the VSS git repo to VSS-archived (which was split into VSS-text and VSS-extra for future development) broke the VSS package on Alire.


eliericha:
Hello,

Before proceeding with an update of VSS hashes, AdaCore is working on reverting back to the original hashes.

The change of hashes was likely caused by the renaming of the repository. So we're renaming it back to see if we can restore the previous situation.

So please hold this PR until we see if simply reverting the name will resolve the situation.

Thanks!

PR #1626, opened 2025-10-12
hp41cx_tools 1.8.1

krischik:


krischik:
@mosteo The crate has 571 unit tests but it seems they are not actually executed. Maybe if you have some time you can look at it.

mosteo:
Yes, you need to explicitly invoke them with a test action, but then I guess we'll hit that bizarre problem in which some task dies and the tests never end. I'll try to reproduce locally.

mosteo (approved):

PR #1630, opened 2025-10-17
swan_ada 0.1.0

xadaemon:

### DIFF
--- a/index/sw/swan_ada/swan_ada-0.1.0.toml
+++ b/index/sw/swan_ada/swan_ada-0.1.0.toml
@@ -1,0 +1,15 @@
+name = "swan_ada"
+description = "Stuff We All Need - utilities collection for ada"
+version = "0.1.0"
+
+authors = ["Matheus Xavier"]
+maintainers = ["Matheus Xavier <mxavierdev@gmail.com>"]
+maintainers-logins = ["xadaemon"]
+licenses = "Apache-2.0"
+website = "https://github.com/xadaemon/swan_ada"
+tags = []
+
+[origin]
+commit = "3bba2f4c6154ef939a610cd61230fe5edd5484d8"
+url = "git+https://github.com/xadaemon/swan_ada.git"
+


index/sw/swan_ada/swan_ada-0.1.0.toml
```diff
@@ -0,0 +1,15 @@
+name = "swan_ada"
```
mosteo, line 1: @xadaemon, thank you for your submission.

Since this is the first version of this crate, I think it would be better to just name the crate `swan`. In general we frown upon on using "ada" when in the Ada ecosystem as it is just noise. For some projects it's unavoidable, but in this case it seems better all around, also because the packages will start with `Swan.` rather than the `Swan_Ada.` mouthful.

I know renaming all your packages will be a chore, and this isn't a mandatory change, but I greatly encourage you to make this change.

Of course, the repo can stay as `swan_ada`.

index/sw/swan_ada/swan_ada-0.1.0.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "Stuff We All Need - utilities collection for ada"
+version = "0.1.0"
+
+authors = ["Matheus Xavier"]
+maintainers = ["Matheus Xavier <mxavierdev@gmail.com>"]
+maintainers-logins = ["xadaemon"]
+licenses = "Apache-2.0"
+website = "https://github.com/xadaemon/swan_ada"
+tags = []
```
mosteo, line 10: A couple of tags at least are highly recommended to increase the visibility of your crate. You can take a look at https://alire.ada.dev/tags/ for inspiration.

mosteo (changes requested):

xadaemon:
Hey, I will apply your suggestions later today, thanks for the feedback 😀

pt., 17 paź 2025, 12:56 użytkownik Alejandro R Mosteo <
***@***.***> napisał:

xadaemon:
I will close this pr as the new lib name generated another PR #1631

PR #1633, opened 2025-10-18
x1804 1.0.2

poldy:


poldy:
The Windows failures look to be a problem running curl, nothing to do with my changes? And I'm not sure why that stage even runs on Windows, I thought I had excluded it.

mosteo:
It seems to be a transient error in the server that `curl` is trying to contact. As for the Windows failures, the crate retrieval is run even if tests are not enabled on that platform.

mosteo (approved):

PR #1643, opened 2025-10-22
hp41cx_tools 1.8.4

krischik:


krischik:
Ups. That's no good. That's a bug to be fixed.

PR #1658, opened 2025-11-02
AWS 25.0 crate (proposal)

stcarrez:
This is a proposal for another way to integrate the Ada Web Server as a crate. The issue with the current approach is the Makefile-based environment used by AWS and the fact that we have to completely build and then install AWS to use it.

In this proposal, only the 'make setup' phase is done to let the AWS Makefile do the important setup and prepare the required files in target directory. That target directory is named by AWS by using `gcc -dumpmachine` which brings other issues and to solve and fix things, that directory is moved to a fixed directory `build` so that every target will use that fix value.

The installation and build is not made and instead we reference the two important GNAT projects 'aws.gpr' (top-level) and `build/projects/aws_config.gpr` which was generated by `make setup`.  The build is then handled by Alire since it knowns the AWS projects.  The AWS GNAT projects rely on the following important variables which are set through the makefiles and we can set them for each target as follows:

TGT_DIR = "../build"
TARGET = "build"

It is a proposal in the hope it will solve the AWS crate integration issues....

### DIFF
--- a/index/aw/aws/aws-25.0.0.toml
+++ b/index/aw/aws/aws-25.0.0.toml
@@ -1,0 +1,60 @@
+description = "Ada Web Server"
+website = "https://github.com/adacore/aws"
+name = "aws"
+version = "25.0.0"
+authors = ["AdaCore"]
+licenses = "GPL-3.0-only WITH GCC-exception-3.1"
+maintainers = ["chouteau@adacore.com"]
+maintainers-logins = ["Fabien-Chouteau"]
+project-files = ["aws.gpr", "build/projects/aws_config.gpr"]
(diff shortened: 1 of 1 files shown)


stcarrez:
@Fabien-Chouteau, good news, the build passes with my AWS 25.0 proposal, only the author's check failed which is expected.

Fabien-Chouteau:
Thanks @stcarrez , that works for me. I guess you tested this with your crates that depend on AWS.

stcarrez:
Yes, I was able to use it and run my unit tests on utilada_aws.

The crate is based on the AWS tar 25.0 that you created for a first submission.
If you publish a AWS tar with templates for the 25.1.0, we could try if it the latest AWS release can be used with this method.

Fabien-Chouteau (approved):

PR #1659, opened 2025-11-07
Create aws-25.2.0.toml

Fabien-Chouteau:


Fabien-Chouteau:
cc @stcarrez

mosteo (approved):

mosteo:
Waiting for any input from @stcarrez but OK to me.

stcarrez:
I've been able to build several of my projects by using the aws-25.2.0 crate that you proposed. I'm building on Linux and I also tried this afternoon on MacOS. Both are working well with the crate! Please, go ahead!

In parallel, I'm reaching several issues with alire dependencies that I am solving by using definitions like:

```
aws = "^24.0|^25.0"
xmlada = "^24.0|^25.0"
```

I'm not convinced that incrementing the major number for xmlada helps for projects to use that library. The API itself does not change much over time and it prevents to use newer versions in some cases.

The upgrade AWS 24 -> AWS 25 is difficult from the Alire point of view (not from the Ada compilation side) because of dependencies in xmlada ~24 or ~25 which are somewhere in the tree (AWS 24 -> xmlada 24, AWS 24 -> gnatcoll -> libgpr -> xmlada 24).

mosteo:
Thanks Stephane. Yes, that kind of "anchoring" to deep dependencies is a bother. It also happens with crates that depend on a particular GNAT version.

PR #1672, opened 2025-11-17
tzif 1.0.0-rc1

abitofhelp:

### DIFF
--- a/index/tz/tzif/tzif-1.0.0-rc1.toml
+++ b/index/tz/tzif/tzif-1.0.0-rc1.toml
@@ -1,0 +1,46 @@
+name = "tzif"
+description = "Parsing and querying of IANA's compiled timezone information"
+version = "1.0.0-rc1"
+authors = ["Michael Gardner"]
+maintainers = ["Michael Gardner <mike@abitofhelp.com>"]
+maintainers-logins = ["abitofhelp"]
+licenses = "BSD-3-Clause"
+website = "https://github.com/abitofhelp/tzif_ada.git"
+project-files = ["tzif.gpr"]
+tags = [
+    "timezone",
+    "zoneinfo",
+    "tzif",
+    "tzdata",
+    "iana",
+    "time",
+    "rfc9636",
+    "parser",
+    "functional",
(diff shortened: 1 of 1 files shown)


index/tz/tzif/tzif-1.0.0-rc1.toml
```diff
@@ -0,0 +1,46 @@
+name = "tzif"
+description = "Parsing and querying of IANA's compiled timezone information"
+version = "1.0.0-rc1"
```
mosteo, line 3: Is there a particularly strong reason to publish a release candidate? We usually don't want such short-lived versions in the index.

abitofhelp: Hi @mosteo, 

I have three closely related crates that are complete, other than polishing the documentation: Functional, used by TZif, and used by ZoneInfo.  I published release candidates for TZif and ZoneInfo to signal that the full release, v1.0.0, of these crates is imminent and to "hold" the names I used for these crates.  I know there is no concept of reserving or holding, officially, but it would be terrible after spending so much time creating something that is needed, and then having to rebrand everything... package names, file names, etc.  I promise to publish v1.0.0 of TZif and ZoneInfo within the next week.  Would this be acceptable?  Hot on their heels will be an Ada linter/format/fixer tool that supports Ada 2022.  It uses libadalang and really works! :)

mosteo: Well, given this information I think there's no risk of someone cybersquatting `tzif`/`zoneinfo` from you, as we would have to review it. I prefer not having those RCs "forever" in the index if it's a matter of a few days :)

abitofhelp:
Thank you!

Mike
_____________________________
Mike Gardner
Principal Software Engineer
A Bit of Help, Inc.
+1 602.810.8161
abitofhelp.com
LinkedIn https://www.linkedin.com/in/michaeljosephgardner/
GitHub https://github.com/abitofhelp

abitofhelp:
Superceded by v2.0.0

PR #1678, opened 2025-11-21
booch_light 1.0.0-alpha

kevlar700:

### DIFF
--- a/index/bo/booch_light/booch_light-1.0.0-alpha.toml
+++ b/index/bo/booch_light/booch_light-1.0.0-alpha.toml
@@ -1,0 +1,75 @@
+name = "booch_light"
+description = "Booch components for constrained systems"
+version = "1.0.0-alpha"
+
+authors = ["Kevin Chadwick"]
+maintainers = ["Kevin Chadwick <kc-ada@elansys.co>"]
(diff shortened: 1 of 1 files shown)


index/bo/booch_light/booch_light-1.0.0-alpha.toml
```diff
@@ -0,0 +1,75 @@
+name = "booch_light"
+description = "Booch components for constrained systems"
+version = "1.0.0-alpha"
```
mosteo, line 3: Hi Kevin, the community index is generally intended for "stable" releases given that it reaches all the users. Is it really necessary to publish this alpha here? (This is all relative as 0.x releases for example are accepted unquestioned. But still, I hope you see what I mean.)

I'd suggest either wait for the actual 1.0.0 or publish early releases as 0.x versions. Or, if you have some users collaborating on this, may be have your own index for pre-release testing.

kevlar700: Fair enough. It was mostly just because it's a quite small modification of Booch's released code and I did find one issue recently that didn't meet my usual standards so it's a bit of a confused release number. Probably best to keep the API open. I'll change it. Thank you.

PR #1684, opened 2025-11-25
Add coap_spark and coap_client version 0.10.0

mgrojo:


mgrojo:
In my action, the crate builds for windows-latest, and I don't understand the difference.
https://github.com/mgrojo/coap_spark/actions/runs/19686120043

Regarding macos, it is failing in wolfssl, so why is it different from the PR #1682?

mosteo:
Some libraries are undertested in our current setup as they're not linked against. But here the error is during compilation, so it really should have happened at the time.

I can't either see clear reasons for these failures. It may be some tricky side-effect of running under `alr test` instead of directly (as I imagine you're testing). Given that your tests pass, I'm inclined to merge. OK with you?

mgrojo:
> I can't either see clear reasons for these failures. It may be some tricky side-effect of running under alr test instead of directly (as I imagine you're testing). Given that your tests pass, I'm inclined to merge. OK with you?

OK, I agree.

mosteo (approved):

PR #1694, opened 2025-12-02
functional 2.2.1

abitofhelp:


abitofhelp:
I had to amend this crate because it uses submodules and the consumers of this library were using the same submodules; colliding.  After this version is published, I will be able to publish 3 crates: hybrid_app_ada, hybrid_lib_ada, tzif.

mosteo (approved):

PR #1702, opened 2025-12-09
transition retired macos-13 images to macos-15-intel

AldanTanneo:
The `macos-15-intel` image is the last x86_64 based image for macos on GitHub.

Also restore the macports workflow to `macos-latest` (the issue appears to be fixed in the actions' repository)


mosteo (approved):

mosteo:
Indeed both work OK, thanks.

PR #1706, opened 2025-12-11
xmlada 26.0.0

AldanTanneo:


AldanTanneo:
Can someone explain why github is sending the wrong hash every other request in CI?? it's very weird, is the download silently failing?

mosteo (approved):

PR #1723, opened 2025-12-18
gnatcoll_minimal 26.0.0

AldanTanneo:
The main GNATCOLL crate got split in 3 crates:

- gnatcoll_minimal, this one, that doesn't depend on anything and has some core functionality
- gnatcoll (actually gnatcoll_core, but Fabien prefers to keep using the previous name), having everything in the old gnatcoll except the libgpr related functionality
- gnatcoll_projects, with the libgpr stuff.

Hopefully we can use gnatcoll in Alire when this is merged, and not the old fork (gnatcoll-slim), since the libgpr dependency is dropped :)

### DIFF
--- a/index/gn/gnatcoll_minimal/gnatcoll_minimal-26.0.0.toml
+++ b/index/gn/gnatcoll_minimal/gnatcoll_minimal-26.0.0.toml
@@ -1,0 +1,35 @@
+name = "gnatcoll_minimal"
+version = "26.0.0"
+description = "GNAT Components Collection - Minimal"
+website = "https://github.com/adacore/gnatcoll-core"
+authors = ["AdaCore"]
+licenses = "GPL-3.0-or-later WITH GCC-exception-3.1"
+maintainers = ["chouteau@adacore.com", "sagaert@adacore.com"]
+maintainers-logins = ["Fabien-Chouteau", "AldanTanneo"]
+project-files = ["minimal/gnatcoll_minimal.gpr"]
+tags = ["gnatcoll", "asserts", "encoders", "stream", "storage"]
+
+[configuration]
+disabled = true
+
+[gpr-externals]
+GNATCOLL_BUILD_MODE = ["DEBUG", "PROD"]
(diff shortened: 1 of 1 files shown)


index/gn/gnatcoll_minimal/gnatcoll_minimal-26.0.0.toml
```diff
@@ -0,0 +1,35 @@
...
+project-files = ["minimal/gnatcoll_minimal.gpr"]
+tags = ["gnatcoll", "asserts", "encoders", "stream", "storage"]
+
+[configuration]
+disabled = true
+
+[gpr-externals]
+GNATCOLL_BUILD_MODE = ["DEBUG", "PROD"]
+GNATCOLL_OS = ["windows", "unix", "osx"]
+LIBRARY_TYPE = ["static", "relocatable", "static-pic"]
```
mosteo, line 18: I gather these externals will be shared across gnatcoll crates? Or should that be `GNATCOLL_MINIMAL`? Or will the one in `gnatcoll` be `GNATCOLL_CORE`?

AldanTanneo: these are shared, and across the gnatcoll bindings too; should they only be specified in the root crate or in all crates? the names do not change from crate to crate

mosteo: As long as they match I don't think it is a problem, but since they can only create friction if misused, I'd advise to define them only in the root crate. Maybe leave a comment for documentation.

AldanTanneo: as this is the root crate shared between all gnatcoll related crates, it's okay here :)

mosteo: Ah ok thanks, I was on autopilot.

mosteo (approved):

PR #1737, opened 2026-01-05
fix: limit GNAT version for gnatdoc 25

AldanTanneo:


AldanTanneo:
Hm, Debian repo GNAT doesn't like :thinking:
And of course GNAT 14 on macos \*sigh\*

mosteo (approved):

mosteo:
Since the metadata change is pretty straightforward, I would simply merge, unless you want to do something about macOS. The debian issue is with our checks that have some shortcomings when it comes to forced compilers.

Fabien-Chouteau (approved):

PR #1746, opened 2026-01-06
libadalang 26.0.0

AldanTanneo:


AldanTanneo:
I don't understand why the windows CI cannot meet the `make` dependency...

mosteo:
> I don't understand why the windows CI cannot meet the `make` dependency...

Missing skipped dependencies point to a solver timeout in non-interactive mode, but to be sure I'll have to check solver defaults and whether a warning should be expected in that case...

mosteo:
It's indeed a solver timeout, because the default non-interactive is quite short. I'll look into fixing it in the checks for starters.

mosteo:
It's fixed now.

Fabien-Chouteau (approved):

PR #1747, opened 2026-01-06
various build fixes for AdaCore libraries

AldanTanneo:


index/la/langkit_support/langkit_support-26.0.0.toml
```diff
@@ -9,6 +9,13 @@ licenses = "Apache-2.0 WITH LLVM-exception"
 website = "https://github.com/AdaCore/langkit"
 project-files = ["langkit/support/langkit_support.gpr"]
 
+[gpr-externals]
+LANGKIT_SUPPORT_BUILD_MODE = ["dev", "prod", "prof"]
+LANGKIT_SUPPORT_LIBRARY_TYPE = ["static", "relocatable", "static-pic"]
+
+[gpr-set-externals]
+LANGKIT_SUPPORT_BUILD_MODE = "prod"
```
mosteo, line 17: For this on an other manifests in the PR: normally we say to leave externals unset and use a default for the preferred value, as otherwise the client can't override it. I'm guessing these releases won't work without this and you're aware of this recommendation.

AldanTanneo: Well it's not that they won't work, it's that the default is the dev profile :( which is annoying when building, say, gnatformat from source

Fabien-Chouteau (approved):

PR #1769, opened 2026-01-13
hp41cx_tools 1.8.9

krischik:


krischik:
It looks like `Long_Integer` on macOS and Windows are of different size which makes Windows fail.

PR #1770, opened 2026-01-13
Release/adabots 2.1.0

TamaMcGlinn:
Created via alr publish with alr 2.0.2+9b80158.


index/tb/tb6612_driver/tb6612_driver-0.1.0.toml
```diff
@@ -0,0 +1,18 @@
+name = "tb6612_driver"
+description = "Driver for the TB6612 dual DC motor driver IC."
+version = "0.1.0"
```
mosteo, line 3: I'm guessing this is one we failed to forward-port from `1.3.0` to `1.4.0`, so it's OK.

mosteo (approved):

PR #1771, opened 2026-01-13
hp41cx_tools 1.8.10

krischik:


krischik:
**ubuntu-lts** checked out the wrong package. That sometime happens.

mosteo (approved):

PR #1772, opened 2026-01-13
remove gnatcoll_core build action

AldanTanneo:
The action causes too much pain for too little benefit. Only now the optimization variables (like `GNATCOLL_XXHASH_ARCH` and the like) will not be properly set on the targets they're supported on, and require manual fiddling.

Sadly the conditionals in the manifest are parametrized on the host, not the target... Should we just pretend they do, or leave the unoptimized default?

### DIFF
--- a/index/gn/gnatcoll/gnatcoll-26.0.0.toml
+++ b/index/gn/gnatcoll/gnatcoll-26.0.0.toml
@@ -26,13 +26,6 @@ GNATCOLL_BLAKE3_ARCH = ["x86_64-linux", "aarch64-linux", "x86_64-windows", "gene
 gnat = ">=14"
 gnatcoll_minimal = "~26.0.0"
 
-[[actions]]
-type = "pre-build"
-# This action defines the default externals for your current target. Setting them in
-# alire.toml will override them. Requires python and gprconfig.
-command = ["python3", "gnatcoll_core.gpr.py", "build", "--configure-only", "--enable-constant-updates"]
-directory = "core"
-
 [origin]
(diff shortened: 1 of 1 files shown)


mosteo (approved):

mosteo:
At some point we may have to make Alire x-target aware. Even so, at that time I guess we would introduce new variables so both the host and the target info can be used, and the existing ones will always refer to the host.

AldanTanneo:
> At some point we may have to make Alire x-target aware. Even so, at that time I guess we would introduce new variables so both the host and the target info can be used, and the existing ones will always refer to the host.

We could use `gprconfig` and parse its output (there's a flag for stable-ish output for programmatic access) to determine target properties, but the issue here is that would prevent us from having target specific dependencies, as `gnat` should be resolved first before we can use `gprconfig`.

We would need to split `gnat` dependency resolution and regular dependency resolution, at which point does it even make sense to have `gnat` in the same system as regular dependencies?

mosteo:
> We would need to split gnat dependency resolution and regular dependency resolution, at which point does it even make sense to have gnat in the same system as regular dependencies?

This would align with the need for a "tools-only" dependency set as discussed recently on Matrix, needed only in the host for building, separate from the actual code dependencies. This happens in other frameworks too (I'm familiar with ROS for example), so it may be an undertaking we need to do at the same time. (Or possibly precede the improvements for x-compilation).

We should carefully consider how all of this interacts with the build stages, too, or if a simpler solution may exist (perhaps leveraging nested crates).

PR #1774, opened 2026-01-16
gnattest 26.0.0

AldanTanneo:
lots of `sed` invocations there... @mosteo as part of potential predefined actions, do you think having a `patch` action could be reasonable?


mosteo:
> lots of `sed` invocations there... @mosteo as part of potential predefined actions, do you think having a `patch` action could be reasonable?

Makes sense of first thought. I haven't given it much though besides obvious copy, move, touch. I still want to research similar capabilities in other PMs.

mosteo (approved):

PR #1781, opened 2026-01-20
gnatformat relocatable fix

AldanTanneo:
needs #1780, makes `LIBRARY_TYPE=relocatable alr install gnatformat` work (on linux at least)


AldanTanneo:
Why do the `(alr dev)` CIs show the build log as it proceeds, but the `(Alire toolchain)` (with stable alr) don't?

mosteo:
> Why do the `(alr dev)` CIs show the build log as it proceeds, but the `(Alire toolchain)` (with stable alr) don't?

Doesn't immediately ring a bell :/

Fabien-Chouteau (approved):

PR #1787, opened 2026-01-22
change adasat default build mode

AldanTanneo:


Fabien-Chouteau (approved):
I am not a huge fan of patching sources with sed, but OK.

Another feature to add to the manifest I guess:
```toml
patch = """
--- file1.txt
+++ file1.txt
@@ -1 +1 @@
-This is the original text.
+This is the modified text.
"""
```

PR #1788, opened 2026-01-25
tamga 1.0.0

ApaydinBahadir:


AldanTanneo:
The target branch should be `stable-1.4.0`, not `stable-1.3.0`

The `alr` you used is out of date and still targets the old branch.

PR #1789, opened 2026-01-25
Add adamodbus 1.0.1

flottokarotto:
## New crate: adamodbus

**adamodbus** - SPARK-verified Ada 2022 Modbus library with SunSpec support and C API

### Features
- Modbus RTU, ASCII, and TCP protocol support
- Master (Client) and Slave (Server) implementations
- 100% SPARK-verified protocol core
- SunSpec Alliance energy profiles
- C API for interoperability
- go-e Charger support

### Links
- Repository: https://github.com/flottokarotto/AdaModbus
- License: MIT

### Changes from v1.0.0 submission (addressing review feedback)
- Removed `gnatprove` from dependencies (install via `alr with gnatprove` in CI)
- Fixed `maintainers-logins` to correct GitHub username
- Removed `gpr-set-externals` to allow build mode override
- Kept `aunit` as dependency (small, needed for tests)


index/ad/adamodbus/adamodbus-1.0.0.toml
```diff
@@ -0,0 +1,35 @@
...
+ADAMODBUS_BUILD_MODE = "debug"
+
+[[depends-on]]
+gnat = ">=12"
+
+[[depends-on]]
+aunit = "^26.0.0"
+
+[[depends-on]]
+gnatprove = "^15.1.0"
```
AldanTanneo: We usually advise to not put `gnatprove` as a public dependency, as anyone needing to build your library will need to pull a 200MB+ package. If you depend on the spark library, the `SPARKlib` crate is available.

flottokarotto: I'm using gnatprove in the ci pipeline, is there a best practice approach?

AldanTanneo: Maybe executing `alr with gnatprove=15.1.0` in your CI?

index/ad/adamodbus/adamodbus-1.0.0.toml
```diff
@@ -0,0 +1,35 @@
...
+
+tags = ["modbus", "embedded", "protocol", "serial", "tcp", "spark", "industrial", "sunspec", "energy", "solar"]
+
+# For TLS support, use the separate adamodbus_tls crate
+
+[gpr-externals]
+ADAMODBUS_BUILD_MODE = ["debug", "release"]
+
+[gpr-set-externals]
+ADAMODBUS_BUILD_MODE = "debug"
```
AldanTanneo: Also: externals set with `[gpr-set-externals]` are not overridable, so currently your crate will always build in debug mode. Alire generates config files with a build profile, you may want to use those.

AldanTanneo (approved):

index/ad/adamodbus/adamodbus-1.0.0.toml
mosteo: Alternatively, having a nested crate that depends on your main library and on gnatprove and carries out testing/validation. See https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing

flottokarotto: thanks for the advice, i did something similar with the TLS project, because of the aws dependency.

PR #1800, opened 2026-02-09
st7789 0.1.0-dev

markhermeling:

### DIFF
--- a/index/st/st7789/st7789-0.1.0-dev.toml
+++ b/index/st/st7789/st7789-0.1.0-dev.toml
@@ -1,0 +1,18 @@
+name = "st7789"
+description = "ST7789 display driver for RP2040 8-bit parallel interface via PIO + DMA"
+version = "0.1.0-dev"
+
+authors = ["Mark Hermeling"]
+maintainers = ["Mark Hermeling <mark@hermeling.org>"]
+maintainers-logins = ["markhermeling"]
+licenses = "Apache-2.0"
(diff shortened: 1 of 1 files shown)


index/st/st7789/st7789-0.1.0-dev.toml
```diff
@@ -0,0 +1,18 @@
+name = "st7789"
```
Fabien-Chouteau, line 1: Since the driver is only supported on the rp2040 (and maybe rp235X in the future), I recommend using a name that hints at that (e.g. `st7789_rp`)

index/st/st7789/st7789-0.1.0-dev.toml
```diff
@@ -0,0 +1,18 @@
...
+description = "ST7789 display driver for RP2040 8-bit parallel interface via PIO + DMA"
+version = "0.1.0-dev"
```
Fabien-Chouteau, line 3: the `-dev` suffix is added by `alr init` to indicate a development version. For a release you want to remove the suffix and just keep `0.1.0`.

And then in your working branch switch to `0.2.0-dev` for instance.

Fabien-Chouteau (changes requested):

index/st/st7789/st7789-0.1.0-dev.toml
markhermeling: Thanks! Will work on this in a couple of weeks.

PR #1803, opened 2026-02-13
Fix Zanyblue 1.4.0 crate name in nested crates.

Blady-Com:


mosteo:
Hi @Blady-Com, changing commit information goes against best practices of release versioning. I'd suggest to have a new release with the fixes you need.

Blady-Com:
Hi @mosteo, I agree in most cases.
I was messed in mistakes I did with nested crate names mixed with issues of bin folder location, see [Alire issue](https://github.com/alire-project/alire/issues/2071).
In fact, I don't manage the release version, it is done by its author.
It would be nice if this release could be publish with this version number in respect to its author.
Thanks, Pascal.

mosteo:
I see, I didn't realize you're not the author in this case.

Frankly, changing a release commit after publishing is a big no-no in my eyes. Can't you communicate with the author to have a patch version out? Or, if he doesn't want to release, if he can guarantee that the next upstream version is going to be 1.5 (or 1.4.2), we could use 1.4.1 here without fear of future clash.

Worst case that he doesn't agree to anything, I would go for 1.4.1-1 (as a pre-release).

Blady-Com:
Unfortunately, the author haven't post anything new for 6 years. He considered Alire but without publishing anything. So I brought the code to Github so that it could be available with Alire.
I don't know what he planned for the future.
Would be something like _1.4.0-official_ acceptable?

AldanTanneo:
> Would be something like _1.4.0-official_ acceptable?

The way semver works, `1.4.0-something` is considered a prerelease version of `1.4.0`, so the latest would still be `1.4.0`

mosteo:
As César said, you need to go to at least `1.4.1-whatever` to supersede the bad 1.4.0. But take into account the ordering rules of /whatever/ in case you would eventually want to publish another pre-release:

https://semver.org/#spec-item-11

Blady-Com:
I hadn't any answer of the author.
How do I delete a create of the index?

AldanTanneo:
What do you mean delete a crate?

mosteo:
> How do I delete a create of the index?

There's no way to remove a crate (or release). When this has been requested in the past, the best you can do is set `available = false` in the release you want to make unavailable.

Blady-Com:
How to do that on an existing published release?

mosteo:
> How to do that on an existing published release?

You submit a PR to the index, just like this one, modifying the manifest. It has to be done manually, alas. Or possibly it can be done through the GitHub web interface, given that it's a small edition.

mosteo (approved):

PR #1818, opened 2026-03-04
hostarm 26.5.1

jquorning:
Created via `alr publish` with `alr 2.1.1-dev`


mosteo:
I've seen similar problems on macOS but I don't know a solution :/

As for the Debian/Ubuntu failure:

> gpr-util-put_resource_usage__unix.adb:34:24: error: "Timeval" not declared in "Thin_Common"

Also not immediately aware of where this is coming from.

jquorning:
It looks like AWS is compiled with GNAT 15 even though `gnat = "<15"` is explicitly stated.

PR #1821, opened 2026-03-11
Add TP7 Gnoga.

Blady-Com:


mosteo:
> tp7.ads:79:46: error: value not in range of type "Standard.Long_Long_Float"

Looks like some hard-coded constants aren't sitting well with some architectures.

Blady-Com:
Oh really? But they're 64-bit architectures!
Anyway, how do I specifically disable only those architectures?
I couldn't find in the documentation ubuntu-24.04-arm or macos-latest items.

mosteo:
Yes, it's puzzling, but it's not the first time I've seen that. I don't really know the ultimate reason, maybe some of our experts on embedded do. You might try asking in the ada-lang.io forum.

To disable, you can use the (involuntarily undocumented) `host-arch` variable. See the possible values here:

https://github.com/alire-project/alire/blob/2681b313d08224014fa3f5e4f76e9c75347213d2/src/alire/alire-platforms.ads#L11

You would use that with the `available` property (https://github.com/alire-project/alire/blob/v2.1.0/doc/catalog-format-spec.md#release-information)

Fabien-Chouteau (approved):

PR #1826, opened 2026-03-14
ZanyBlue release 1.5.0.

Blady-Com:


Blady-Com:
The error is `zbinfo-main.adb:1:26: (style) incorrect line terminator [-gnatyd]`.

I've verified: the line endings are correct in the repository.

How to get out of this?

Fabien-Chouteau:
I think this is a default switch of GNAT, we disable it with the Alire configuration system: https://github.com/alire-project/alire/blob/c64c53d5e7420e3a322a12e5fc63fd598a22a390/src/alire/alire-utils-switches.adb#L165

Blady-Com:
Indeed, the switch can be removed but it is required.
Isn't there a way to ensure that the git invocation with Alire preserves the line endings of the files in the Github repository?

mosteo:
You can add attributes to the git repository to force treatment, but it's a can of worms to fight the defaults:

https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings
https://dev.to/kevinshu/git-and-normalization-of-line-endings-228j

Fabien-Chouteau:
> Indeed, the switch can be removed but it is required.

Why is it required?

krischik:
> How to get out of this?

Have your considered adding `.gitattributes` to the repository and manually handling line endings:

```
# force correct line ending on well known files.
#
*.applescript			text eol=lf
*.bat				text eol=crlf
*.btm				text eol=crlf
*.cmd				text eol=crlf
*.command			text eol=lf
*.M65				text eol=lf
*.md				text eol=lf
*.pl				text eol=lf
*.sh				text eol=lf
*.terminal			text eol=lf
*.vim				text eol=lf
*.py				text eol=lf
*.ada   			text eol=lf
*.adb   			text eol=lf
*.ads   			text eol=lf
*.gpr   			text eol=lf
*.toml   			text eol=lf
```

Since the file is added to the repository it should force the agents to use the correct line endings.

Blady-Com:
> Have your considered adding `.gitattributes` to the repository and manually handling line endings:
...
> Since the file is added to the repository it should force the agents to use the correct line endings.

I've added [.gitattributes](https://github.com/Blady-Com/zanyblue/blob/dev_1.5/.gitattributes) in the repo but It doesn't seem having any good result.

What could be wrong?

PS:
> Why is it required?

The author had set these styles switches, I don't wish to change as much as possible.

Blady-Com:
I was hoping to solve the issue when setting my [.gitattributes](https://github.com/Blady-Com/zanyblue/blob/dev_1.5/.gitattributes) with eol = lf.
But I guess something else shall be set in git configuration in the check build script or in Alire.

Fabien-Chouteau:
> The author had set these styles switches, I don't wish to change as much as possible.

I'm sorry but you kind of have to. I tried to work around this issue in git/GitHub, and it always comes back to bite you. 

The other option is to merge with the know problems, it's not great for people using this crate but we can do it.

Blady-Com:
I've finally found a workaround with a GPR configuration: BUILD = "Production" which doesn't activate the style checks.

mosteo (approved):

PR #1828, opened 2026-03-18
pico_ada_c01_blink 1.1.0

krischik:


krischik:
> FAIL: Declared executable not found after compilation: blink.exe

That's not surprising as this is a cross compile for the Raspberry Pi Pico which doesn't result in Windows executables. I did compile on Windows once and you just get the file without the ".EXE" extension. As you should since it's a cross compile.

AldanTanneo (approved):

PR #1831, opened 2026-03-21
pico_ada_c01_blink 1.2.0

krischik:

### DIFF
--- a/index/pi/pico_ada_c01_blink/pico_ada_c01_blink-1.2.0.toml
+++ b/index/pi/pico_ada_c01_blink/pico_ada_c01_blink-1.2.0.toml
@@ -46,7 +46,7 @@ validation.runtime_checks   = "Everything"
 
 [[depends-on]]
 pico_bsp                    = "^2"
-gnatprove                   = "^15"
+# gnatprove                   = "^15" - for future use
 
 # vim: set textwidth=72 nowrap tabstop=8 shiftwidth=4 softtabstop=4 expandtab :
 # vim: set filetype=toml fileencoding=utf-8 fileformat=unix foldmethod=diff :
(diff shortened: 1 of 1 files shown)


index/pi/pico_ada_c01_blink/pico_ada_c01_blink-1.2.0.toml
```diff
@@ -0,0 +1,61 @@
...
+release.compile_checks      = "Warnings"
+release.contracts           = "No"
+release.runtime_checks      = "Default"
+validation.compile_checks   = "Warnings"
+validation.contracts        = "Yes"
+validation.runtime_checks   = "Everything"
+
+[[depends-on]]
+pico_bsp                    = "^2"
+gnatprove                   = "^15"
```
mosteo: If this is for a tutorial, you may want to move `gnatprove` to a nested crate to highlight recommended practices. Or is this on purpose as part of the tutorial?

https://github.com/alire-project/alire/blob/master/doc/catalog-format-spec.md#using-pins-for-crate-testing

krischik: Dang. That's a leftover from a failed experiment. I can't actually prove it as `pico_bsd` will fail prove.

damaki:
Looks like the Windows CI is failing because of this error message:

```
FAIL: Declared executable not found after compilation: blink.exe
```

but this is expected for this crate because the executable is an ELF file built by the `gnat_arm_elf` toolchain instead of the native toolchain.

Could this be a problem with the `alr test` command on Windows?

krischik:
> but this is expected for this crate because the executable is an ELF file built by the `gnat_arm_elf` toolchain instead of the native toolchain.

That's what I suspect as well:

https://github.com/alire-project/alire-index/pull/1828#issuecomment-4084658460

krischik:
@Fabien-Chouteau @mosteo:

The Windows build is looking for an EXE file which doesn't exist because this is a Raspberry Pi Pico executable. 

```console
Success: Build finished successfully in 13.79 seconds.
Found declared GPR file: pico_ada_c01_blink.gpr
FAIL: Declared executable not found after compilation: blink.exe
```

There is nothing I can do about.

AldanTanneo (approved):

PR #1852, opened 2026-04-06
pico_xbsp 1.3.2

krischik:


Fabien-Chouteau (approved):
I would love to see this contributed to the original pico_bsp :)

krischik:
I'not against it but I would need to discuss this with @JeremyGrosser.

PR #1855, opened 2026-04-07
Update external package names for Ubuntu 24 LTS

poldy:
This works for a local-only Alire crate. However, I wouldn't be able to publish for distros other than Ubuntu because it needs a `use "ncursesada.gpr";` at the start of `<crate name>.gpr` . I don't know if there's any way to have that "use" statement be conditional based on if an external from the distro was used or not, a cursory search didn't find anything.

### DIFF
--- a/index/fl/florist/florist-external.toml
+++ b/index/fl/florist/florist-external.toml
@@ -8,4 +8,4 @@ maintainers-logins = ["mosteo"]
 kind = "system"
    [external.origin."case(distribution)"]
    "debian" = ["libflorist2018-dev"]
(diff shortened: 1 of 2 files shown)


mosteo:
This is OK because it doesn't worsen the current situation.

You can't have conditional withs in a `.gpr` file. Best you can do with Alire is selectively use one `.gpr` file or other depending on the detected environment (that is, the `project-files` entry accepts case expressions). This only works for regular, non-external crates, though.

We could have a wrapper crate that provides this to clients transparently, I think.

mosteo (approved):

PR #1859, opened 2026-04-09
munts_rp2040_firmware_embedded release 2.24820.1 Thu Apr  9 05:46:29 …

pmunts:
Phil's RP2040 Microcontroller Embedded Profile Firmware

Initial release.

### DIFF
--- a/index/mu/munts_rp2040_firmware_embedded/munts_rp2040_firmware_embedded-2.24820.1.toml
+++ b/index/mu/munts_rp2040_firmware_embedded/munts_rp2040_firmware_embedded-2.24820.1.toml
@@ -1,0 +1,31 @@
+name = "munts_rp2040_firmware_embedded"
+description = "Phil's RP2040 Microcontroller Embedded Profile Firmware"
+version = "2.24820.1"
+authors = ["Philip Munts"]
+maintainers = ["Philip Munts <phil@munts.net>"]
+maintainers-logins = ["pmunts"]
+licenses = "BSD-1-Clause"
+
+executables = ["munts_rp2040_firmware_embedded"]
+
+[[depends-on]]
+embedded_rp2040 = "^15"
+rp2040_hal = "^2"
+
+[configuration.values]
+embedded_rp2040.Board = "generic_board"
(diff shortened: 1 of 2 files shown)


AldanTanneo (changes requested):
The manifest is missing a "website" field, making your project hard to locate when exploring `alire.ada.dev`

krischik:
Interesting. What is the difference to **embedded_rp2040**? A `long_description` would be nice.

pmunts:
I dropped the web site link for `munts_rp2040_firmware_embedded` because it is a **program** crate.  For a **library** crate, such as `munts_rp2040_lib_embedded`, a web site link makes sense as there are source files to inspect.  For a **program** crate, there really isn't anything else to inspect, as the crate is completely self-contained.

`munts_rp2040_firmware_embedded` and `munts_rp2040_lib_embedded` are perhaps less related than their naming suggests; while the former is intended to be a client for the latter, either can be used without the other.

krischik:
That explains why the compile fails. You have the same problem I have:

```console
Success: Build finished successfully in 48.37 seconds.
Found declared GPR file: munts_rp2040_firmware_embedded.gpr
FAIL: Declared executable not found after compilation: munts_rp2040_firmware_embedded.exe
```

The windows compile looks for and EXE file that doesn't exist.

AldanTanneo:
> a web site link makes sense as there are source files to inspect

I'm of the opinion it always makes sense to have a link to the project homepage, be it a GitHub repo or some other page. Otherwise it shows up on the Alire website without anything to indicate where it comes from.

krischik:
> it always makes sense to have a link to the project homepage

Correct. Same for `long_description`. People will find the crate in the [create list](https://alire.ada.dev/crates.html) and wonder if it's useful for them. It's a public index after all.

@pmunts If these are private packages which are only useful for you and no one else they should not be in the main index at all. If that is the case you should set up a private index. Alire supports private indices. They are very easy to setup and use. Easier then the main index as you don't need to bother with pull requests and you can use `[[pins]]`.

And if they are not private packages they should be properly documented.

AldanTanneo (approved):

PR #1864, opened 2026-04-11
community-bb-runtimes 15.3.0

damaki:


damaki:
Looks like the CI failed because a couple of curl commands to download the crate failed. I suspect this is just a transient failure since it passed on other OSes. Could you try re-running the failed jobs?

krischik:
> Could you try re-running the failed jobs?

You can trigger a re-run yourself by making a null change (add a space, new line or comment) to the alire.toml file. You can do it right here with the online editor.

AldanTanneo (approved):

PR #1889, opened 2026-04-29
bbt 0.3.0

LionelDraghi:
Created via `alr publish` with `alr 2.2.0-dev`

### DIFF
--- a/index/bb/bbt/bbt-0.3.0.toml
+++ b/index/bb/bbt/bbt-0.3.0.toml
@@ -1,0 +1,33 @@
+name = "bbt"
+description = "Black box test runner"
+version = "0.3.0"
+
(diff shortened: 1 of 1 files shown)


mosteo:
There is one test failing.

LionelDraghi:
Sorry Alex for my late reply. I have the impression that the main issue is with installing the rpl utility used in the tests, so it’s probably something in build-test.yml.

I’m not available at the moment to investigate this further, but the software works fine for me on Linux/Debian, Windows, and Intel macOS.

Would it be possible to merge/push it to the repository as it is?

mosteo (approved):

mosteo:
OK, merging then. Thanks @LionelDraghi .

PR #1894, opened 2026-05-01
septum 0.1.1

pyjarrett:


pyjarrett:
Some good stuff in this update:

✅ Added support for unload command to drop directories.
✅ Added list-file command to list a number or all files.
✅ Added loading of all "text-like" files, not just known extensions.
🛠️ Fixed slow printing on large files.

Fabien-Chouteau (approved):

PR #1904, opened 2026-05-09
Create gwindows-1.5.4.toml for GWindows framework

zertovitch:
GWindows release, 09-May-2026 [revision 635]
=========================================

629: GWindows.Clipboard: added Clipboard_HTML procedure (copy HTML fragment
        to the clipboard; useful for copying more than plain texts)

626: GWindows.Common_Controls: added check box feature to List_View's

### DIFF
--- a/index/gw/gwindows/gwindows-1.5.4.toml
+++ b/index/gw/gwindows/gwindows-1.5.4.toml
@@ -1,0 +1,81 @@
+description = "GWindows - Ada Framework for Windows Development"
+name = "gwindows"
+version = "1.5.4"
+
+authors = [
+        "David Botton",
+        "Gautier de Montmollin"
+]
+website = "https://github.com/zertovitch/gwindows"
+licenses = "MIT"
+tags = ["desktop", "framework", "gui", "rad", "toolkit", "windows"]
+maintainers = [
+  "Felix Patschkowski <felix.patschkowski@nexperia.com>",
+  "gdemont@hotmail.com"
+]
+maintainers-logins = [
+  "patschkowski",
+  "zertovitch"
+]
+
+long-description = """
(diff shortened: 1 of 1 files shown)


index/gw/gwindows/gwindows-1.5.4.toml
```diff
@@ -0,0 +1,81 @@
...
+
+
+project-files = [
+        "gnatcom/gnatcom.gpr",
+        "gnatcom/gnatcom_tools.gpr",
+        "gwindows/gwindows.gpr",
+        "gwindows/gwindows_contrib.gpr",
+        "gwindows/gwindows_samples.gpr"
+]
+executables = [
```
AldanTanneo, line 51: Is this supposed to be used as a library? If so, is building the examples when building the library really needed?

zertovitch: In this context, yes.
People who are not yet using the library, but want to give a try, love having a few pre-cooked examples and demos to play with.
Then, they will perhaps look further and may be incentivised to use the library themselves.
There are also a few GNATCOM command-line tools at the end of the list and it is required. It was added in the first place by Felix Patschkowski who made this .toml originally: some users are interested principally at the GNATCOM part.
For people already using the library or crates depending on it, the executables are just a few extra MB on the SSD. It is harmless. Plus it is a test for the linker phase of building something with those bindings.

mosteo: This has been like this for all `gwindows` releases so there's not much harm in having one more, but please consider moving those examples to a documented nested crate or even a separate client crate `gwindows_examples`.

mosteo (approved):

index/gw/gwindows/gwindows-1.5.4.toml
zertovitch: A mysterious issue in the new LEA crate relates exactly to the build of those examples, so I remove that - see PR 1917...

PR #1916, opened 2026-05-15
Create lea-0.94.0.toml for LEA editor

zertovitch:
LEA 0.94
* For Ada or .gpr code, the copy operation puts an HTML-formatted version of the selection into the clipboard.
* Copy operation: when nothing is selected, copy the current line (VS / VS Code - style).
* Added “Alire mode”. The user can control alr build and alr run: Build -> alr build; Run -> alr run.
* LEA checks externally changed files and provide options to the user for handling that.
* Minor improvements (menu texts, colors, ...)

### DIFF
--- a/index/le/lea/lea-0.94.0.toml
+++ b/index/le/lea/lea-0.94.0.toml
@@ -1,0 +1,71 @@
+description = "LEA is a Lightweight Editor for Ada "
+name = "lea"
+version = "0.94.0"
+
+authors = ["Gautier de Montmollin"]
+website = "https://github.com/zertovitch/lea"
+licenses = "MIT"
+tags = ["application", "lea", "editor", "hac"]
+maintainers = ["gdemont@hotmail.com"]
+maintainers-logins = ["zertovitch"]
+
+long-description = """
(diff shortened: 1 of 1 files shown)


zertovitch:
Some surnatural entity must have read the discussion about building executables of examples in the GWindow library...
For a mysterious reason, it works with the GWindows crate itself, but here a file is not found by the linker.
Also, there was no issue previously.
An easy solution is to make the proposed change in the GWindows crate, it should solve the issue here and make the whole crate build much faster. I do it right now - See PR 1917.
I will later trigger a re-run of the tests here.

zertovitch (commented):
Triggering a new test

zertovitch:
The current issue is:
x86_64-w64-mingw32/bin/ld.exe: cannot find ..\..\MemoryModule.o: No such file or directory

It is related to this line somewhere in the code:
pragma Linker_Options ("..\..\MemoryModule.o");

Does the linking happen in a directory different than the object directory? A new gprbuild behaviour?

zertovitch:
Issue is solved!

mosteo (approved):

zertovitch:
Gr8!
The related change in the LEA repo is:
https://github.com/zertovitch/lea/commit/6bcff7fe7f80a1e9773f1bf50f7aceb69eac0454

PR #1922, opened 2026-05-18
ada message pack

mlorek:


index/ad/ada_msg_pack/ada_msg_pack-0.1.0.toml
```diff
@@ -0,0 +1,23 @@
+name = "ada_msg_pack"
```
mosteo: We generally aren't very fond of crate names that start with `ada` because, in the Ada ecosystem, that's normally redundant. It's not strictly forbidden, but since this is the first version of the crate you have the opportunity to drop it. Let me know if you still want to go ahead with this name.

Note that the crate name can be different from the repo one.

mosteo:
The manifest location in the index now must match the new crate name.

mosteo:
> Tag string is not valid: msg_pack

You can use a dash instead.

index/ms/msg_pack/msg_pack-0.1.0.toml
```diff
@@ -0,0 +1,27 @@
...
+website = "https://github.com/mlorek/msgpack"
+tags = ["serialization", "msgpack", "messagepack", "msg-pack"]
+
+[[actions]]
+type = "test"
+command = ["alr", "build"]
+
+[[actions]]
+type = "test"
+command = ["alr", "exec", "--", "gprbuild", "-p", "-P", "tests/tests.gpr"]
```
krischik, line 18: You can use an Alire for tests as well. Example: https://sourceforge.net/p/adacl/git/ci/develop/tree/adacl/test/

(No fix needed. Just and idea for the next release).

krischik:
@mosteo @Fabien-Chouteau @AldanTanneo I think all the requested changes habe been made. Could you trigger the workflow?

krischik:
@mosteo @Fabien-Chouteau @AldanTanneo The workflow needs starting again and I think this crate is ready to be merged as well. All the requested changes have been done.

Fabien-Chouteau (approved):

PR #1937, opened 2026-05-27
ada_imgui 0.1.0

tonygair:
**Crate**: `ada_imgui` 0.1.0
**Source**: https://github.com/the-dark-factory/ada-imgui
**License**: MIT
**Maintainer**: The Dark Factory Ltd

Bindings to Dear ImGui via cimgui. Immediate-mode GUI for Ada applications.

### Build notes

cimgui (the C++ side) needs to be built once before `alr build`; `scripts/build-cimgui.sh` in the source repo does this. Documented in the crate README. Tested on macOS aarch64 + Alire-shipped GCC 14.

### Context

This is one of three Ada bindings being submitted today by The Dark Factory Ltd (https://github.com/the-dark-factory/ada-imgui → see also `ada-imgui`, `ada-stb`, `ada-ccv` on the same GitHub org). The line is MIT-licensed, dedicated to the Ada community, with the same maintenance posture across the three crates.

Marked draft for initial review — happy to iterate on whatever the index reviewers want (Alire actions for the C-side build, additional metadata, etc.). Flag the concerns and I'll address.

### DIFF
--- a/index/ad/ada_imgui/ada_imgui-0.1.0.toml
+++ b/index/ad/ada_imgui/ada_imgui-0.1.0.toml
@@ -1,0 +1,31 @@
+name = "ada_imgui"
+description = "Ada bindings to Dear ImGui via cimgui"
+version = "0.1.0"
+authors = ["The Dark Factory Ltd"]
+maintainers = ["tony.gair@thedarkfactory.co.uk"]
+maintainers-logins = ["tonygair"]
+licenses = "MIT"
+website = "https://github.com/the-dark-factory/ada-imgui"
+tags = ["binding", "gui", "imgui", "cimgui"]
+project-files = ["ada_imgui.gpr"]
+auto-gpr-with = false
+
+[build-switches]
+"*".style_checks = ["-gnaty"]
+
+#  cimgui (the C++ side) is currently vendored under vendor/cimgui
+#  and built by scripts/build-cimgui.sh. Until that build step is
+#  expressed as an Alire action or the C++ library is packaged as
+#  a separate crate, downstream users must run
+#  `./scripts/build-cimgui.sh` once before `alr build`. Documented
+#  in README.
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Hello @tonygair , thanks for your contribution.

I have a few comments that are applicable to all three PRs:

- We highly recommend not having crates with names that start with `ada_`. If you really want to show these are binding, it's better to use a `_ada` suffix so `imgui_ada`, similar to how the Rust community is using a `-rs` suffix for bindings.
- Instead of asking users to compile the C projects by hand., you can try to do is to compile the C code using gprbuild. This is what I did in this binding for instance: https://github.com/Fabien-Chouteau/littlefs-ada In particular those two lines in littlefs.gpr: https://github.com/Fabien-Chouteau/littlefs-ada/blob/64a02efecd2ad7d55abe04b767ae48c0b187feea/littlefs.gpr#L7
```
   for Languages use ("Ada", "C");
   for Source_Dirs use ("src/", "config/", "littlefs/");
```
- `.gpr` files are not "standard", I highly recommend using the `.gpr` file generated by `alr init` and only doing small modifications when needed. Looking at `ada_imgui.gpr`, you have a switch that is only compatible with macOS (`-Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk`), and you don't have different build modes for release vs development situations. 
- In the `ada_stb` project, `ada_stb.ads` file, you have declared a crate version constant `Crate_Version : constant String := "0.1.0-dev";` This is redundant with Alire's crate configuration system which will provide this constant, and will always be in sync with the version defined in `alire.toml`

tonygair:
Renamed crate to imgui_ada (repo the-dark-factory/imgui-ada) to match the <lib>_ada convention; resubmitting.

PR #1940, opened 2026-05-31
adael 1.0.0

krischik:


krischik:
```
REVIEW: crate 'adael' starts with 'ada'
Error: Process completed with exit code 1.
```

@mosteo I have to think about the name. I wanted to name of the embedded library to match the name Desktop library ("E" for embedded and "C" class) but I see you have tightened the rules. I could use "AdaCL-embedded" - which might even be more recognisable. Or something entirely new.

mosteo:
We could make an exception, but to be frank, I didn't connect adael to adacl.

krischik:
You are right. It doesn't connect.

PR #1941, opened 2026-06-02
crdt 1.1.0 (via `alr publish`)

bladeacer:


index/cr/crdt/crdt-1.1.0.toml
```diff
@@ -0,0 +1,26 @@
...
+maintainers = ["bladeacer <wg.nick.exe@gmail.com>"]
+maintainers-logins = ["bladeacer"]
+licenses = "MIT"
+website = "https://github.com/bladeacer/ada_crdt"
+tags = ["crdt", "spark", "ada"]
+
+executables = ["test_crdt"]
+
+[[depends-on]]
+gnatprove = "^15.1.0"
```
Fabien-Chouteau: `gnatprove` and `gnatdoc_bin` are tools used only for development and should not be listed as release dependencies.

Unfortunately, we don't have a way to specify a development dependency in Alire. But you can remove the lines from the crdt-1.1.0.toml manually.

bladeacer: Can the repo's alire.toml still have the aforementioned dev dependencies?

bladeacer: Will push a fix for this.

Fabien-Chouteau (approved):

PR #1948, opened 2026-06-05
stb_ada 0.1.0

tonygair:

### DIFF
--- a/index/st/stb_ada/stb_ada-0.1.0.toml
+++ b/index/st/stb_ada/stb_ada-0.1.0.toml
@@ -1,0 +1,30 @@
+name = "stb_ada"
+description = "Ada bindings to Sean Barrett's STB header-only utility libraries"
+version = "0.1.0"
+authors = ["The Dark Factory Ltd"]
+maintainers = ["tony.gair@thedarkfactory.co.uk"]
+maintainers-logins = ["tonygair"]
+licenses = "MIT"
+website = "https://github.com/the-dark-factory/stb-ada"
+tags = ["binding", "image", "stb", "stb-image", "stb-truetype"]
+project-files = ["stb_ada.gpr"]
+auto-gpr-with = false
+
+[build-switches]
+"*".style_checks = ["-gnaty"]
+
+#  STB is header-only C. csrc/stb_impl.c instantiates the
+#  STB_*_IMPLEMENTATION bodies; scripts/build-stb.sh compiles
+#  it into vendor/stb/libstb.a. Until that step is wired as
+#  an Alire action, downstream consumers must run
+#  `./scripts/build-stb.sh` once before `alr build`.
+
+[available.'case(os)']
+macos = true
+linux = true
+windows = false  # Untested.
+
+[origin]
+commit = "008823b9164e888608461613eb94be8b749bb865"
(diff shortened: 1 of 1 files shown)


index/st/stb_ada/stb_ada-0.1.0.toml
```diff
@@ -0,0 +1,30 @@
...
+maintainers = ["tony.gair@thedarkfactory.co.uk"]
+maintainers-logins = ["tonygair"]
+licenses = "MIT"
+website = "https://github.com/the-dark-factory/stb-ada"
+tags = ["binding", "image", "stb", "stb-image", "stb-truetype"]
+project-files = ["stb_ada.gpr"]
+auto-gpr-with = false
+
+[build-switches]
+"*".style_checks = ["-gnaty"]
```
Fabien-Chouteau, line 14: Enabling style checking across all modes isn't ideal, since future versions of GNAT may introduce new style checks that might fail with this release of your code.

In any case, you are using a custom .gpr file that does not use the configuration switches generated by Alire. So these `build-switches` settings here will have absolutely no effect.

index/st/stb_ada/stb_ada-0.1.0.toml
```diff
@@ -0,0 +1,30 @@
...
+#  STB is header-only C. csrc/stb_impl.c instantiates the
+#  STB_*_IMPLEMENTATION bodies; scripts/build-stb.sh compiles
+#  it into vendor/stb/libstb.a. Until that step is wired as
+#  an Alire action, downstream consumers must run
+#  `./scripts/build-stb.sh` once before `alr build`.
+
+[available.'case(os)']
+macos = true
+linux = true
+windows = false  # Untested.
```
Fabien-Chouteau, line 25: Untested or you know it won't work?

Fabien-Chouteau (commented):
Hello @tonygair ,

I am reinstating the comments from the previous PR (#1937), we can merge the crate as is if you want to, but please consider items below to make it more useful for users:

- Instead of asking users to compile the C projects by hand., you can try to do is to compile the C code using gprbuild. This is what I did in this binding for instance: https://github.com/Fabien-Chouteau/littlefs-ada In particular those two lines in littlefs.gpr: https://github.com/Fabien-Chouteau/littlefs-ada/blob/64a02efecd2ad7d55abe04b767ae48c0b187feea/littlefs.gpr#L7
```
   for Languages use ("Ada", "C");
   for Source_Dirs use ("src/", "config/", "littlefs/");
```
- `.gpr` files are not "standard", I highly recommend using the `.gpr` file generated by `alr init` and only doing small modifications when needed. Looking at `ada_imgui.gpr`, you have a switch that is only compatible with macOS (`-Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk`), and you don't have different build modes for release vs development situations. 
- In the `ada_stb` project, `ada_stb.ads` file, you have declared a crate version constant `Crate_Version : constant String := "0.1.0-dev";` This is redundant with Alire's crate configuration system which will provide this constant, and will always be in sync with the version defined in `alire.toml`

tonygair:
Superseded by #1955 — renamed `stb_ada` → `df_stb` to satisfy the new-crate-name policy.

PR #1955, opened 2026-06-08
df_stb 0.1.0

tonygair:
Supersedes #1948 — renamed `stb_ada` → `df_stb` to satisfy the new-crate-name policy (no `ada`/`spark` in new crate names). `df_` is our publisher namespace (The Dark Factory). Otherwise unchanged: Ada bindings to Sean Barrett's STB header-only utility libraries.

### DIFF
--- a/index/df/df_stb/df_stb-0.1.0.toml
+++ b/index/df/df_stb/df_stb-0.1.0.toml
@@ -9,15 +9,14 @@ website = "https://github.com/the-dark-factory/stb-ada"
 tags = ["binding", "image", "stb", "stb-image", "stb-truetype"]
 project-files = ["df_stb.gpr"]
 auto-gpr-with = false
+notes = "STB is header-only C; the post-fetch action builds vendor/stb/libstb.a from csrc/stb_impl.c (plain cc, no extra tooling)."
 
 [build-switches]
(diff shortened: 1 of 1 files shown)


index/df/df_stb/df_stb-0.1.0.toml
```diff
@@ -0,0 +1,29 @@
...
+auto-gpr-with = false
+
+[build-switches]
+"*".style_checks = ["-gnaty"]
+
+#  STB is header-only C. csrc/stb_impl.c instantiates the
+#  STB_*_IMPLEMENTATION bodies; scripts/build-stb.sh compiles
+#  it into vendor/stb/libstb.a. Until that step is wired as
+#  an Alire action, downstream consumers must run
+#  `./scripts/build-stb.sh` once before `alr build`.
```
mosteo: Same comment on putting this into a `notes` field. Any reason why this doesn't work already as a post-fetch action?

tonygair: Done — wired build-stb.sh as a post-fetch action (it's a plain `cc` compile, no extra tooling) and moved the note into the `notes` field. Re-pinned to c3442ad.

mosteo (approved):

PR #1957, opened 2026-06-08
df_box2d 0.1.0

tonygair:
Ada bindings to Box2D v3 (Erin Catto's pure-C 2D physics engine). Generated thin C-ABI binding. The df_ prefix is our publisher namespace (The Dark Factory).

### DIFF
--- a/index/df/df_box2d/df_box2d-0.1.0.toml
+++ b/index/df/df_box2d/df_box2d-0.1.0.toml
@@ -1,0 +1,38 @@
+name = "df_box2d"
+description = "Ada bindings to Box2D v3, Erin Catto's 2D physics engine (pure C)"
+version = "0.1.0"
+authors = ["The Dark Factory Ltd"]
+maintainers = ["tony.gair@thedarkfactory.co.uk"]
+maintainers-logins = ["tonygair"]
+licenses = "MIT"
+website = "https://github.com/the-dark-factory/box2d-ada"
+tags = ["binding", "physics", "box2d", "game", "simulation"]
+project-files = ["df_box2d.gpr"]
+auto-gpr-with = false
+
+#  The src/ binding is machine-generated (gcc -fdump-ada-spec), so the
+#  usual style checks don't apply to it.
+[build-switches]
+"*".style_checks = "no"
+
+#  `alr test` runs the behavioural smoke test (build against the real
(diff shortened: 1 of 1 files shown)


index/df/df_box2d/df_box2d-0.1.0.toml
```diff
@@ -0,0 +1,38 @@
...
+#  has no per-subprogram units worth AUnit-testing (its imports are the C
+#  ABI), so the test is behavioural, exit-code based, with no AUnit dependency.
+[[actions]]
+type = "test"
+command = ["bash", "scripts/test.sh"]
+
+#  Box2D v3 (vendor/box2d) is built into vendor/box2d/build/src/libbox2d.a by
+#  scripts/build-box2d.sh (CMake). Until that step is wired as an Alire action,
+#  downstream consumers must run `./scripts/build-box2d.sh` once before
+#  `alr build`. Requires cmake and a C compiler.
```
mosteo: Sorry to repeat myself...

tonygair: Moved into the notes field. Left as a documented step rather than a post-fetch action because the vendored Box2D build needs CMake, which is not guaranteed on the build host.

index/df/df_box2d/df_box2d-0.1.0.toml
```diff
@@ -0,0 +1,25 @@
...
+authors = ["The Dark Factory Ltd"]
+maintainers = ["tony.gair@thedarkfactory.co.uk"]
+maintainers-logins = ["tonygair"]
+licenses = "MIT"
+website = "https://github.com/the-dark-factory/box2d-ada"
+tags = ["binding", "physics", "box2d", "game", "simulation"]
+project-files = ["df_box2d.gpr"]
+auto-gpr-with = false
+
+notes = "Box2D v3 is vendored and built into a static lib by scripts/build-box2d.sh (CMake); run it once before alr build. Requires cmake and a C compiler."
```
mosteo, line 13: If you want to automate this, you can provide an external definition for `cmake`, and make this crate depend on it. See e.g. https://github.com/alire-project/alire-index/blob/stable-1.4.0/index/ma/make/make-external.toml

You don't need to provide both kinds of external (version and system package). Usually the package one is enough and preferred because it triggers installation.

mosteo (approved):

PR #1963, opened 2026-06-10
simple_components metadata: fix casing of crate names

mosteo:


mosteo:
Error is unrelated to the change.

PR #1971, opened 2026-06-13
Release/pico xbsp 1.8.0

krischik:
Next version of pico xbsp as preparation for chapter 8 of the Pico Ada Tutorial.


krischik:
What happened here?

PR #1973, opened 2026-06-15
libsap 0.3.0

damaki:

### DIFF
--- a/.github/workflows/build-distros.yml
+++ b/.github/workflows/build-distros.yml
@@ -69,7 +69,7 @@ jobs:
         branch: master
 
     - name: Test crate (${{matrix.tag}})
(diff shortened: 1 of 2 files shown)


damaki:
Looks like some actions are failing because they're using [mosteo-actions/docker-run@v1](https://github.com/mosteo-actions/docker-run) which seems to be based on Node.js 20 (judging by [this line](https://github.com/mosteo-actions/docker-run/blob/37d5df0a34d24b220e607d6d9cf96c282c19cfc2/action.yml#L31)) which is now end-of-life. Perhaps it needs to be updated to Node24?

mosteo:
That should be fixed now, thanks for the report.

mosteo (approved):

PR #1977, opened 2026-06-17
AdaCL release 7.1.2

krischik:
Update adacl_embedded and minor fixes.


krischik:
@mosteo @Fabien-Chouteau @AldanTanneo  Apart from the name class all build fail with:

```
Error: Cannot find module '@actions/core'
```

That happens to all builds and all users and is very confusing to the user.

Fabien-Chouteau (approved):

PR #1979, opened 2026-06-18
pico_xbsp 1.8.0

krischik:


krischik:
Depends on [pull request 1977](https://github.com/alire-project/alire-index/pull/1977). Please do #1977 first and restart the workflow. The build should then pass without missing dependencies.

Fabien-Chouteau (approved):

PR #2000, opened 2026-06-30
gnatprove 16.1.0

AldanTanneo:


AldanTanneo:
There is a crash on windows-latest where I'm guessing Git is configured to convert LF newlines to CRLF newlines, and the amount of error messages is making the error output crash because it's trying to flatten everything on the stack? Or it's an infinite recursion bug in AAA?

It doesn't seem to happen on alr dev, so maybe it's an Alire issue that is fixed in nightly?

@mosteo

mosteo:
This has started happening recently and I don't remember immediately any change that could have triggered it.

mosteo (approved):

PR #2003, opened 2026-07-02
gnat+cross 15.3.1

AldanTanneo:
This is the latest point release of GNAT FSF 15.

It includes two new cross compilers: one for aarch64-elf and another for x86_64-elf.


AldanTanneo:
The macOS failure should be fixed by this compiler update :)

AldanTanneo:
I'm making an executive decision and merging this. It can always be rolled back later if need be.

PR #2004, opened 2026-07-06
bbs_simcpu 0.5.0

BrentSeidel:


BrentSeidel:
Funny that it fails on the MacOS build, since that's what I use for development.  I've seen the following reported in other PRs, so it may be a know issue:

   25 | #include <stdlib.h>
      |          ^~~~~~~~~~

krischik:
> it may be a know issue

It is indeed a known issue.

krischik:
      apparent.
- [ ] For repeat releases, check the Diff workflow output to quickly spot the
- [x] `gpr-set-externals` settings are appropriate (see Common pitfalls below).
- [x] System library dependencies are declared and not left implicit. 

@mosteo @Fabien-Chouteau @AldanTanneo Looks OK.

AldanTanneo (approved):

PR #2012, opened 2026-07-11
adacl_aunit 7.2.0

krischik:
Test simple unit test call.


krischik:
@mosteo This one is all green but it also doesn't use `gnatprove` and it has only has no actual unit tests (just the empty framework). They still need to be added.

mosteo (approved):

PR #2013, opened 2026-07-11
release adacl_eastrings 7.2.0

krischik:
Re-enable unit tests.


krischik:
Added regex - makes it simpler for me and the moderators.

krischik:
@mosteo just one is missing. But I post that separately as there are problems with the unit tests and the whole release is about reactivating the unit test.

mosteo (approved):

PR #2014, opened 2026-07-12
release adacl 7.2.0

krischik:
With unit tests using scripts to call the test. This release is a bit of an Experiment how unit tests work in the git worker.


krischik:
@mosteo Success - including test. I can see the tests results in the log. My local Windows did all sorts of strange stuff so I'm glad it worked just fine.

mosteo (approved):

PR #2040, opened 2026-07-29
standard_atomic_ops 1.0.0

pat-rogers:

### DIFF
--- a/index/st/standard_atomic_ops/standard_atomic_ops-1.0.0.toml
+++ b/index/st/standard_atomic_ops/standard_atomic_ops-1.0.0.toml
@@ -1,0 +1,30 @@
+name = "standard_atomic_ops"
+version = "1.0.0"
+description = "Language-defined atomic operation packages for GNAT runtimes."
+long-description = """
+This crate supplies the GNAT source files for the Ada 2022 language-defined units for atomic operations: System.Atomic_Operations, System.Atomic_Operations.Exchange, System.Atomic_Operations.Modular_Arithmetic, and System.Atomic_Operations.Integer_Arithmetic.
+They are provided for use with those bare-board GNAT runtimes that do not (yet) include them, for example Arm-elf (the only target tested to date). 
+The sources here are verbatim copies taken from a native runtime. As such, the implementations are GNAT-specific so their use with other compilers will not work.
+Once the shipped GNAT runtimes include them you can simply remove any references to this crate.
+Note that the license includes the GCC Runtime Library Exception.
+"""
+licenses = "GPL-3.0-or-later WITH GCC-exception-3.1"
+authors = ["AdaCore"]
+maintainers = ["Pat Rogers <progers@classwide.com>"]
+maintainers-logins = [
+"pat-rogers",
+]
+website = ""
+project-files = [
+"standard_atomic_ops.gpr",
+]
+tags = [
+"atomic",
+"standard",
+"runtime"
+]
+
+[origin]
+commit = "4443a5b4b2f1c9171258a3563a8567311aa05de4"
+url = "git+https://github.com/pat-rogers/ada2022_atomic_ops.git"
+


pat-rogers:
As was the case with a previous crate, I believe the build failures on Windows are a known problem.

mosteo:
I'm not 100% sure we can handwave the error on Windows because those are compiler packages that require some exact switches for compilation, right?

Maybe it's worth a try to add a `.gitattributes` to the repo with `* text=auto eol=lf` to force the line endings.

My concern is that nobody will be able to build those on Windows if GNAT requires the style + warnings as error checks when building compiler units, but I'm far from sure about this. If you think this is OK as-is, just let me know and I'll merge.

pat-rogers:
Hi Alejandro,
(Good to hear from you again!)
I cannot see the build errors in my browser. I just see one line with the word "Error" in red. But when I did the publish step locally again just now (skipping the actual submit) I still got no errors for the build. And this is on Windows. 
These files do require the "-g" compiler switch located in the crate's gpr file.
What errors are presented? And did something happen with line-endings in the sources? They should be Unix-style.
Thanks,
Pat

mosteo:
Hi Pat (also great to see you over here),

The two Windows checks are failing with a stream that ends in:
```
s-atoope.ads:35:01: (style) incorrect line terminator [-gnatyd]
s-atoope.ads:36:33: (style) incorrect line terminator [-gnatyd]
s-atoope.ads:37:12: (style) incorrect line terminator [-gnatyd]
s-atoope.ads:38:03: (style) incorrect line terminator [-gnatyd]
s-atoope.ads:39:30: (style) incorrect line terminator [-gnatyd]
standard_atomic_ops.gpr: error: [Ada Compile] s-atoope.ads (standard_atomic_ops.gpr) failed with status 1
stderr: Command ["gprbuild", "-s", "-j0", "-p", "-P", "D:\a\alire-index\alire-index\standard_atomic_ops_1.0.0_4443a5b4\standard_atomic_ops.gpr"] exited with code 4
```
(Maybe you can visit the error log directly at https://github.com/alire-project/alire-index/actions/runs/30484251470/job/90685823649?pr=2040 )

I'm not sure right now why it passes locally for you. There must be some difference in build switches or your local checkout. If indeed you're using Unix-style line endings locally, I would ask if you did something deliberate to get Unix-style line endings on Windows.

pat-rogers:
> The two Windows checks are failing with a stream that ends in:
> 
> |s-atoope.ads:35:01: (style) incorrect line terminator [-gnatyd] s- 
> atoope.ads:36:33: (style) incorrect line terminator [-gnatyd] s- 

OK, thanks.

> (Maybe you can visit the error log directly at https://github.com/alire- 
> project/alire-index/actions/runs/30484251470/job/90685823649?pr=2040 
> <https://github.com/alire-project/alire-index/actions/runs/30484251470/ 
> job/90685823649?pr=2040> )

Yes, I can see it now. It was a browser plugin issue on my side.

> I'm not sure right now why it passes locally for you. There must be some 
> difference in build switches or your local checkout. If indeed you're 
> using Unix-style line endings locally, I would ask if you did something 
> deliberate to get Unix-style line endings on Windows.

I have my tools set to use Unix-style, even though I am on Windows. That was an AdaCore requirement and I have just kept following that rule, so the source files are Unix-style when I examine them locally.

The cause was on my side: the crate repositories carried no .gitattributes, so a checkout on a Windows runner with core.autocrlf=true converted the sources to CRLF.

As a result, every crate repository I am preparing now (including this one) carries a root .gitattributes, containing:

   * text=auto eol=lf

and the trees have been checked, so every tracked text blob is definitely stored with LF.

I believe that should fix the problem. Sorry for the noise!

How would you like me to proceed?

Thanks,

Pat

pat-rogers:
I will close this PR and issue another one with the updated project that includes the .gitattributes, such that the line-endings will be Unix-style, and we can then see if that does indeed solve the problem.

pat-rogers:
I opened #2045, which passes.

PR #2042, opened 2026-08-01
glyph 0.1.0

thirstymelon:

### DIFF
--- a/index/gl/glyph/glyph-0.1.0.toml
+++ b/index/gl/glyph/glyph-0.1.0.toml
@@ -1,0 +1,16 @@
+name = "glyph"
+version = "0.1.0"
+description = "Lightweight embedded graphics framework for Ada 2022"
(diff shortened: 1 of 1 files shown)


mosteo (approved):

mosteo:
In regard to https://github.com/thirstymelon/glyph/blob/93655627ec27a04b9ae8a8e05e44017da4e04220/src/glyph.ads#L19

Alire will provide this info using the manifest as source of truth in the `config/` generated files, so you might consider using that instead to avoid inadvertent skew.

mosteo:
@thirstymelon , FYI since this is your first submission, you don't need to merge changes in the index while your crate is under review (unless of course you depend on those changes). The squash merge policy we use is unaffected otherwise.

No harm either if you do, but it will retrigger checks that may have already passed.

PR #2043, opened 2026-08-03
cachelines 0.1.0

yrashk:


AldanTanneo:
- If the crate is not available on windows, you can specify it in the `alire.toml`:
  ```toml
  [available.'case(os)']
  windows = false
  '...' = true
  ```
- If it's meant for embedded and such, please note that the way you set the alignment in the GPR files relies on Alire host OS detection, _not_ target OS.

yrashk:
Addressed in [yrashk/ada-cachelines@5c97574](https://github.com/yrashk/ada-cachelines/commit/5c975744fccb618712e256fd230dce7081ceade2): Windows is now declared unavailable, and cross/embedded builds use documented explicit target architecture and OS overrides with test coverage.

mosteo (approved):

PR #2044, opened 2026-08-03
covex 1.3.0 (via `alr publish`)

bladeacer:
Changelogs: https://github.com/bladeacer/adacovex/blob/main/docs/changelogs/adacovex-1.3.0.md

gnatprove added as a dependency, as adacovex specifically relies on it. (Future language support would also rely on alire to manage gnatprove instead of expecting the user to manually install/configure it.

It is also added in anticipation of language agnostic support (bringing gnatprove and adacovex to other non-Ada languages by transpiling to Ada IR)


bladeacer:
Closing as gnatprove should not be a dependency.

PR #2048, opened 2026-08-05
covex 1.5.0 (via `alr publish`)

bladeacer:
**Changelogs**
https://github.com/bladeacer/adacovex/blob/main/docs/changelogs/adacovex-1.5.0.md


bladeacer:
Closing as gnatprove should not be a dependency.

PR #2056, opened 2026-08-09
ZB 2.0.1: Upgrade UXStrings >= 0.9.4.

Blady-Com:


index/zb/zblib/zblib-2.0.1.toml
```diff
@@ -0,0 +1,20 @@
...
+maintainers = ["Pascal Pignard <blady.net@orange.fr>"]
+maintainers-logins = ["Blady-Com"]
+licenses = "custom-zanyblue"
+website = "https://sourceforge.net/projects/zanyblue"
+tags = ["localization", "l10n", "international", "i18n"]
+
+project-files = ["src/zanyblue.gpr"]
+
+[[depends-on]]
+uxstrings = ">=0.9.4"
```
mosteo, line 15: This is a bit unorthodox as that will accept breaking changes in 1.x or any other major version. Normally one would want something like `"~0.9.4"`. 

If you indeed want `">="`, just confirm and I'll merge.

Blady-Com: I confirm with >=.

mosteo (approved):

PR #2059, opened 2026-08-14
uarp_sdk 0.5.14

chabanov:
Ada client for the UARP (Snaga) Universal Agent Runtime Platform API, generated
from the platform's OpenAPI document: full coverage of the API surface, with the
transport, error, retry, pagination and event-stream layers written by hand.

- Source: https://github.com/Snaga-AI/uarp-sdks/tree/main/packages/ada
- Licence: MIT
- Origin: a source archive attached to the `v0.2.0` release, `sha256` in the manifest

### Checked before submitting

- Builds and its 93 tests pass on Ubuntu in CI, on GNAT 15 via `gnat_native^15`,
  which is where this index builds crates.
- `alr get uarp_sdk` retrieves and deploys the crate through this manifest, so
  the origin URL and hash are verified against the published archive rather than
  a local copy.
- `alr show --solve` resolves cleanly: `gnatcoll=26.0.0`, `gnat=15.3.1`.

### One thing worth flagging

The crate links libcurl through a small non-variadic C shim. `libcurl` is not
declared as an Alire dependency because the `libcurl` external only detects
Linux package managers, and the crate is also built on macOS, where libcurl
ships with the system. On Linux it needs `libcurl4-openssl-dev` (or the
equivalent) present. Happy to declare it as an external instead if you would
rather it were explicit.


chabanov:
Updated this to 0.3.0.

0.2.0 stopped paginated iteration at the first empty page. This API applies the page size before filtering, so it answers a request for two items with none while `has_more` is still true — a caller would have been told a collection was empty when it was not. 0.2.0 is now deprecated on npm and yanked on crates.io, so it is not a version worth putting in the index.

Same checks as before, redone against the new archive: `alr get uarp_sdk` retrieves and deploys 0.3.0 from the published tarball, and the crate builds with its 93 tests passing on Ubuntu with `gnat_native^15`.

chabanov:
Updated this to **0.5.3**.

0.3.0 sat here unreviewed while four releases landed upstream, and it predates
two fixes that matter for anyone who would have picked it up:

- the SSE parser read only one of the three wire shapes the platform actually
  emits, so a completed run could decode as an empty stream;
- a client whose credentials travel another way sent `Bearer ` with nothing
  after it, which a validating server can refuse.

Submitting the current release instead of one with known defects. The manifest
is generated by `alr publish` (alr 2.1.1), unedited — it carries both sha256
and sha512, where the 0.3.0 file had sha256 only. The source archive is
attached to the upstream release:
https://github.com/Snaga-AI/uarp-sdks/releases/tag/v0.5.3

Apologies for the noise: I opened #2065 for 0.5.3 before noticing this one was
still open. Closing that one in favour of this, which is the older queue
position.

mosteo:
> On Linux it needs `libcurl4-openssl-dev` (or the
> equivalent) present. Happy to declare it as an external instead if you would
> rather it were explicit.

Build errors are because of this. Given that `libcurl` is already available as an external definition, I think it would be best to add it to the manifest. You can conditionally depend on it only on Linux/Windows:

```
[[depends-on]]
[depends-on."case(os)"."linux|windows"]
libcurl = "*"
```

chabanov:
Updated to **0.5.13**, and it carries the change you asked for.

The manifest now declares libcurl conditionally, in your spelling:

```toml
[[depends-on]]
[depends-on."case(os)"."linux|windows"]
libcurl = "*"
```

`alr show` lists `when Linux => libcurl*`, `when Windows => libcurl*`; on macOS nothing is declared, since libcurl ships with the system, and `alr show --solve` resolves without it. The crate's own CI builds it on Ubuntu with `libcurl4-openssl-dev` installed; the runners here are the proof that the external resolves where the package is absent.

Also in 0.5.13 relative to 0.5.3: the gnatcoll dependency is `>=25.0.0 & <27.0.0` (the crate uses GNATCOLL.JSON only), so it resolves next to `aws 25.2` (`gnatcoll ~25.0.0`), which the caret on 26 made impossible.

The manifest is generated by `alr publish` (alr 2.1.1), unedited, with sha256 and sha512; the source archive is attached to the upstream release: https://github.com/Snaga-AI/uarp-sdks/releases/tag/v0.5.13

Thank you for the pointer, and for your patience with a submission that has moved three times.

mosteo:
There are linker errors on Windows:
```
c:/users/runneradmin/appdata/local/alire/cache/toolchains/gnat_native_16.1.0_bbd69633/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/as.exe: uarp-models.o: too many sections (164648)
C:\Users\RUNNER~1\AppData\Local\Temp\ccw9DEHT.s: Assembler messages:
C:\Users\RUNNER~1\AppData\Local\Temp\ccw9DEHT.s: Fatal error: can't write 7 bytes to section .text$uarp__models__a2a_task_vectors__first_index of uarp-models.o: 'file too big'
c:/users/runneradmin/appdata/local/alire/cache/toolchains/gnat_native_16.1.0_bbd69633/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/as.exe: uarp-models.o: too many sections (164648)
C:\Users\RUNNER~1\AppData\Local\Temp\ccw9DEHT.s: Fatal error: uarp-models.o: file too big
uarp_sdk.gpr: error: [Ada Compile] uarp-models.adb (uarp_sdk.gpr) failed with status 1
```
I'm not immediately aware of how to fix that.

chabanov:
@mosteo Thanks — that was a real break, and the submission now points at 0.5.14, which fixes it.

It is not a linker problem. A PE-COFF object header holds its section count in 16 bits, so 65535 is the ceiling, and `uarp-models.o` asked for 164648. It asked because `uarp-models.ads` declares 1824 subprograms and instantiates `Ada.Containers.Vectors` 126 times, all of it landing in that one object, and because the build profile passes `-ffunction-sections -fdata-sections` — one section per subprogram, plus its `.pdata$` and `.xdata$` unwind sections on x86-64 Windows. The generated `config/uarp_sdk_config.gpr` even names the assumption that fails here:

```
,"-ffunction-sections" -- Separate ELF section for each function
```

ELF has no such ceiling, which is why every Linux and macOS runner was happy and nothing in our own CI could see it.

The fix is those two switches turned back off, for that one unit, in `uarp_sdk.gpr`:

```ada
for Switches ("uarp-models.adb") use
  Ada_Switches & ("-fno-function-sections", "-fno-data-sections");
```

It costs only the linker's dead-code stripping inside `uarp-models.o`; the rest of the crate stays sectioned as the profile asks.

Verified on Windows 10 with alr 2.1.1, gnat_native 15.3.1 and gprbuild 26.0.1, end to end from this manifest — the fork added as a local index, then:

| | |
|---|---|
| `alr get uarp_sdk=0.5.14` | tarball fetched, both hashes checked, solved to gnatcoll 26.0.0 and libcurl 8.21.0 (system package) |
| `alr build` in the retrieved crate | `libuarp_sdk.a` archived, **100** sections in `uarp-models.o`, down from 176494 |
| crate's own suite | **127/127**, over real HTTP and SSE |

The count differs from your 164648 because that log is a release build and mine is development; the ceiling is the same either way.

Our CI also gained a `windows-latest` job for the Ada crate, so this class of break has somewhere to surface other than your runners next time. Sorry for the round trip.

mosteo (approved):

PR #2055, opened 2026-08-08
adsa 1.5.1: new release

charlie5:
Patch release: fixes Windows partition links (v1.5.0's are broken there) and Debian-family shared-libgnat toolchains. Home: https://codeberg.org/charlie5/aDSA

PR #2021, opened 2026-07-16
sparklib fsf library_kind hotfix

AldanTanneo:
Patch the `SPARKlib_Library_Kind` type, that hasn't been modified to accept `relocatable` and `static-pic` libraries. See https://github.com/AdaCore/SPARKlib/issues/2.

PR #2063, opened 2026-08-16
adsa 1.5.0: origin URL moved to GitHub (hash unchanged)

charlie5:
The adsa project has moved hosting from Codeberg to GitHub (github.com/charlie5/aDSA).

This updates the already-published 1.5.0 manifest's origin URL (and website) to the GitHub release asset. The tarball is byte-identical: the sha512 in the manifest is unchanged, and I have re-verified it against the file GitHub actually serves.

Related: #2055 (the pending 1.5.1 release) has been updated the same way, so both crate versions resolve to GitHub.

### DIFF
--- a/index/ad/adsa/adsa-1.5.0.toml
+++ b/index/ad/adsa/adsa-1.5.0.toml
@@ -5,7 +5,7 @@ version = "1.5.0"
 licenses = "GPL-3.0-or-later WITH GCC-exception-3.1"
 maintainers = ["Rod Kay <rodakay5@gmail.com>"]
 maintainers-logins = ["charlie5"]
-website = "https://codeberg.org/charlie5/aDSA"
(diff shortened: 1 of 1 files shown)


charlie5:
The "Check Author" failure is the expected by-design guard for PRs that modify an already-published manifest ("ensure manually no maintainers were modified") — flagging this for manual review as it requests.

For the reviewer, the complete diff is two fields in `adsa-1.5.0.toml`:
- `website`: `codeberg.org/charlie5/aDSA` → `github.com/charlie5/aDSA`
- `origin.url`: the same tarball's new home under GitHub releases

Untouched: `maintainers`, `maintainers-logins`, the sha512 (the GitHub-served asset is byte-identical — re-verified), and everything else. Same pattern as previous in-place fixes like #2027 / #2021.

mosteo (approved):

PR #2065, opened 2026-08-17
uarp_sdk 0.5.3

chabanov:
Adds `uarp_sdk` 0.5.3 — an Ada client for the UARP (Snaga) Universal Agent
Runtime Platform API. This is the crate's first submission to the index.

- Source archive is attached to the upstream GitHub release:
  https://github.com/Snaga-AI/uarp-sdks/releases/tag/v0.5.3
- Manifest generated by `alr publish` (alr 2.1.1), unedited.
- Depends on `gnatcoll^26.0.0`.
- `libcurl` is linked via `-lcurl` from the project's own GPR rather than
  declared as an external, since Alire's `libcurl` detection only covers
  Linux package managers and the crate also targets macOS.

The library builds as relocatable, static or static-pic through the
`UARP_SDK_LIBRARY_TYPE` scenario variable.

### DIFF
--- a/index/ua/uarp_sdk/uarp_sdk-0.5.3.toml
+++ b/index/ua/uarp_sdk/uarp_sdk-0.5.3.toml
@@ -1,0 +1,27 @@
+name = "uarp_sdk"
+description = "Ada client for the UARP (Snaga) Universal Agent Runtime Platform API"
+version = "0.5.3"
+licenses = "MIT"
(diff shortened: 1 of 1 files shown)


chabanov:
Closing in favour of #2059, which proposes the same 0.5.3 manifest and has the older queue position — I opened this one before noticing #2059 was still open. Sorry for the duplicate.

PR #2071, opened 2026-08-20
opencv_core 0.1.0

zackboll:

### DIFF
--- a/index/mi/mingw_w64_gcc/mingw_w64_gcc-external.toml
+++ b/index/mi/mingw_w64_gcc/mingw_w64_gcc-external.toml
@@ -1,0 +1,16 @@
+name = "mingw_w64_gcc"
+description = "GNU C and C++ compiler toolchain for MinGW-w64"
+
+maintainers = ["Zack Boll <zackboll@gmail.com>"]
+maintainers-logins = ["zackboll"]
+
+[[external]]
+kind = "system"
+
+[external.origin.'case(distribution)']
+msys2 = ["mingw-w64-x86_64-gcc"]
+
+[external.available.'case(distribution)']
+msys2 = true
+'...' = false
+
--- a/index/op/opencv/opencv-external.toml
+++ b/index/op/opencv/opencv-external.toml
@@ -1,0 +1,16 @@
+description = "Open Source Computer Vision Library"
+name = "opencv"
+
+maintainers = ["Zack Boll <zackboll@gmail.com>"]
+maintainers-logins = ["zackboll"]
+
+[[external]]
+kind = "system"
+
+[external.origin."case(distribution)"]
+arch = ["opencv"]
+"centos|fedora|rhel" = ["opencv-devel"]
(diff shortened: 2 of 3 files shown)


mosteo (changes requested):
Errors in checks seem to be caused by missing underlying opencv. Ideally this dependency should be provided via an external crate definition for opencv packages, see e.g. https://github.com/alire-project/alire-index/blob/stable-1.4.0/index/li/libcurl/libcurl-external.toml

That new external crate can be provided in this same PR so both are tested together without delay.

zackboll:
Added an opencv system external following the libcurl example and made opencv_core depend on it. I also added the existing pkg_config external as a dependency since the pre-build configuration uses pkg-config. The changes are pushed to this PR; the new workflow runs currently appear to be waiting for approval.

mosteo:
I can see different kinds of errors in case some are easily fixable before merging:
- In Ubuntu LTS, the testsuite fails [two tests](https://github.com/alire-project/alire-index/actions/runs/32674909900/job/97377682608?pr=2071#step:6:4616).
- In some platforms there's a macro missing, this may be related to the OpenCV version.
- Other platforms err in the script because opencv4 is not found.

zackboll:
I will try to take a look at this.  I am developing against OpenCV 4.10 on Debian 13.  I can probably support multiple 4.x versions with some effort.

zackboll:
It looks like some of the pipelines tried to build against OpenCV5.  I will take a look at the API differences, and decide whether it makes sense to have one Ada thick binding that supports both OpenCV4 and OpenCV5, or whether it would make more sense to have separate crates targeting CV4/CV5.

mosteo:
If your binding is 4-specific, you could change the general dependency on opencv to ^4, and see if that improves things.

zackboll:
I was able to handle minor variations in OpenCV version behavior in the C++ shim layer.  4.6 failures have been resolved.  It looks like OpenCV 5 failures were all related to changes in empty Mat behavior, I should have an updated pull request soon.

zackboll:
Looks like I still have failures for a variety of reasons.

My project currently has GitHub runners for:
Debian12-OpenCV-4.1
Debian12-OpenCV-4.6
Debian13-OpenCV-4.10
Debian13-OpenCV-5.0

I plan to add additional runners that more closely mirror the failing platforms next.

mosteo:
It seems there's only a Windows hurdle remaining.

zackboll:
I was able to get a Windows runner to build / pass unit tests in my repo, I will have to look at what is different between the runners.

https://github.com/zackboll/opencv_core_ada/actions/runs/33471747300/job/99742734530

zackboll:
It looks like everything builds on Windows now, and the remaining issue is an environment issue for executing the unit tests.  I will update opencv_core.toml today to have the needed environment change for Windows.

mosteo (approved):

mosteo:
All green and merged. Thanks for the effort and the patience.

PR #2077, opened 2026-08-23
lithophane_ada 0.1.0

smionean:

### DIFF
--- a/index/li/lithophane_ada/lithophane_ada-0.1.0.toml
+++ b/index/li/lithophane_ada/lithophane_ada-0.1.0.toml
@@ -1,0 +1,26 @@
+name = "lithophane_ada"
+description = "Lithophane generator with Ada"
+version = "0.1.0"
+
+authors = ["Simon Bean"]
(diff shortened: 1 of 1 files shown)


index/li/lithophane_ada/lithophane_ada-0.1.0.toml
```diff
@@ -0,0 +1,26 @@
+name = "lithophane_ada"
```
mosteo, line 1: We strongly encourage not using "ada" in the name of a crate for the Ada ecosystem as it's basically noise in most cases. In your case just "litophane" would be a fine crate name.

Note that you can keep your repository name as-is (and code as well, although that's more visible for users).  

You'll need to rename both the crate and the manifest file. Or you could resubmit, as you prefer.

mosteo (changes requested):

index/li/lithophane_ada/lithophane_ada-0.1.0.toml
smionean: I will resubmit. Thank you

PR #2078, opened 2026-08-23
lithophane 0.1.0

smionean:


mosteo (approved):

index/li/lithophane/lithophane-0.1.0.toml
```diff
@@ -0,0 +1,26 @@
+name = "lithophane"
+description = "Lithophane generator with Ada"
+version = "0.1.0"
```
krischik, line 3: A long_description is aways good for visibility. It's optional of course so maybe something for the next release.

PR #2080, opened 2026-08-24
Ada Class Library release 7.2.1

krischik:
New adacl_serial crate. New decimal number formater.


krischik:
@mosteo 

> ERROR: Solution contains more than one compiler?

Where does this come from because I don't specify any compiler at all.

mosteo:
> Where does this come from because I don't specify any compiler at all.

This is a rare situation in our checks when trying to find a compiler able to compile a release. Nothing on your side, I must fix it at some point.

mosteo (approved):

PR #2087, opened 2026-08-24
Add libsdl3, libsdl3_ttf and libsdl3_image externals

ovenpasta:
SDL3 development files, so Ada crates binding SDL3 can declare a system dependency. The index currently has libsdl2, libsdl2_ttf, libsdl2_image and libsdl2_mixer, but nothing for SDL3.

Package names were verified against each distribution's own package database rather than pattern-matched from the SDL2 entries, which would have been wrong in three places:

- MacPorts drops the prefix: the port is `SDL3`, not `libsdl3` (`libsdl3` 404s).
- MSYS2 switched to lowercase: `mingw-w64-x86_64-sdl3`, where SDL2 is `SDL2`.
- Gentoo is split: `media-libs/libsdl3` for the core, `media-libs/sdl3-ttf` and `media-libs/sdl3-image` for the satellites.

`centos` is omitted from all three: the Fedora package pages list only F43-45 and rawhide, with no EPEL branch, so an entry there would name a package that does not resolve. Adding `centos|` later is a one-word change if EPEL picks them up.

Availability, for reference: Debian trixie and Ubuntu 25.10 / 26.04 LTS carry all three; bookworm and 24.04 LTS carry none.

### DIFF
--- a/index/li/libsdl3/libsdl3-external.toml
+++ b/index/li/libsdl3/libsdl3-external.toml
@@ -1,0 +1,16 @@
+description = "Simple DirectMedia Layer 3 development files"
+name = "libsdl3"
+
+maintainers = ["ovenpasta@pizzahack.eu"]
+maintainers-logins = ["ovenpasta"]
+
+[[external]]
+kind = "system"
+    [external.origin.'case(distribution)']
+    'debian|ubuntu' = ["libsdl3-dev"]
+    'arch' = ["sdl3"]
+    'fedora' = ["SDL3-devel"]
+    'homebrew' = ["sdl3"]
+    'macports' = ["SDL3"]
+    'gentoo' = ["media-libs/libsdl3"]
+    'msys2' = ["mingw-w64-x86_64-sdl3"]
--- a/index/li/libsdl3_image/libsdl3_image-external.toml
+++ b/index/li/libsdl3_image/libsdl3_image-external.toml
@@ -1,0 +1,16 @@
+description = "Simple DirectMedia Layer 3 development files - Image"
+name = "libsdl3_image"
+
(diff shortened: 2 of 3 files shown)


krischik:
@mosteo These are very special packages and outside my expertise.

ovenpasta:
> @mosteo These are very special packages and outside my expertise.

I've put my name as the maintainer but feel free to take ownership.

krischik:
> > @mosteo These are very special packages and outside my expertise.
> 
> I've put my name as the maintainer but feel free to take ownership.

You misunderstand. I help Alex with reviewing the pull requests. Basically going down [the checklist](https://github.com/alire-project/alire-index/blob/stable-1.4.0/REVIEWING.md) and help you when there is something missing.

And speaking of it reviewing : I just noted that the crates have no tags. Tags are actually mandatory. You don't need to make a new pull request. You can just add them to using the web editor.

ovenpasta:
> And speaking of it reviewing : I just noted that the crates have no tags. Tags are actually mandatory. You don't need to make a new pull request. You can just add them to using the web editor.

I don't see tags in any of the other externals including the sdl2 ones but if it is a new requirement I will add them but then worth also fixing the sdl2 ones at least...

mosteo:
> an entry there would name a package that does not resolve

To keep in mind for the future, this is no problem as the externals systems is designed with the idea that not all given packages may exist (e.g. old ones removed in newer distros). The ones found are resolved to their versions. It's true that each package takes a bit of time to resolve so it's better not to have too many unresolvable packages.

mosteo (approved):

mosteo:
> I don't see tags in any of the other externals

This is a bad practice that originated in that those packages already are documented by the system. You can include tags in future releases of externals if you wish.

PR #2088, opened 2026-08-25
adsa 1.5.3

charlie5:
New release of `adsa` — a Partition Communication Subsystem for Ada's Distributed Systems Annex (Annex E), replacing the discontinued PolyORB/GLADE runtime.

Diagnostics-only release: no wire, protocol or API change (1.5.2 and 1.5.3 deployments interoperate byte-identically). It removes the three ways the PCS could fail without reporting anything — a partition whose main raised an exception deadlocked on its own serve tasks with the exception never printed; an unregistered RCI unit was ten silent seconds; and leftover build artefacts could silently strip the DSA stubs from a partition. All found via a user report on a textbook example.

Release notes: https://github.com/charlie5/aDSA/releases/tag/v1.5.3

New manifest (no published manifest modified); plain git origin pinning the release tag commit `4926831`.

### DIFF
--- a/index/ad/adsa/adsa-1.5.3.toml
+++ b/index/ad/adsa/adsa-1.5.3.toml
@@ -1,0 +1,41 @@
+name = "adsa"
+description = "Ada Annex E (DSA) runtime + gnatdist replacement, backed by ZeroMQ"
+version = "1.5.3"
+
+licenses = "GPL-3.0-or-later WITH GCC-exception-3.1"
+maintainers = ["Rod Kay <rodakay5@gmail.com>"]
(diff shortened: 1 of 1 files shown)


krischik:
      apparent.
- [x] For repeat releases, check the Diff workflow output to quickly spot the
- [x] `gpr-set-externals` settings are appropriate (see Common pitfalls below).
- [x] System library dependencies are declared and not left implicit. If a

@mosteo @Fabien-Chouteau @AldanTanneo looks OK.

mosteo (approved):

PR #2093, opened 2026-08-27
Update Gnoga 2.2.3 series of programs.

Blady-Com:


Blady-Com:
Seems that the failed builds are due to a lack of sqlite3 lib. How to provide sqlite3 for them?

mosteo:
There seems to be a missing dependency on sqlite3 that is not available by default in all platforms:

> /root/.local/share/alire/toolchains/gnat_native_16.1.0_9f74f58a/x86_64-pc-linux-gnu/bin/ld: cannot find -lsqlite3: No such file or directory

Blady-Com:
> There seems to be a missing dependency on sqlite3 that is not available by default in all platforms:

Yes, thats I got, see my previous post.

Shouldn’t I set Available to false for those plateforms with missing sqlite3, should I ?
What else can be done?

krischik:
Hello @Blady-Com — a brief note on timing.
Alex has been unwell (nothing too serious, but it has made it hard to sit at the computer), so index submissions have not been processed as usual. He has posted an update here: https://github.com/alire-project/alire/discussions/2184
There is now a backlog of a few weeks / about forty pull requests. I can help with review comments, but I am not able to approve or merge. We are waiting for Alex to be back on form and/or reaching some of the reserve moderators.
Please bear with us — your submission has not been forgotten. Get well soon, Alex.

AldanTanneo:
> Shouldn’t I set Available to false for those plateforms with missing sqlite3, should I ? What else can be done?

You can create a crate manifest for the sqlite3 system dependency; see for instance, the `libgmp` crate manifest, which specifies what system package is required depending on the OS.

Blady-Com:
> You can create a crate manifest for the sqlite3 system dependency; see for instance, the `libgmp` crate manifest, which specifies what system package is required depending on the OS.

I would have done that but I don't how sqlite3 is installed in these distributions.
What are the actual lib names? How to get version-output? etc.
I'll appreciate any help.

PR #2099, opened 2026-08-30
Add backoff_ada, prometheus_ada 0.1.0

kokhlo:
Adds two new crates to the community index:

- backoff_ada 0.1.0 — exponential backoff with jitter and retry
- prometheus_ada 0.1.0 — Prometheus metrics client (Counter/Gauge/Histogram, labels, text exposition v0.0.4)

Both tarballs are tagged releases (v0.1.0) with sha512 hashes computed from the published archives.


mosteo (changes requested):
The `_ada` suffixes are frowned upon since in the Ada ecosystems that's usually noise. Note that you can keep the repo name as you wish, even after renaming the crates. Ideally, the Ada packages should also not have `_Ada`, but that's up to you.

Note also that the manifests files you're submitting are not at their expected locations.  They should be at `index/ba/backoff/backoff-0.1.0.toml` and likewise for prometheus. If you use `alr publish`, it should do it for you or at least guide you.

kokhlo:
Addressed both points:

- Crates renamed: `backoff_ada` → `backoff`, `prometheus_ada` → `prometheus` (repo names unchanged, per your note that those can stay).
- Manifests relocated to the expected index paths: `index/ba/backoff/backoff-0.1.0.toml` and `index/pr/prometheus/prometheus-0.1.0.toml`. The old-root copies are removed from this branch.

Next submission will go through `alr publish` as suggested — good call on it handling placement.

kokhlo:
Ping — backoff and prometheus were renamed per your review and pushed on Sep 2. Happy to adjust further if anything else is off.

mosteo:
It seems now there's a problem with the declared hashes for the source tarballs.

kokhlo:
You're right — thanks for catching it. Both declared hashes were stale: I had recomputed them against locally-produced tarballs instead of the ones GitHub generates from the tag.

Fixed in 7083d9a6c8 and eec48eba28: both manifests now carry the sha512 recomputed from the actual tarballs served at the `[origin]` URLs (verified stable across two independent downloads):

- `backoff-0.1.0.toml` → `sha512:89e458b66cfd04f35aefb4d34b10588b4a76c671652e3ae645a884689c6b346c5a96603d8933082fe5362189c9366cf35711381176e7f0000d37c7a4a8803591`
- `prometheus-0.1.0.toml` → `sha512:379ce381a52a9273020486a1376d3a443882e941d0efe98e6ef2ffac02c780784fab5ddb7068823533ce402eac5b119343ab467e1f17be989a49ba865036e400`

For future submissions I'll use `alr publish`, which computes the hash from the served archive directly and avoids this class of mistake entirely.

PR #2104, opened 2026-08-31
cpuload 0.0.1

adelnoureddine:


adelnoureddine:
CPU Load is a library to get CPU usage on Windows and Linux, for the entire system, for a specific process by its ID, and for a specific application by its name (tracking all of the applications' processes).

mosteo (approved):

PR #2111, opened 2026-09-02
png_2_ada 1.0.0

RREE:

### DIFF
--- a/index/pn/png_2_ada/png_2_ada-1.0.0.toml
+++ b/index/pn/png_2_ada/png_2_ada-1.0.0.toml
@@ -1,0 +1,20 @@
+name = "png_2_ada"
+description = "convert a PNG image into an embedded Ada package"
+version = "1.0.0"
+
+authors = ["Rolf Ebert"]
+maintainers = ["Rolf Ebert <rolf.ebert.gcc@gmx.de>"]
+maintainers-logins = ["RREE"]
+licenses = "MIT OR Apache-2.0 WITH LLVM-exception"
+website = ""
+tags = ["image", "embedded", "png"]
+
+executables = ["png_2_ada"]
+
+[[depends-on]]
+gid = "^14.0.0"
+
+[origin]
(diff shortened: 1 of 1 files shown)


RREE:
The automatic name check fails as this crate is named 'png_2_ada'. I think it is a valid name as the program generates Ada code for a given PNG image file. What else would be an appropriate name without the three letters 'ada'?

krischik:
@RREE

Thanks for the explanation — that is a perfectly reasonable question.

The automatic name check is flagging `png_2_ada` because most crates in the index are already written in Ada, so putting `ada` in the crate name is usually treated as redundant. The published guidance is mainly about avoiding `ada` as a *prefix* (it makes crates harder to find in a list). An `ada` *suffix* is normally reserved for bindings to an existing non-Ada library, such as `gtkada` or `sdlada`.

Your case is a bit different: this is a converter that *emits* Ada source from a PNG, so the `ada` part is describing the output rather than restating the implementation language. That is a fair argument, and it may well be accepted.

If you would rather sidestep the check altogether, names that still describe the tool without `ada` could include something like `png_codegen`, `png2src`, `png_embed`, or `png_to_source`. Please do not feel obliged to rename it if you think `png_2_ada` is the clearest name.

I am only helping with the review and cannot approve the pull request myself — Alex will make the final decision on the name.

If you would like to keep the name as it is, please **mark the pull request as ready for review** when you are happy with it — it is still in draft at the moment, so it will not be considered for merging until then.

mosteo:
I'm late to the party but as Martin explained, the rule is there to avoid cases where "ada" doesn't add anything substantive to crate names. In this case I think it's perfectly acceptable as it is the purpose of the tool.

RREE:
Alejandro,

I renamed it to png_embed in the meantime. But versions 1.0.0 and 1.0.1 fail for different reasons. Perhaps you could restart the automatic tests of 1.0.1

21.09.2026 19:19:38 Alejandro R Mosteo ***@***.***>:

PR #2114, opened 2026-09-04
adalang_analyzer 1.4.0

mmartign:

### DIFF
--- a/index/ad/adalang_analyzer/adalang_analyzer-1.4.0.toml
+++ b/index/ad/adalang_analyzer/adalang_analyzer-1.4.0.toml
@@ -1,0 +1,80 @@
+# AdaLang Analyzer
+#
+# Copyright (C) 2026, Spazio IT
+#
+# This file is part of AdaLang Analyzer.
+#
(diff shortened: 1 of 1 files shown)


mmartign:
Please **do not merge this one** — it is superseded by #2115 (`adalang_analyzer 1.4.1`).

Shortly after opening this PR, a new benchmark corpus (`--verify` vs GNATprove over 124 units of the AdaCore SPARK testsuite) turned up two correctness bugs in `--verify`:

- **FP-064** — a possible unsoundness: an index into an array slice was proved safe against the array type's index subtype rather than the slice's own (possibly empty) bounds.
- **FP-065** — a false positive: `pragma Assert (False)` on an unreachable, unevaluated-guard branch was reported as a definite error.

Both are fixed in 1.4.1. 1.4.0 and 1.4.1 are otherwise identical, so I'd rather not put 1.4.0 in the index at all. Closing this in favour of #2115 — sorry for the churn.

mmartign:
Superseded by #2115 (adalang_analyzer 1.4.1).

PR #2115, opened 2026-09-06
adalang_analyzer 1.4.1

mmartign:
Created via `alr publish` with `alr 2.1.0`.

Bug-fix release. Supersedes the still-open 1.4.0 PR #2114 — please take 1.4.1 instead of 1.4.0. The two releases differ only in two `--verify` correctness fixes (and the benchmark corpus that found them):

- **FP-064** (possible unsoundness): an index into an array slice (`A (Lo .. Hi) (I)`) was proved safe against the array type's declared index subtype rather than the slice's own bounds, which may be empty.
- **FP-065** (false positive): `pragma Assert (False)` on an unreachable branch whose guard could not be evaluated was reported as a definite error.

Manifest diff vs 1.3.0 is version + origin commit only.


mmartign:
Superseded by #2124 (adalang_analyzer 1.5.0).

PR #2119, opened 2026-09-08
joularcore 0.0.3

adelnoureddine:


adelnoureddine:
Updated version with support for EMI interface and PawnIO driver on Windows, and few other fixes for CPU/GPU reading on Windows, Linux and macOS.

adelnoureddine:
Seems the error in the check if related to github's network, not my crate.

krischik:
> Seems the error in the check if related to github's network, not my crate.

You can force another build by making a change to the toml file. Deleting one of the empty lines at the end is a good candidate for that.

PR #2120, opened 2026-09-08
cpuload 0.0.3

adelnoureddine:


adelnoureddine:
Updated version with fixes for all OSes, use readlink on Linux, use nanoseconds for all systems, report unavailable load, and other minor fixes.

adelnoureddine:
Seems the error in the check if related to github's network, not my crate.

PR #2129, opened 2026-09-14
Ada Class Library adacl_desktop release 8.0.0

krischik:
Split adacl into adacl and adacl_desktop.

Needs adacl 8.0 to successfully build.


krischik:
@mosteo I have cleaned up the packages structure to better support both desktop and embedded development so there is a new adacl package.

AldanTanneo (approved):

PR #2148, opened 2026-09-17
Add 6 kokhlo Ada crates: redis_ada, websocket_ada, snowflake_ada, prometheus_ada, backoff_ada, ipfs

kokhlo:
This PR adds six Ada crates to the Alire community index:

1. **redis_ada 0.2.0** — Redis client library (RESP2 protocol codec + connection layer)
2. **websocket_ada 0.2.0** — RFC 6455 WebSocket client (framing + handshake)
3. **snowflake_ada 0.1.0** — Twitter Snowflake distributed ID generator
4. **prometheus_ada 0.1.0** — Prometheus text-format metrics client
5. **backoff_ada 0.1.0** — Exponential backoff with jitter and retry
6. **ipfs 0.1.0** — IPFS client library (CID, CAR, UnixFS, gateway client)

All crates:
- MIT licensed
- Maintainer: kokhlo
- CI green (GitHub Actions)
- Production-ready initial releases

Repository URLs: https://github.com/kokhlo/{redis-ada,websocket-ada,snowflake-ada,prometheus-ada,backoff-ada,ipfs}


krischik:
Hello @kokhlo — thank you for submitting these crates; it is lovely to see a batch of useful libraries landing in the index.

A small naming note from the [Alire policies](https://alire.ada.dev/docs/policies): putting `ada` in a crate name is generally discouraged. It is redundant here, because the Alire index mostly only contains Ada (and SPARK) code, and it can make crates harder to find in a list.

There are well-known exceptions:

- older crates, libraries, tools that were published before this guidance existed;
- language bindings for an existing non-Ada library, where an `ada` suffix is accepted (for example `sdlada`, `gtkada`).

These crates look like original Ada implementations rather than bindings, so names such as `redis_ada`, `websocket_ada`, `snowflake_ada`, `prometheus_ada` and `backoff_ada` would preferably drop the `_ada` (for example `redis`, `websocket`, and so on). `ipfs` is already fine on that point.

Only the crate *name* is affected. You can rename on GitHub without opening a new pull request.

If you would rather keep the current names, please leave a short explanation addressed to @mosteo so the maintainers can consider an exception.

I am only helping with review and do not make the final decision myself. Happy to look again once you have either renamed or explained the names — thank you again for the contribution.

mosteo:
As Martin explained, in this case I don't see a need for the `_ada` in the crate names.

As for the failing tests, the manifest folders are wrong. Two letter folders are expected as the top classifier, you can take a look at the index in this repository in case of doubt.

PR #2151, opened 2026-09-19
Create hac-0.45.0.toml for HAC Ada Compiler

zertovitch:
New in 0.45:
* aggregates (in the special context of x := (...))
* record default values


krischik:
      apparent.
- [ ] For repeat releases, check the Diff workflow output to quickly spot the
- [ ] `gpr-set-externals` settings are appropriate (see Common pitfalls below).
- [x] System library dependencies are declared and not left implicit. If a
      needed external crate does not exist yet, request its external to be
      added in the same PR. Examples:

@alire-project/crate-reviewers (@mosteo @AldanTanneo @Fabien-Chouteau) Please have a look at the gpr-set-externals.

AldanTanneo (approved):

AldanTanneo:
I don't see a `[gpr-set-external]` section, only a `[gpr-external]`. It could be better documented (what are the defaults?) but it's not a blocker for merging.

